PR #653, opened 2014-12-29
dependent: :delete, :destroy, :delete_orphans, :destroy_orphans

subvertallchris:
`:destroy` and `:delete` (should that be `:delete_all`?) do exactly what you expect. The other two...

Imagine a graph of Route and Stop nodes. A Route may have 0 or more Stops, a Stop may be shared between multiple Routes, a Stop must always have at least 1 Route. You want to ensure that when a route is destroyed, its stops will be destroyed if they don't have any other routes. Enter `:delete_orphans` and `:destroy_orphans`. They both perform a match to find nodes whose only relationship of the given type is to the node being destroyed. `:delete` happens in Cypher, `:destroy` happens in Ruby.

There is currently no protection against loops within Ruby. TBD whether we'll handle that in this PR or another one.


subvertallchris:
Should the option for delete be `dependent: :delete_all`? That's what it is in ActiveRecord, I think. Just seems like a weird pairing with `dependent: :destroy`.

subvertallchris:
Better tests now. Haven't tested in JRuby, not sure if it'll be any happier.

I'd like to do a tiny bit of refactoring to `has_many` and `has_one`. For now, I added a `#{name}_query_proxy` alias on `has_many` but it should behave exactly like the corresponding method for `has_one`. I need a way to access the QP object for an association using the same method, regardless of whether it's many or one, and this seems like the way to do it.

lib/neo4j/active_node/has_n.rb
```diff
@@ -183,7 +168,16 @@ def #{name}(node = nil, rel = nil, query_proxy = nil)
...
       end
+
+      private
+
+      def build_association(macro, direction, name, options)
+        association = Neo4j::ActiveNode::HasN::Association.new(macro, direction, name, options)
+        @associations ||= {}
+        @associations[name] = association
+        create_reflection(macro, name, association, self)
+        association
```
cheerfulstoic: I have this semi-weird style thing where I like to use `#tap` for this sort of thing.  What do you think?

    Neo4j::ActiveNode::HasN::Association.new(macro, direction, name, options).tap do |association|
      @associations ||= {}
      @associations[name] = association
      create_reflection(macro, name, association, self)
    end

lib/neo4j/active_node/query/query_proxy_enumerable.rb
```diff
@@ -0,0 +1,67 @@
+module Neo4j
+  module ActiveNode
+    module Query
+      # Methods related to returning nodes and rels from QueryProxy
+      module QueryProxyEnumerable
```
cheerfulstoic: Yay, good module!

lib/neo4j/active_node/has_n.rb
subvertallchris: You're right, `tap` is the right way to do it. It'll be in the next PR.

spec/e2e/association_dependency_spec.rb
```diff
@@ -0,0 +1,183 @@
...
+          expect(DependentSpec::Comment.count).to eq 0
+        end
+      end
+    end
+
+    context 'things are going terribly' do
+      describe 'Grzesiek, in frustration, destroys his account' do
+        it 'destroys all bands, tours, routes, stops, and comments' do
+          expect { @user.destroy }.not_to raise_error
+          expect(DependentSpec::Tour.count).to eq 0
```
cheerfulstoic: What about something like:

    {
      DependentSpec::Tour => 0,
      DependentSpec::Route => 0,
      DependentSpec::Stop => 0,
      DependentSpec::Comment => 0,
      DependentSpec::Band => 0
    }.each do |model_class, count|
      expect(model_class.count).to eq count
    end

cheerfulstoic: Hmmm, I'm waffling on that one though.  I'm not sure that that is entirely better or not

cheerfulstoic:
Seems good to me!

spec/e2e/association_dependency_spec.rb
subvertallchris: I played around with different approaches but kind of settled on what's there, the dumbest/most straightforward, cause it's easiest to read.

cheerfulstoic: Yeah, don't mind me... ;p

I think the one advantage that is has is that at a glance you can tell than there aren't subtle differences between the lines because they're all executing that one line in the each.

subvertallchris:
Working on merging in all the master changes, then I think we're good to go here... once we decide if `dependent: :delete` should be `dependent: :delete_all` to match ActiveRecord. I vote yes.

subvertallchris:
:warning: 
Attn @cheerfulstoic!
Gonna need to do a `git push -f` on this branch in the morning to get it up to date.
:warning:

subvertallchris:
See #658.

PR #655, opened 2015-01-05
do not reload node upon calling 'frozen?' method

dpisarewski:


subvertallchris:
If you're still working from your local branch, make sure to pull the changes in https://github.com/neo4jrb/neo4j/commit/45c87b318e2cf77c131fc47c6c363f2891d44db7. Freezing the node's props causes a problem when `destroy` transactions fail -- the props are still frozen! This fixes that.

PR #658, opened 2015-01-06
Dependent merge

subvertallchris:
It's like #653... but Github is able to merge it!


subvertallchris:
I vote `delete_all` and `delete_all_orphans`. Second?

cheerfulstoic:
I'm find with `delete_all(_orphans)`

subvertallchris:
Actually, gonna stick with `:delete`. `:delete_all` has some different connotations depending on the type of association. We are deleting records, so we'll stick with that. Merge time!

PR #659, opened 2015-01-06
move dev dependencies to gemspec

subvertallchris:
Just makes things a bit cleaner.


subvertallchris:
Pretty superficial change but @cheerfulstoic, you cool with this?

cheerfulstoic:
Sure!

You should have guard running while you develop so that it'll fix things like that one commit you made automagically

subvertallchris:
That's a really good point.

PR #661, opened 2015-01-08
includes method lets you use one query to load two associations

subvertallchris:
This isn't ready for merge yet but opening the PR to raise awareness. It needs some features, maybe some API tweaks, and a lot of cleanup. QueryProxy is doing too much, there may be another class hiding in there.

ActiveRecord has at least three different methods that let you load two associations at once: `:eager_load`, `:preload`, and `:includes`. Going with `:includes` here cause it feels the most natural.

Very easy to use, inspired by AR but a little different. `students = Student.where(name: 'Lauren').includes(:lessons)`. It builds a Cypher MATCH and then OPTIONAL MATCH to get the lessons, then stores the results in each parent node's association cache. When you then do `students.each { |lauren| lauren.lessons }`, it retrieves the results from the association cache and doesn't have to do a query!

A nice bonus is that you can chain any QueryProxy methods onto `includes`. `Student.where(age: 19).includes(:lessons).rel_where(grade: 'A+')`. The downside here is that, at the moment, you would get to those results by rewriting that same query in your view: `student.lessons.rel_where(grade: 'A+')`.

To do:

* `includes_with_rel`
* Allow passing of arguments to set node and rel identifier so you have more control over filtering
* And a bigger one...

Should it put filtered results in the root association so you don't have to call the filter again when returning? When you do `Student.where(age: 19).includes(:lessons).rel_where(grade: 'A+')`, would you expect it to show the filtered results when you do `student.lessons` or would you expect to have to do `student.lessons.rel_where(grade: 'A+')` again? I kind of think it should be in `:lessons`, I already told it I wanted to filter... This would be easy enough to implement, we just need to decide if it's the most obvious way of handling it.


subvertallchris:
Also, JRuby... come on!

subvertallchris:
So... I think I want to skip the whole filter-association-details part of this for now so we can merge. Filterable includes can be part of another method or a tweak down the road.

subvertallchris:
The big change tonight is that I removed a lot of code in the effort to make it not allow filtering of included association matches, forgetting that it would just apply the results to the `OPTIONAL MATCH` created to preload and then stash the results in the base association. In other words, doing this:

```
lessons = Lesson.all.includes(:teachers).where(age: 30)
lessons.each do |lesson|
  lesson.teachers do |teacher|
    # something here
  end
end
```
...will not result in an extra query to the database, the filtered match will already be accessible in each `lesson`'s `teachers` association cache. This looks like a weird default but in practice, it was exactly what I expected to happen every time I was testing this feature, since I feel like `includes(:teachers)` implies that whatever you preload and filter will be in there.

This update also brings support to `first` and `last`.

Assuming nobody HATES this, I'd like to merge and we can play around with it. We have something going at work where we can use this immediately and I'd like to see how it feels in an app. Brian, maybe you can do the same?

subvertallchris:
I'm closing this until I'm closer to being ready to merge.

Issue #607, opened 2014-11-25
Allow custom typecasters


dpisarewski:
I created a class to save timestamps with microsecond precision, a type converter and typecaster for it.

``` ruby
class Memo
  include Neo4j::ActiveNode
  property :created_at, index: :exact, type: MicrosecondTimestamp, typecaster: MicrosecondTimestampTypecaster.new
  property :updated_at, index: :exact, type: MicrosecondTimestamp, typecaster: MicrosecondTimestampTypecaster.new

  serialize :created_at, MicrosecondTimestamp
  serialize :updated_at, MicrosecondTimestamp
end
```

``` ruby
class MicrosecondTimestamp
  def initialize(*args)
    timestamp = args.first
    if timestamp.is_a?(Date) or timestamp.is_a?(Time)
      @timestamp = timestamp.to_datetime
    elsif timestamp.is_a?(Numeric)
      @timestamp = Time.at(timestamp).to_datetime
    else
      @timestamp = DateTime.new(*args)
    end
  end

  def to_s
    @timestamp.iso8601(6)
  end

  def inspect
    to_s
  end

  def method_missing(method, *args)
    @timestamp.send(method, *args)
  end
end
```

``` ruby
module Neo4j::Shared::TypeConverters
  # Converts timestamps with microsecond precision
  class MicrosecondTimestampConverter
    class << self

      def convert_type
        MicrosecondTimestamp
      end

      def to_db(value)
        return nil if value.nil?
        value.to_time.to_datetime.iso8601(6)
      end

      def to_ruby(value)
        return nil if value.nil?
        MicrosecondTimestamp.new(DateTime.parse(value))
      end
    end
  end
end
```

``` ruby
class MicrosecondTimestampTypecaster
  def call(value)
    value
  end
end
```

But after pulling recent changes it's broken. I get the following error:

```
ArgumentError:
       a typecaster must be given
     # /Users/dieter/.rvm/gems/ruby-2.1.2@8actions/gems/active_attr-0.8.4/lib/active_attr/typecasting.rb:49:in `typecast_attribute'
     # /Users/dieter/.rvm/gems/ruby-2.1.2@8actions/bundler/gems/neo4j-05f6936405ad/lib/neo4j/shared/property.rb:187:in `block in attribute!'
[...]

PR #671, opened 2015-01-19
ability to specify custom type converter on a property

dpisarewski:
It should add an option to property declaration which allows to specify a custom type converter for it. Mentioned in https://github.com/neo4jrb/neo4j/issues/607

For example:

```ruby
class RangeConverter
  class << self
    def convert_type
      Range
    end

    def to_db(value)
      value.to_s
    end

    def to_ruby(value)
      ends = value.to_s.split('..').map{|d| Integer(d)}
      ends[0]..ends[1]
    end
  end
end

class MyModel
  include Neo4j::ActiveNode

  property :range, type_converter: RangeConverter
end
```


cheerfulstoic:
Looks like it failed on travis due to rubocop.  I'd suggest running `guard` if you're making any changes to the gem so that it auto-corrects errors and also tells you about things it can't auto-correct

subvertallchris:
@dpisarewski, sorry for the slow merge here. This is really cool! I think we *might* actually be able to fix the existing TypeConverters module to allow `property :name, type: MyClass` without needing a different option. I stumbled upon it the other night while refactoring that module.

Can you create a `type_converters_refactor` branch on `neo4jrb/neo4j` with this PR's code? I'd like to experiment with it. I can fork your branch and create it myself if you don't have time.

PR #676, opened 2015-01-26
Typecasters

subvertallchris:
This combines @dpisarewski's #671 with the ability to do automatic type conversion using `type: YourClass`. There's a little crossover between the two but I think they both offer something unique enough to be worthwhile.

Dieter made it so you can do `property :my_range, type_converter: RangeConverter`, and explicitly tell the model to use a specific conversion class. I think this is useful if you have a basic primitive type that needs something special done to it to save it to the database... Maybe you want the DB version of DateTime to be a string instead of an integer or something. (@cheerfulstoic, isn't there an open issue that requests doing just that?)

My contribution lets you register a typecaster for a specific class, so you can do `property :my_range, type: Range` and it will take care of the rest. It keeps things simple and if nothing else, it means we support the real Time class.

I need to look at it more, I might be able to remove some of my code and piggyback off of Dieter's since they do some very similar things. Need to run for now, I'll check it more this week.


subvertallchris:
I combined some of our code to reduce duplication.

cheerfulstoic:
This seems good, though it just occurred to me (not sure why it didn't before) that maybe we should adopt the same convention as ActiveRecord's `serialize`.  We already have it for JSON, but I think it makes sense because that's essentially what we're doing.  We're making it clear to the user that the string in the neo4j database is a string but that they're storing a serialized Ruby object in it

subvertallchris:
After chatting with Brian, we cleared up something philosophically and think a modification to this is in order, which I just committed. The distinction made:

* A property's `type` option describes the class of object the user's code can expect, not how the database will persist the property.
* Since both "typecasting" and "type conversion" are floating around, we think `serializer` makes more sense for the option than `type_converter`. Serializing is what we are doing, so... let's be literal about it. :-D

@dpisarewski, how do you feel about this? I think it's all pretty cool. You'll notice that I merged your changes in with mine, so this is a proper group effort and I want to make sure you'll into it.

subvertallchris:
Been four days, no objects, merging!

dpisarewski:
Sorry for the late reply. That's really cool, that's what I wanted from the beginning. I guess we should document this feature in the wiki.

PR #692, opened 2015-02-20
When doing QueryProxy queries on models which are subclasses...

cheerfulstoic:
specify all labels in the `MATCH`

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -16,7 +16,7 @@ Metrics/AbcSize:
 # Offense count: 2
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 244
+  Max: 245
 
 # Offense count: 2
 Metrics/CyclomaticComplexity:
--- a/lib/neo4j/active_node/labels.rb
+++ b/lib/neo4j/active_node/labels.rb
@@ -9,6 +9,7 @@ module Neo4j
       WRAPPED_MODELS = []
 
       MODELS_FOR_LABELS_CACHE = ActiveSupport::Cache::MemoryStore.new
+      MODELS_FOR_LABELS_CACHE.clear
 
       included do |model|
         def self.inherited(model)
@@ -66,6 +67,14 @@ module Neo4j
             (model.mapped_label_names & labels).size
           end.last
         end
+      end
+
+      def self.clear_model_for_label_cache
+        MODELS_FOR_LABELS_CACHE.clear
+      end
+
+      def self.clear_wrapped_models
+        WRAPPED_MODELS.clear
       end
 
       protected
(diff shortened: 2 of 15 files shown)


subvertallchris:
I was thinking about this last night and now I'm not so sure about it. When doing `MATCH (u:User:GithubUser)`, wouldn't it match nodes with either label, not both labels?

ikwattro:
Nope it will return nodes with both labels. The problem with such query is a bit performance when dealing with big databases, check section `avoid the need to match on multiple labels` here http://graphaware.com/neo4j/2015/01/16/neo4j-graph-model-design-labels-versus-indexed-properties.html

subvertallchris:
Yeah, I double checked just after posting and noticed that. My bad. The stuff mentioned in that link under "Use well named relationship types to avoid some use of labels" was what I was talking about the other day, @cheerfulstoic.

cheerfulstoic:
Gotcha.  It was actually @jexp who suggested I use multiple labels.  I actually re-worked neo4apis to fit that idea.  I wonder if making multiple labels more performant is on the radar

cheerfulstoic:
Also, it just occurred to me, does this apply to neo4j 2.2?  One min, off to check ;p

jexp:
The performance impact is not so bad, it is cheaper to check for a label than for a property

cheerfulstoic:
Ah ha!  2.2 seem to help here!  I wrote a quick Ruby script ([code here](https://gist.github.com/cheerfulstoic/bae681a589c4dc172bad)) to get the number of DB hits for each query and the label order doesn't seem to matter in 2.2 (not that `User.login` isn't indexed here):

## `neo4j 2.1.7`

```
$ ruby profile_test.rb
MATCH (u:User) RETURN *: 8389
MATCH (u:User:GitHub) RETURN *: 8388
MATCH (u:User:StackOverflow) RETURN *: 8388
MATCH (u:User) WHERE u:GitHub RETURN *: 8388
MATCH (u:StackOverflow:User) RETURN *: 8646
MATCH (u:GitHub:User) RETURN *: 46574
MATCH (u:User) WHERE u.login = 'cheerfulstoic' RETURN *: 16776
MATCH (u:User:GitHub) WHERE u.login = 'cheerfulstoic' RETURN *: 25164
```

## `neo4j 2.2.0-M04`

```
$ ruby profile_test.rb
MATCH (u:User) RETURN *: 8389
MATCH (u:User:GitHub) RETURN *: 8388
MATCH (u:User:StackOverflow) RETURN *: 8388
MATCH (u:User) WHERE u:GitHub RETURN *: 8388
MATCH (u:StackOverflow:User) RETURN *: 8388
MATCH (u:GitHub:User) RETURN *: 8388
MATCH (u:User) WHERE u.login = 'cheerfulstoic' RETURN *: 16776
MATCH (u:User:GitHub) WHERE u.login = 'cheerfulstoic' RETURN *: 25164
```

jexp:
2.2. uses a cost based planner, so whichever label has the lower cardinality in the database (is more selective) is used first and then the second label is checked against it.

Issue #697, opened 2015-02-26
Is there a way to order on relationship properties?


nrdobie:
I'd like to be able to pull five of the latest items by a date property that is stored on a relationship. These items aren't guaranteed to be added to the database in chronological order so I can't just pull the last five items. Is there a function like `order` that can sort relationships or do I have to use a Cypher query?

Here are the models I have:

```ruby
class Console
  include Neo4j::ActiveNode

  property :name,        type: String
  property :manufacture, type: String

  validates_presence_of :name
  validates_presence_of :manufacture

  has_one :out, :successor,   model_class: Console
  has_one :in,  :predecessor, model_class: Console, type: 'SUCCESSOR'

  has_many :in, :games, rel_class: ReleasedFor, model_class: Game, unique: true

end

#####

class Game
  include Neo4j::ActiveNode

  property :title, type: String

  validates_presence_of :title

  has_many :out, :released_for, rel_class: ReleasedFor, model_class: Console, unique: true

end

#####

class ReleasedFor
  include Neo4j::ActiveRel

  from_class Game
  to_class   Console
  type       'RELEASED_FOR'

  property :date, type: Date
end
```

Here is the code I try running:

```ruby
console.games.order(date: :desc).limit(5).each { |g| puts g.title }
```

subvertallchris:
There are no options the ActiveRel class or ActiveNode associations that'll do that for you. You could create a scope or a class method that'll do it, then you can call `console.ordered_games.each`.

```
def self.ordered_games
  all.games.order(date: :desc).limit(5)
end
```

Call `all` at the beginning hook into the scope. Even though it's defined as a class method, you'll be able to use it on instances or within a QueryProxy chain.

nrdobie:
I was able to get it to work using `console.games.order("#{console.games.rel_var}.date DESC").limit(5).each { |g| puts g.title }`. Although it would be nice to have a function similar to `.rel_where` like `.rel_order` that would make it easier to do this.

subvertallchris:
That's a really good idea and it'd be very easy to implement. I can
probably find a few minutes to do it tomorrow.

cheerfulstoic:
I have to say I don't like this `rel_*` trend...  You don't have a `REL WHERE` in cypher, and I don't think we should have one in the ruby.

A few ideas:

```ruby
  .where({var_on_node: value1}, var_on_rel: value2)
  # same idea for .order, etc...

[...]

PR #698, opened 2015-02-27
Added rel_order for ordering on relationships.

nrdobie:
This will allow you to order based on relationship properties easily rather than using the current solution in I came up with in #697.

### DIFF
--- a/lib/neo4j/active_node/query/query_proxy.rb
+++ b/lib/neo4j/active_node/query/query_proxy.rb
@@ -112,7 +112,7 @@ module Neo4j
           @model.current_scope = previous
         end
 
-        METHODS = %w(where rel_where order skip limit)
+        METHODS = %w(where rel_where order rel_order skip limit)
 
         METHODS.each do |method|
           module_eval(%{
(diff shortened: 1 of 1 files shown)


subvertallchris:
Awesome! Could you throw a test in and run `rubocop -a` to get it into compliance?

nrdobie:
It looks like the issue with Rubocop is that with the additional function the class is now 247 lines long which exceeds the allowed 244 lines. To get compliance it looks like I'd have to either increase the limit, try to restructure the class to decrease the number of lines, or move some of the functionality of the QueryProxy class to a separate class.

I should be able to get some test written this evening.

cheerfulstoic:
I've made some changes to `master` to reduce the line count.  I think it could use some more substantial refactoring, but this should help for now

DavidChouinard:
Any update regarding merging this?

subvertallchris:
Good question. @nrdobie, what say you?

cheerfulstoic:
Closing in favor of #995 since so much has changed.  Thanks for the basis of this @nrdobie !

cheerfulstoic:
Released as 6.0.0.alpha.8

PR #713, opened 2015-03-08
Fix bug in delete_all: relations should be matched optionally

garmoshka-mo:
Otherwise you trap in situation, when nodes without relationships are ignored and stay in database

### DIFF
--- a/lib/neo4j/active_node/query/query_proxy_methods.rb
+++ b/lib/neo4j/active_node/query/query_proxy_methods.rb
@@ -55,7 +55,7 @@ module Neo4j
         def delete_all(identifier = nil)
           query_with_target(identifier) do |target|
             begin
-              self.query.with(target).match("(#{target})-[#{target}_rel]-()").delete("#{target}, #{target}_rel").exec
+              self.query.with(target).optional_match("(#{target})-[#{target}_rel]-()").delete("#{target}, #{target}_rel").exec
             rescue Neo4j::Session::CypherError
               self.query.delete(target).exec
             end


cheerfulstoic:
Thanks, this is great!  Could you put in a failing spec for this too?

garmoshka-mo:
Not sure what is 'failing spec'? 
To be honest, i'm not familiar with specs too much, in my projects using just tests..

cheerfulstoic:
I spec is basically just the RSpec term for test.  The idea is behavior-driven design, but that's not so important.

A failing spec would be a spec that would fail if your code wasn't in place.  That way we make sure this case doesn't happen again.

The appropriate place for such a spec would be here:
https://github.com/neo4jrb/neo4j/blob/master/spec/e2e/query_proxy_methods_spec.rb#L180

Note that to run the specs you should have a neo4j instance running on port 7474 (and you don't want to have anything important in it)

Though looking at the specs, it seems like there are a number of places which don't have additional relationships and they're getting deleted just fine.  If you can't run the tests, can you give an example of the situation you're running across?

Thanks

garmoshka-mo:
Unfortunately was unable to run rspec:

```
vagrant@precise32:~/ruby-test/lib/neo4j$ rspec spec/e2e/query_proxy_methods_spec.rb -e 'delete_all'
/usr/lib/ruby/vendor_ruby/rspec/core/configuration.rb:386:in `load': /home/vagrant/ruby-test/lib/neo4j/spec/e2e/query_proxy_methods_spec.rb:11: syntax error, unexpected ':', expecting kEND (SyntaxError)
...y :out, :lessons, model_class: 'Lesson', rel_class: 'Enrolle...
                              ^
/home/vagrant/ruby-test/lib/neo4j/spec/e2e/query_proxy_methods_spec.rb:11: syntax error, unexpected ',', expecting kEND
...essons, model_class: 'Lesson', rel_class: 'EnrolledIn'
                              ^
/home/vagrant/ruby-test/lib/neo4j/spec/e2e/query_proxy_methods_spec.rb:12: syntax error, unexpected ':', expecting kEND
      has_many :out, :things, model_class: false, type: 'lessons'
                                          ^
/home/vagrant/ruby-test/lib/neo4j/spec/e2e/query_proxy_methods_spec.rb:12: Can't assign to false
      has_many :out, :things, model_class: false, type: 'lessons'


....



vagrant@precise32:~/ruby-test/lib/neo4j$ rspec -v
2.5.1
```


But on simple test language it would look like this:
```
  test 'make sure we can simply burn down everybody with given type' do
    # Test - is ActiveNode with label :Test
    Nodes::Test.create(name: 'Xuev Zhid')

    Nodes::Test.all.delete_all

    x = Neo4j::Session.query
            .match('(n:Test)')
            .pluck('count(n)').first.to_i
    assert x == 0
  end
```

Unless you have `optional_match` - this test will fail

subvertallchris:
Yeah, I think it'd fail to get nodes without rels if you did
NodeClass.all.delete_all. NodeClass.delete_all would work, which is what we
usually use in tests to wipe out a class.

cheerfulstoic:
I think that this is worth adding it because it could confuse later.

I added a failing spec (see here: https://github.com/neo4jrb/neo4j/commit/e8684a0e765e54c30a4809efe02cdc9a7dadbf24 )

cheerfulstoic:
Merged!  Thanks for the help!

cheerfulstoic:
Oh, by the way, I also backported this as version 4.1.4

cheerfulstoic:
I'm not sure why you're getting the syntax error, by the way.  What version of ruby are you using?  I can see that sort of error happening before 1.9 (the gem requires 1.9.3)

garmoshka-mo:
I use ruby 2.0
```
vagrant@precise32:/etc/4neo4$ ruby -v
ruby 2.0.0p481 (2014-05-08 revision 45883) [i686-linux]
```
Probably need to look deeper to wat's wrong with it..
I installed rspec from Ubuntu's apt-get - maybe it's not configured well there..

cheerfulstoic:
Maybe you need to run `bundle` first to get rspec

Issue #720, opened 2015-03-16
QueryProxy#first and #last return same object in ordered multi-object list


posgarou:
I ran into some curious behavior today.

I have the following class and method:

```ruby
class Recipe
    has_many :out, :steps, type: :has_step, model_class: Graph::RecipeStep

    def steps_in_order
      steps.order(number: :asc)
    end
end
```

When I run `foo.steps_in_order.to_a`, everything looks normal: it has been sorted successfully by the number property.

```
[1] pry(#<RSpec::ExampleGroups::GraphRecipe::StepsInOrder>)> ordered_ingredients.to_a
[
    [0] #<Graph::RecipeStep number: 1, optional: false, text: "You put your left foot in, you take your left foot out">,
    [1] #<Graph::RecipeStep number: 2, optional: false, text: "You put your left foot in, you take your left foot out">,
    [2] #<Graph::RecipeStep number: 3, optional: false, text: "You put your left foot in, you take your left foot out">,
    [3] #<Graph::RecipeStep number: 4, optional: false, text: "You put your left foot in, you take your left foot out">,
    [4] #<Graph::RecipeStep number: 5, optional: false, text: "You put your left foot in, you take your left foot out">,
    [5] #<Graph::RecipeStep number: 6, optional: false, text: "You put your left foot in, you take your left foot out">
]
```

But if I call `#first` and `#last` directly on the QueryProxy object without calling `#to_a` first, then both methods return the first element:

```ruby
[2] pry(#<RSpec::ExampleGroups::GraphRecipe::StepsInOrder>)> ordered_ingredients.first
#<Graph::RecipeStep number: 1, optional: false, text: "You put your left foot in, you take your left foot out">
[3] pry(#<RSpec::ExampleGroups::GraphRecipe::StepsInOrder>)> ordered_ingredients.last
#<Graph::RecipeStep number: 1, optional: false, text: "You put your left foot in, you take your left foot out">
```

If I change the method above to `order(number: :desc)`, then `#first` and `#last` still return the same element, but this time it is the element that comes first in descending order (the step with number 6):

```ruby
[2] pry(#<RSpec::ExampleGroups::GraphRecipe::StepsInOrder>)> ordered_ingredients.first
#<Graph::RecipeStep number: 6, optional: false, text: "You put your left foot in, you take your left foot out">
[3] pry(#<RSpec::ExampleGroups::GraphRecipe::StepsInOrder>)> ordered_ingredients.last
[...]

PR #721, opened 2015-03-16
use COLLECT and HEAD/LAST for first and last

subvertallchris:
This should correct https://github.com/neo4jrb/neo4j/issues/720 but has two failing specs.

### DIFF
--- a/lib/neo4j/active_node/query/query_proxy_methods.rb
+++ b/lib/neo4j/active_node/query/query_proxy_methods.rb
@@ -3,18 +3,21 @@ module Neo4j
     module Query
       module QueryProxyMethods
         class InvalidParameterError < StandardError; end
+        FIRST = 'HEAD'
+        LAST = 'LAST'
 
         def first(target = nil)
-          query_with_target(target) { |var| first_and_last("ID(#{var})", var) }
+          first_and_last(FIRST, target)
         end
 
         def last(target = nil)
-          query_with_target(target) { |var| first_and_last("ID(#{var}) DESC", var) }
+          first_and_last(LAST, target)
         end
 
-        def first_and_last(order, target)
-          self.order(order).limit(1).pluck(target).first
+        def first_and_last(func, target)
+          query_with_target(target) { |var| self.pluck("#{func}(COLLECT(#{var})) as #{var}").first }
         end
+        private :first_and_last
 
         # @return [Integer] number of nodes of this class
         def count(distinct = nil, target = nil)
--- a/lib/neo4j/active_node/query/query_proxy_methods.rb
+++ b/lib/neo4j/active_node/query/query_proxy_methods.rb
@@ -15,8 +15,22 @@ module Neo4j
         end
 
         def first_and_last(func, target)
-          query_with_target(target) { |var| self.pluck("#{func}(COLLECT(#{var})) as #{var}").first }
+          new_query, pluck_proc = if self.query.clause?(:order)
+                                    new_query = self.query.with(identity)
+                                    pluck_proc = proc { |var| "#{func}(COLLECT(#{var})) as #{var}" }
+                                    [new_query, pluck_proc]
+                                  else
+                                    new_query = self.order(order).limit(1)
+                                    pluck_proc = proc { |var| var }
+                                    [new_query, pluck_proc]
+                                  end
+          result = query_with_target(target) do |var|
+            final_pluck = pluck_proc.call(var)
+            new_query.pluck(final_pluck)
+          end
+          result.first
         end
+
         private :first_and_last
 
         # @return [Integer] number of nodes of this class


subvertallchris:
@cheerfulstoic  could you eyeball this when you have a sec? Is there a better way to do this and/or do you know what the deal is with those failing specs? Won't be able to check it more until tomorrow.

cheerfulstoic:
I feel like if you're doing a `collect` that, even though it's only returning one item, the server is still potentially going to be doing a lot of work before it gets that head or tail.

I think we need to use have some way to determine if there is an order set on the chain.  If there it, we shouldn't add the order by `ID()`

subvertallchris:
That's definitely an idea. Is there an easy way to find out if a method has already been called on a chain? Something along the lines of `query.clauses_include?(:order)`?

cheerfulstoic:
It does now!  See: https://github.com/neo4jrb/neo4j-core/commit/f2edd20a52a0ac0fc044806fe1dc4a7844bc497f

subvertallchris:
Thanks! I'll see if I can finish this up tonight.

subvertallchris:
A bit of a delay there but back on it now. If someone is using `ORDER` earlier in their query, we're going to need to use `LAST(COLLECT(#{var}))` or find a way to modify their `ORDER` to use `DESC`. I'm going to patch this so it'll work for now but it'd be nice to find an inexpensive way to tweak the existing `ORDER` method.

cheerfulstoic:
If they're using `order` at an earlier point, can't we just do `limit(1)`?  I don't think we need the `collect`

subvertallchris:
`limit` makes it stop matching once it hits the first result, so the order
isn't actually applied until it matches everything. If you want the last
result, you have to match everything and use `desc` in a `with` followed by
a `return limit (1)` or `last(collect())`. `collect` adds no overhead that
I can see once we already have a set of results so it's not hurting
anything.

subvertallchris:
Haven't forgotten about this, will come back to it soon. If it's causing a problem for you or anyone else, let me know and I'll bump it up in the priorities list.

lib/neo4j/active_node/query/query_proxy_methods.rb
```diff
@@ -3,18 +3,35 @@ module ActiveNode
...
+                                    [new_query, pluck_proc]
+                                  else
+                                    new_query = self.order(order).limit(1)
+                                    pluck_proc = proc { |var| var }
+                                    [new_query, pluck_proc]
+                                  end
+          result = query_with_target(target) do |var|
+            final_pluck = pluck_proc.call(var)
+            new_query.pluck(final_pluck)
+          end
```
cheerfulstoic: Maybe just `end.first` so that we don't need the variable?

cheerfulstoic:
I'm not sure if you benchmarked, but it still seems like 'collect' would add overhead on the server because it needs to change a set of rows into one row with an array.  I also think the code could be much simpler if we just used `RETURN ... ORDER BY ... LIMIT 1`...

subvertallchris:
It only collects if the user has already ordered, that was the whole
problem. My benchmarks indicated COLLECT had almost no performance impact.
And rubocop complains if you do "end.first".

cheerfulstoic:
But I still don't think that we need to collect.  The problem was that when the user specifies an `order` previously that we were overwriting that with our own `id` order.  All we need to do is simply only apply the `order(:uuid)` only if no order was previously specified.  Then we get the `first` result of the `QueryProxy` chain so far (which either got the order from the user, or from us if it wasn't previously specified)

subvertallchris:
How would you handle "last" in cases where they've already ordered? "first"
isnt the problem, you can just limit 1. Last is the problem. There's no
Cypher function to return the last result without modifying the existing
order to DESC (or remove DESC if it was already called) and then limiting
to 1, which feels extremely messy, or skipping total result num - 1, which
you can't do unless you go into the query knowing the total number of
results. Cypher doesn't let you use expression results as arguments of
LIMIT or SKIP, unfortunately.

I also couldn't find any evidence that COLLECT actually slows things down.
ORDER is the big problem. If someone's match results are big enough that
they're worried about performance, calling "first" is going to be a problem
no matter what.

cheerfulstoic:
It's good to know that `collect` doesn't have much of a performance impact, but I actually care more that it makes the code more complicated.  I agree that looking at the `order` argument is somewhat messy, but I still feel like it's cleaner than generating a `collect`.  I actually just looked at ActiveRecord and it seems like that's exactly what they do.  See the following three methods (the first calls the second and the second calls the third):

http://apidock.com/rails/ActiveRecord/QueryMethods/reverse_order

http://apidock.com/rails/v4.2.1/ActiveRecord/QueryMethods/reverse_order%21

http://apidock.com/rails/v4.2.1/ActiveRecord/QueryMethods/reverse_sql_order

subvertallchris:
If only there was some way of looking at the contents of query clause objects without resorting to the private methods... ;-)

I don't see how this is any better that what we have now:
```
o.to_s.split(',').map! do |s|
  s.strip!
  s.gsub!(/\sasc\Z/, ' DESC') || s.gsub!(/\sdesc\Z/, ' ASC') || s.concat(' DESC')
end
```

PR #725, opened 2015-03-23
use neo_id in activerel create

subvertallchris:


subvertallchris:
Merging this, only that same JRuby issue.

PR #728, opened 2015-03-24
check for presence of constraint before creating

subvertallchris:


spec/shared_examples/example_app_shared.rb
```diff
@@ -96,7 +96,7 @@ def reload_model(model)
         it 'should return the first model if no conditions passed' do
           user = create_model(user_class)
           create_model(user_class)
-          user_adapter.find_first.should eq(user)
+          user_adapter.find_first.neo_id.should eq(user.neo_id)
```
subvertallchris: I don't know why this was necessary buuut... here we are.

PR #736, opened 2015-03-30
don't always make a constraint on id_property

Tenari:


subvertallchris:
Snuck this in with https://github.com/neo4jrb/neo4j/pull/738 in case you didn't see.

Issue #566, opened 2014-11-10
Unable to create relationship with non-persisted nodes


dpisarewski:
I'm wondering why neo4j tries to persist a relationship when I just assign an another node.

```
[31] pry(main)> >> m = Memo.create
=> #<Memo client_uuid: nil, created_at: Mon, 10 Nov 2014 15:14:03 +0100, description_text: nil, id: "07cf3f15-56b9-4a5e-884c-352753b0fc9a", latitude: nil, longitude: nil, made_at: nil, title: nil, updated_at: Mon, 10 Nov 2014 15:14:03 +0100>
[32] pry(main)> 
[33] pry(main)> >> s = Statement.new
=> #<Statement created_at: nil, id: nil, updated_at: nil>
[34] pry(main)> 
[35] pry(main)> >> s.memo = m
Unable to create relationship with non-persisted nodes
/Users/dieter/.rvm/gems/ruby-2.1.2@8actions/bundler/gems/neo4j-2a25adfa38f1/lib/neo4j/active_node/has_n.rb:177:in `memo='
```

The 'new' method behaves differently but also doesn't assign the associated node.

```
[57] pry(main)> >> m = Memo.create
=> #<Memo client_uuid: nil, created_at: Mon, 10 Nov 2014 15:30:41 +0100, description_text: nil, id: "23fd064d-994c-4c5a-9a8a-2435bd88bb44", latitude: nil, longitude: nil, made_at: nil, title: nil, updated_at: Mon, 10 Nov 2014 15:30:41 +0100>
[58] pry(main)> 
[59] pry(main)> >> s = Statement.new(memo: m)
=> #<Statement created_at: nil, id: nil, updated_at: nil>
[60] pry(main)> 
[61] pry(main)> >> s.memo
=> nil
```

Is it possible to make autosaving associations optional?

dpisarewski:
Related issue https://github.com/neo4jrb/neo4j/issues/647

subvertallchris:
This behavior is there very deliberate but it's reasonable to want it to be a little different when running tests. Why don't we try to separate the `def #{name}=(other_node)` portion of the `has_one` definition into its own method, then you can make a `neo4j-factory_girl` gem that overloads it with your change?

dpisarewski:
[...]

PR #764, opened 2015-04-20
option for handling non-persisted node upon assignment of an association

dpisarewski:
Solves issue #566


spec/e2e/has_one_spec.rb
```diff
@@ -28,6 +28,20 @@
       it 'raises an error when trying to create a relationship' do
         expect { unsaved_node.parent = HasOneA.create }.to raise_error(Neo4j::ActiveNode::HasN::NonPersistedNodeError)
       end
+
+      context 'with enabled auto-saving' do
+        before  { Neo4j::Config[:autosave_on_assignment] = true }
+        after   { Neo4j::Config[:autosave_on_assignment] = false }
```
cheerfulstoic: This should probably be set to whatever `Neo4j::Config[:autosave_on_assignment]` was before it was set to `true`.  We should also probably have a spec helper for this, though you don't need to do that if you don't want to ;)

subvertallchris:
It needs to look out for nodes that can't be saved because they're invalid. This should do the trick.

```
define_method("#{name}=") do |other_node|
  handle_non_persisted_node(other_node)
  validate_persisted_for_association!
  clear_association_cache
  association_query_proxy(name).replace_with(other_node)
end

def handle_non_persisted_node(other_node)
  return unless Neo4j::Config[:autosave_on_assignment]
  other_node.save && save
end
```

It should still do the check for non-persisted nodes. It has the added bonus of making it easy for users to monkey patch `handle_non_persisted_node` if they want special behavior.

dpisarewski:
Thanks for the comments, guys. @cheerfulstoic which kind of spec helper do you mean? A spec helper for neo4j specs only, or kinda Rspec DSL extension which users of the gem can use?

cheerfulstoic:
I was thinking about a neo4j(-core) gem specific helper which takes a variable name and a value and sets up `before`/`after` clauses which do what you did but making sure that the value goes back to where it started.

Something like:

```ruby
  context "something" do
    let_config(:var) { value }
  end
```

dpisarewski:
I had to extract code from `define_has_one_methods` method because of rubocop.

```
Offenses:

lib/neo4j/active_node/has_n.rb:210:7: C: Metrics/AbcSize: Assignment Branch Condition size for define_has_one_methods is too high. [18.06/18]
      def define_has_one_methods(name)
      ^^^

134 files inspected, 1 offense detected
```

cheerfulstoic:
Yeah, that seems fine.  It's definitely a bit tight in there ABC-wise because of all of the `define_method` calls.  There also seems to be some duplication of code which might be DRYed up at some point

Issue #734, opened 2015-03-29
Require relationship type in associations


cheerfulstoic:
Generally you're specifying this almost all the time and it sometimes causes problems when you don't specify it.  We should make this a required argument.  If you want to have the association match any type you would give `false`

subvertallchris:
Huge +1 to this

Issue #380, opened 2014-07-12
Chain associations / class methods on QueryProxy


cheerfulstoic:
Check out the query_experiment branch.  For good examples see the query_spec:

https://github.com/andreasronge/neo4j/blob/query_experiment/spec/e2e/query_spec.rb

The code is a bit hairy right now, but I'm really excited about this approach.  It takes over some of the conceptual ground of QuickQuery without the .qq method, it introduces a more ActiveRecordy has_many association, and it allows for calling class methods on QueryProxy objects (which are also now returned from has_many associations)

This is aping a lot of ActiveRecord, but I think it's taking it a step beyond what ActiveRecord is capable of doing.  I'm really excited about it ;)

Some things that would need to be done:
- Other associations / replacing existing associations / porting callbacks and other association features
- More specs / cleanup code

PR #767, opened 2015-04-23
Deprecating optional 'type' argument for associations for issue #734

cheerfulstoic:


cheerfulstoic:
I'm a little concerned about how many `nil`s there are for types in the specs, but I don't think the specs reflect reality in this case.  For the specs I think most of the time we don't care about types, whereas I think users most of the time will care what the types are called

subvertallchris:
What happens when `type` is nil?

cheerfulstoic:
Right now it's the same as before, it automatically chooses.  Since nil is often a default I can see something else being better.

false implies perhaps no type

:auto might conflict with a real type definition

Thoughts?

cheerfulstoic:
Thinking about it more, I actually think `nil` is still the right way to go for specifying that you want the gem to choose an association.

subvertallchris:
I like it. It makes the statement that you know the option is there and you're choosing to let the gem make the decision.

djvs:
As long as you write a migration to check them all! ;)

cheerfulstoic:
There's actually a deprecation warning in there, so the logs will show you where all of the places are (file and line number).  I actually put in the deprecation warning first and then moved all of the specs over and it didn't take long at all (maybe 10-15 minutes).

Also to clarify: My thought is to release this as deprecated with warnings for 5.0 and then remove support for the old style in 5.1

cheerfulstoic:
So we have this other problem of using `origin` which essentially makes the direction and type redundant.  In these cases the association currently would look something like this:

```ruby
has_many :in, :posts, nil, origin: :author
```

We could maybe have a `has_many_inverse_of` like this:

```ruby
has_many_inverse_of :author, :posts, options
```

Though I prefer the association name to come first.  So maybe:

```ruby
has_many_inverse :posts, :author, options
```

I still don't like that so much because the second argument isn't so clear.  Alternatively, we could just change this so that the `type` key in options is required, though it could still be set to `nil`.  I think that solves some of these problems.

Separately, as I've been thinking about this, I'm still a fan of rolling the direction into the name of the method like this:

```ruby
has_one :comment # has_one :out, :comment
belongs_to_one :comment # has_one :in, :comment
relates_to_one :comment # has_one :both, :comment

has_many :comments # has_many :out, :comments
belongs_to_many :comments # has_many :in, :comments
relates_to_many :comments # has_many :both, :comments
```

...mentioned in the thread of doom: https://github.com/neo4jrb/neo4j/issues/380

subvertallchris:
My gut reaction is to leave things alone as much as possible but I think that's mostly because I appreciate the flexibility of the existing methods. It's very easy to shape them to be whatever you want quickly without having to rewrite them completely. On the other hand, there's _also_ a certain clarity to seeing `belongs_to_many` and immediately knowing more about what it does. 

But your new methods and `has_many`/`has_one` don't seem mutually exclusive to me. You could have the new methods act as shortcuts to the old, with the old still marked `public` for those of us who still want to use them. It would be a good compromise and would make the upgrade path very easy. You're gonna need some underlying configuration methods anyway and these are already working very well.

cheerfulstoic:
Yeah, we could definitely have both types of methods, though then the question is what to use in the documentation and do we recommend a particular style?

Anyway, I'd be a lot more interested in feedback on how to implement required `type` and how to make using `origin` nice in this context.

subvertallchris:
I'm down with `nil` type if `origin` is used. Could also go with a more forceful `type: false`.

As for which to use in the documentation, I'd say we could probably lead with the new syntax and then document the old as the advanced option.

cheerfulstoic:
Just thinking about your second to last comment.  Firstly, as far as making big changes, a major release is the time to do it ;)  As a matter of fact, I also want to note that I don't really like how the direction precedes the association name.  The point of the `has_many` / `has_one` (at least from the AR world) is to form a sentence with the first argument.  So perhaps switching the association name and the direction might be nice, like this:

```ruby
has_many :comments, :out
```

I think this is also normally how methods are written anyway.  `:comments` here is primary argument, whereas the direction is more of a modifier-type argument

I think both sets of methods could be implemented, but it would require some `*args` trickery to support the two different uses of `has_many` and `has_one`, which I think I would be fine with in a deprecation situation, but not permanently.

I'm not sure what you mean by "It's very easy to shape them to be whatever you want quickly without having to rewrite them completely.".  Wouldn't it be a matter of changing either the direction in the current syntax or the method name in the proposed syntax?

cheerfulstoic:
Going to finish this up soon, but some notes so that I don't forget:

 * Instead of a required third argument, we should just require the `type` key in the options hash (unless `origin` is specified)
 * Only one of `type`, `origin`, and `rel_class` should be used at once
 * Get info about `model_class` from `ActiveRel` if `rel_class` specified

cheerfulstoic:
Re-opened with PR #795

Issue #766, opened 2015-04-22
`replace_with` method should happen within a transaction


subvertallchris:
Since it's an n+1 (or more) query, it can potentially leave the database in an inconsistent state if if the initial delete is successful but the create fails.

https://github.com/neo4jrb/neo4j/blob/5c960c12ed4f1633bdbbf9c6b2987ffe7e80a619/lib/neo4j/active_node/query/query_proxy_methods.rb#L116

cheerfulstoic:
:+1:

PR #768, opened 2015-04-23
Wraps QueryProxy association= and association << in transactions

subvertallchris:
Solves #766 
It was necessary to ensure `<<` happens inside transactions because it supports arrays. Now, all actions must complete or the entire action will roll back.


spec/e2e/has_many_spec.rb
```diff
@@ -384,4 +394,31 @@ def false_callback(_other)
       end
     end
   end
+
+  describe 'transactions' do
```
subvertallchris: This is just an extra test around associations and transactions in general. It will ensure that nothing accidentally happens outside of transactions.

PR #770, opened 2015-04-27
define attribute_method? to improve performance

subvertallchris:


subvertallchris:
This greatly improves the time required to wrap nodes when returning many objects from the database.

PR #774, opened 2015-05-01
:merge and :find_or_create need to set id property, return node

subvertallchris:


subvertallchris:
Will merge after I fix timestamps.

Issue #753, opened 2015-04-11
Change automatic label names when class is within a module?


subvertallchris:
As pointed out by @smackesey, quoting from Gitter:

"if I have this code:

```
module Animals
  class Sheep
    include Neo4j::ActiveNode
  end
end
```
then the label used to label Sheep is Animals::Sheep, and then this Cypher clause is invalid:
```
MATCH(n:Animals::Sheep)
```
and you have to do this instead:

```
MATCH(n:`Animals::Sheep`)
```
I think it would be better to use Animals_Sheep or something"

`mapped_label_name` can be used to work around it for now but it's worth exploring.

cheerfulstoic:
My comment from Gitter

"it’s a good point. Though if we replace them with something else it becomes easier to type, but it also becomes a piece of magic that users need to know about"

There was also discussion about using just `Sheep` as the label, though my gut says that namespaces exist for a reason and this will lead to conflicts.  Maybe a `_ruby_namespace` or something?  Maybe indexing would be an issue?

subvertallchris:
One vote here for a configurable option. I'd like the ability to ignore or possibly transform module names during load.

If we want to make a change, I'm happy to work on it over the weekend or early next week.

cheerfulstoic:
I'm down with a configurable option.  Maybe default to "none" (for current behavior) and "without_modules" for proposed behavior, with the ability to specify a `#call`able object which takes/returns a string.

PR #790, opened 2015-05-10
config options to handle model namespaces

subvertallchris:
Inspired by #753, this adds two config options:

#### `Neo4j::Config[:module_handling]`

In your app's config file:

```
neo4j.config.module_handling = option
```

`option` can be `:demodulize`, which will strip module names from the model's name using ActiveSupport's `demodulize` method, or a proc that will receive `model.name` as an arg and you can decorate it however you want. Anything else will just get ignored.

#### `Neo4j::Config[:association_model_namespace]`

In config:

```
neo4j.config.association_model_namespace = 'ModuleName'
```

This will prefix the models associations look for when `model_class` is omitted with whatever value you give it. It's useful if all of your models are part of the same module, just set once and it'll take care of the rest.

Examples are in specs. I need to test that it can load correctly when these options are set.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ script:
   - "export CODECLIMATE_REPO_TOKEN=04853df625409de0a0f4e9126aee11bbee3428c81c20c27df6e6ab5c60bff2c8 && export JRUBY_OPTS='-X+O -J-Djruby.launch.inproc=false' && bundle exec rake neo4j:install[$NEO4J_VERSION] neo4j:start default --trace"
 language: ruby
 rvm:
-  - ruby-2.2-head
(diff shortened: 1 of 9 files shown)


cheerfulstoic:
Awesome, thanks ;)  Could you document in `docs/Configuration.rst`?

subvertallchris:
Definitely. There's still a tiny bit more work to do... Wrapping works but needs a tweak. I'll have it fixed and documented in a few mins.

cheerfulstoic:
:+1: ;)

cheerfulstoic:
Oh, also, to generate the docs run `rake docs` and commit the results under `docs`

PR #798, opened 2015-05-14
improve support for match_to with an array of nodes

subvertallchris:
This ensures that if you give `match_to` an array of nodes, it'll use all of their neo ids. It has the nice bonus of accepting an array of wrapped and/or unwrapped nodes if you find yourself in that situation.


cheerfulstoic:
:+1:

PR #802, opened 2015-05-17
Even faster wrap

subvertallchris:
This removes a loathed `symbolize_keys!` that was being called during every single node's wrap by using info the class should already have in its declared property manager. When loading a large group of nodes, it's something approaching 5% faster. It makes no difference whatsoever when loading just a few nodes but do it with a couple hundred at a time and it adds up. Loading unwrapped nodes is still 2x faster but I'm closing the gap.

It also removes a `to_s` that was occurring in Neo4j::Config when looking up the classname property name. That one isn't gonna change the world but it was easy, so might as well.


cheerfulstoic:
Awesome, thanks!  Excited about having a 5.0 rc! ;)

Issue #808, opened 2015-05-19
`count` on a query with `limit` should rewrite the query to use `with`


subvertallchris:
Another bug report from work.

Say you have this:

```
Student.all.friends.limit(5).count
```

You want to find all friends of students, max 5 results, and you want to know how many you actually have. At most, you should get 5, but if there are fewer rels, you should get that number. Cypher does not respect `COUNT` and `LIMIT`, so this:

```
MATCH (s:Student)-[r:FRIENDS_WITH]->(f) RETURN COUNT(f) LIMIT 3
```

...will return the TOTAL number of rels. If you instead write it like this:

```
MATCH (s:Student)-[r:FRIENDS_WITH]->(f) WITH f RETURN COUNT(f)
```

...you'll get what you're looking for.

I think the solution is to just modify `count` to handle this automatically. It's an easy fix.

cheerfulstoic:
Good catch!

PR #810, opened 2015-05-20
inserts a `with` into `count` when `limit` has been called

subvertallchris:


subvertallchris:
Catching the beat from dat #808.

PR #824, opened 2015-06-05
Fix missing unique option

olance:
PR originating from [this discussion](https://github.com/neo4jrb/neo4j/commit/a88eb055cb531c1015c5cdd42c2559f64753248e#diff-17f68f769f3e31e7e8d67a4f4e2e66bdR322) on a missing option in the valid options array for `has_n` relationships.

Its purpose is to fix two problems:
**Problem 1:** The `:unique` option is (incorrectly) not considered a valid option anymore on `has_many`/`has_one` relationships.

**Problem 2:** The tests have not failed in spite of this regression.

## Fixing problem 1
This was just as easy as adding `:unique` to the array of valid options

## Fixing problem 2
This was a bit more tricky, so I hope my solution will be suitable.

The tests haven't picked up this regression because they create associations using `Neo4j::ActiveNode::HasN::Association.new` directly, and options validations are performed at a higher level in `Neo4j::ActiveNode::HasN::ClassMethods#has_many` and `#has_one`.

So I have moved these validations down at the `Neo4j::ActiveNode::HasN::Association` level, which has had some consequences:

* `:type` is now required in the options hash of `Neo4j::ActiveNode::HasN::Association#initialize`. I changed the default value of the `options` argument from `{}` to `{ type: nil }` to keep it easy to create an association that way without any option. However I had to change some specs in `association_specs` to make sure they include a `:type` key in the options they use.

* The error messages from the validation are quite helpful as they indicate the name of the failing association and its owning class. This was possible when checking within `Neo4j::ActiveNode::HasN::ClassMethods` because the module is included inside the model class that would declare the association. As this is not the case with `Neo4j::ActiveNode::HasN::Association`, I have kept the behavior like so:  

  * When an option validation error occurs, `Neo4j::ActiveNode::HasN::Association` will raise the same `ArgumentError` exceptions as before but without the "(ClassName#association_name)" part
  * Within `Neo4j::ActiveNode::HasN::ClassMethods#build_association` (which actually creates the `HasN::Association` instance), `ArgumentError` exceptions are caught and re-raised with the same message, enhanced with the "(ClassName#association_name)" that we are able to get at this level.

* There are tests checking the content of the above error messages. These tests are run against the `has_many`/`has_one` methods. Maybe they should move to the association specs?

Fortunately, the end result is that no other tests have failed due to these changes but the ones that were using the `:unique` option when it was not considered valid anymore.

So I do hope these changes will make it to the codebase :)

### DIFF
--- a/lib/neo4j/active_node/has_n.rb
+++ b/lib/neo4j/active_node/has_n.rb
@@ -292,7 +292,6 @@ module Neo4j::ActiveNode
       #       (note that the ``:destroy_orphans`` option is known to be "very metal".  Caution advised)
       #
       def has_many(direction, name, options = {}) # rubocop:disable Style/PredicateName
-        validate_association_options!(name, options)
         name = name.to_sym
         build_association(:has_many, direction, name, options)
 
@@ -310,7 +309,6 @@ module Neo4j::ActiveNode
       # not specified here
       #
       def has_one(direction, name, options = {}) # rubocop:disable Style/PredicateName
-        validate_association_options!(name, options)
         name = name.to_sym
         build_association(:has_one, direction, name, options)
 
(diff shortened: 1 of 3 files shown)


cheerfulstoic:
These are great changes, thank you so much!  I particularly like how you catch and re-raise the exception to preserve a separation of concerns.  We have a few variables being passed into the `QueryProxy` class to be able to keep track of state like that, but perhaps we could find similar solution to de-couple them.

A couple of items:

 * The specs are passing, but `rubocop` is failing which is causing Travis CI to fail.  Could you run `rubocop -aD` locally to automatically clean up what can be automatically cleaned and to identify the rest of the code style items which need to be fixed?
 * I'm a bit tempted to go to extremes in the place where you re-raise the error.  My first thought was to only re-raise that specific error by checking for the message.  Then I realized that having the class and association name could be useful for any messages, so I wondered if this could be not just ArgumentErrors but all errors.  What do you think?
 * Speaking of that bit I think rubocop will bring this up but according to the Ruby style guide `fail` should be used when you're failing for the first time and `raise` should be used when you're rescuing and re-raising.

Thanks again for the help!  With the above changes I'd be happy to merge them and cut a new rc gem

olance:
Thanks, I'm glad you like the idea :)
As for catching everything I think it could be useful indeed. It's doing nothing more than adding information so I guess there's nothing wrong to it.

I will review rubocop errors and push again ASAP!

subvertallchris:
Very cool stuff! Sorry for not commenting yesterday. I was out in the
country and expected to have internet on my phone. Clearly, I was wrong.
I'll keep an eye out for the next PR and one of us will do the merge once
Travis gives the all-clear.

olance:
@subvertallchris No worries ;)

@cheerfulstoic I've run Rubocop and it did change the `fail` to raise automatically ;)

cheerfulstoic:
Done and released as 5.0.0.rc.3.  Thanks!

olance:
Great :)
Thank you for the quick release! ^_^

PR #825, opened 2015-06-07
QP without label

subvertallchris:
This adds a `labels` option to association methods. `labels: false` will prevent QP from using labels when chaining, as you might guess.

```ruby
student.lessons(:l, :r, labels: false).students(:s, :sr, labels: false).to_cypher
# "MATCH student43918, student43918-[r:`ENROLLED_IN`]->(l), l<-[sr:`ENROLLED_IN`]-(s) WHERE (ID(student43918) = {ID_student43918})" 
```

Cutting the label filter on matches that touch thousands of nodes will cut the number of hits substantially, which may result in big performance gains if you have a big, busy database and your Cypher query touches a lot of nodes.


subvertallchris:
In practice, this has pretty much no impact on performance within Ruby unless you're using `count` or something that doesn't return a large number of nodes. The improved query speed is negated by the gem's slow node-wrapping process.

PR #831, opened 2015-06-12
fixes handling of Time type, undoes silent change to DateTime

subvertallchris:
Long, long ago, one of my first commits to the gem was to add support for the `Time` property type. `ActiveAttr`, which handles some of the type conversion within the gem, does not respect it, but we needed it. Without knowing how else to approach it, I set it to just change it to change properties to DateTime. The two types respond to a lot of the same methods so I thought it might be an ok fix.

Two-ish years later, I think that's a lousy workaround and know how to fix it in a way that works a bit better. So... that's what this is.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - QueryProxy associations accept `labels: false` option to prevent generated Cypher from using labels.
 
+### Changed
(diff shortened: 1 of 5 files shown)


cheerfulstoic:
Yay!  Nice!

cheerfulstoic:
This is a pretty good overview of the differences, by the way:

https://gist.github.com/pixeltrix/e2298822dd89d854444b

Issue #835, opened 2015-06-17
`find_or_create_by` with associations should work


subvertallchris:
It'd be nice to do this:

```
math.students.find_or_create_by(name: 'Chris', age: 31)
```

If the pattern doesn't exist, it should create the node and then associate it with `math`.

It's shorthand for:

```
student = math.students.where(name: 'Chris', age: 31).first
return student unless student.nil?
student = Student.create(name: 'Chris', age: 31)
math.students << student
student
```

Stretch goal:

It should use the ActiveRel model for rel creation if one is defined.

subvertallchris:
Merged, closing.

PR #836, opened 2015-06-17
add find_or_create_by method + tests for query_proxy, close issue #835

rebecca-eakins:
@subvertallchris guided me through this update. Thank you!


cheerfulstoic:
This is great, thanks!  I don't have time to look because I have a few deadlines, but I just wanted to mention that we should probably make this a 5.1 feature.  Also, 5.1 shouldn't be too far in the future.  I think if we don't have a regular release schedule than it's tempting to put lots of last minute features into the RC.

Y'all may have already considered this, but I'm just putting it out theer

PR #838, opened 2015-06-18
Default values for model_class when using ActiveRel class

olance:
As discussed on Gitter, this PR makes the `model_class` option a bit "smarter" by taking an `ActiveRel` class into account when one is used on the association.

When `RelClass._to_class == :any`, `model_class` is defaulted to `false` to avoid having to declare it explicitly (without it, an NameError exception will be raised when calling the association method).

When `RelClass._to_class` is any other value and the `:model_class` option has not been set, it is defaulted to the value of `RelClass._to_class`.

Note: When #837 gets merged, the code should be changed not to use `#_to_class` anymore.

### DIFF
--- a/lib/neo4j/active_node/has_n/association.rb
+++ b/lib/neo4j/active_node/has_n/association.rb
@@ -20,11 +20,10 @@ module Neo4j
         def derive_model_class
           return @model_class unless @model_class.nil?
 
-          unless relationship_class.nil?
-            # TODO: change `#_to_class` to `#to_class` when #837 is merged
-            return false if relationship_class._to_class.to_s.to_sym == :any
-            relationship_class._to_class
-          end
+          return nil if relationship_class.nil?
+          # TODO: change `#_to_class` to `#to_class` when #837 is merged
+          return false if relationship_class._to_class.to_s.to_sym == :any
+          relationship_class._to_class
         end
 
         def target_class_option(model_class)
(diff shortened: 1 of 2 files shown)


olance:
ah, forgot to run Rubocop...

olance:
@subvertallchris rebased after merge of #837, fixed errors and ran Rubocop... all should be good :)

lib/neo4j/active_node/has_n/association.rb
```diff
@@ -17,6 +17,14 @@ def initialize(type, direction, name, options = {type: nil})
           apply_vars_from_options(options)
         end
 
+        def derive_model_class
+          return @model_class unless @model_class.nil?
+
+          return nil if relationship_class.nil?
+          return false if relationship_class.to_class.to_s.to_sym == :any
```
subvertallchris: Should this always be `to_class`? Should the direction change based on the current association's direction?

subvertallchris:
Shot you a PR on your branch that fixes that problem. We can merge this after you merge mine or implement your own way. Great stuff!

cheerfulstoic:
Very awesome, thanks!

lib/neo4j/active_node/has_n/association.rb
olance: eh, you're right I had totally forgotten about directions, sorry!

olance:
@cheerfulstoic my pleasure :)
@subvertallchris merged your PR so you can merge mine when Travis is happy ^^

Issue #717, opened 2015-03-12
Node must be of the association's class when model is specified?


arikan:
While associating one model to another via the simplest declared connection, I've been getting the error below since we upgraded to Neo4j 2.2, Neo4jrb 4.1, Rails 4.2.

```ruby 
class User
  include Neo4j::ActiveNode
  has_one :out, :invitation

class Invitation
  include Neo4j::ActiveNode
  has_one :in, :sender, model_class: User, after: :decrement_sender_count
  has_one :out, :recipient, model_class: User   

@user = User.find(params[:id])
@invitation = Invitation.find_by(:token => params[:invitation][:token])
puts @invitation
@user.invitation = @invitation
```
```
ArgumentError - Node must be of the association's class when model is specified:
```

Invitation object is clearly there as puts `#<Invitation:0x007ff1a9a1acc8>`

The last line tries to make the association with the cypher below, then throws the error above.

> Class#invitation#invitation 13ms MATCH (user18376:`User`), user18376-[rel1]->(result_invitation:`Invitation`) WHERE (ID(user18376) = {ID_user18376}) DELETE rel1 | {:ID_user18376=>18376}

This is weird: `Class#invitation#invitation` I don't get what's happening here? Any thoughts?

subvertallchris:
Try doing a bundle update on neo4j and neo4j-core if you haven't already,
there's been a lot of refactoring going on lately. Brian will probably have
a better idea than me but I'll check it out tomorrow, too, to see if I can
reproduce it.

arikan:
Thanks, just attempted to do an update, already have the latest.
```
Using neo4j-core 4.0.1 from https://github.com/neo4jrb/neo4j-core (at master)
Using neo4j 4.1.2 from https://github.com/neo4jrb/neo4j (at master)
```

Currently looking into it, let you know if I find more hints.

arikan:
Trying to debug this problem, but no chance. One thing clear however, that, the error is caused by the change in the declared relation naming with Neo4jrb 4.1.

I tried the "legacy" as well as the new relation naming scheme, it both throws the same error.  
```rb
config.neo4j.transform_rel_type = :legacy
```

```
ArgumentError - Node must be of the association's class when model is specified:
```

One interesting case: If I restart the server, it works for the first time, then keeps throwing the error above for the rest of the attempts.

subvertallchris:
[...]

PR #839, opened 2015-06-19
use labels instead of is_a? to determine whether nodes are of appropriate type

subvertallchris:
This solves https://github.com/neo4jrb/neo4j/issues/717.

Rails `reload!` method was bumping heads with the `@model` instance variable in QueryProxy. I can't totally pin down why, will need to look at it more another time. In the meantime, this skips around the problem by just using labels to determine whether nodes are of the appropriate type.

No new tests, this is too specific to Rails and it doesn't add anything new, it ends up in the same place using existing methods.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased][unreleased]
 - Nothing yet, placeholder
 
+### Fixed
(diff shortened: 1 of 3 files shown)


cheerfulstoic:
Oh, interesting, that probably makes sense.  If I had to guess, `reload!` is probably destroying the constant and then creating it over again which would explain why it's two different objects

PR #853, opened 2015-06-29
Queries with variable length relationships

olance:
The goal of this PR is to make it possible to query nodes with variable-length relationships from within a QueryProxy query.

The intended usage is as follow:
```ruby
# Find up to 3rd-degree friends
person.friends(:p, :r, rel_length: (1..3))
# Find friends of friends only:
person.friends(:p, :r, rel_length: 2)

# Get the "chain of command" from an employee up to its big boss
employee.manager(nil, nil, rel_length: :any)
# To include the employee in the chain of command:
employee.manager(nil, nil, rel_length: :none_or_more)
```

I am not 100% satisfied with the `:none_or_more` symbol for the `0..` variable-length qualifier, you might have something better in mind?

As one may be interested in querying variable-length relationships without naming the node or the relationship, I'd be tempted to make it possible to use the association "getters" without their first two arguments:
```ruby
# Find friends of friends:
person.friends(rel_length: 2)
```
If you're up for it too, I can add the necessary changes to this PR!

Also, the new feature is yet to be documented... If you're willing to merge the PR, let me know where I should document this and in what form, and I'll add it to the PR before merging.


cheerfulstoic:
This seems really cool!  Originally when I was thinking about this it was as options for `has_(one|many)`.  Do you think that there's a need to change this for each call, do you think fixing the options in the association would be sufficient, or do you think both are required?

I think I would in general be good with checking to see if the first argument to an association call is a Hash to treat that as options as assume `nil` for the node/rel variables.  What do you think @subvertallchris ?

As for the docs, there is a `docs` folder in the project which generates [http://neo4jrb.readthedocs.org].  It would also be good to make sure the various public/facing methods are documented with `YARD` comment syntax.  That also gets put on the readthedocs site and you could link to it from the tutorial-style docs (See the "See also" comment at the bottom of this section for how to do that: http://neo4jrb.readthedocs.org/en/latest/ActiveNode.html#associations)

olance:
Noted for the docs!

I can see why you could like declaring the length in the options for `has_(one|many)`, although I think this is really too restrictive. I hadn't even thought about this because my use case (hierarchy trees, among other things) would possibly never use the same ranges twice ^^

All in all, I'd say it's not a good idea to have this as an option on the definition of the relationships (but that's probably very subjective), however it could probably be a good fit for scopes? Using a default scope (is it possible? can't remember) would then have the same result as defining the length on the relationship, but would not be as much a constraint.

subvertallchris:
I'm a fan of having the option go in the association call, not the the association's definition, for all the reasons you gave. The few times I've needed variable length matches, it was using an existing association that I wouldn't have wanted to redefine.

+1 for checking if the first argument is a hash and if so, using that for options. When I added the `labels` option, I thought I wouldn't mind being forced to always define node and rel ids; in practice, it's annoying and clutters up the code when you have to give arguments that you won't use later.

@olance, this is awesome, btw!

olance:
thanks @subvertallchris :)

@cheerfulstoic what do you think? I could investigate the scope thing a bit, would that satisfy your own view of the feature?

cheerfulstoic:
@subvertallchris Do you think it makes sense to support both?  I haven't done anything with variable lengths in the gem, but it seems like it would be nice not to have to specify the min/max each time you call the association in some cases.

Also I forgot that I wanted to talk about the `none_or_more`.  There may very well be times that you want `1..`, `2..`, etc...  So what about the following:

```ruby
# Use an Integer
path_length: 2

# Use a Range
path_length: (1..5)

# Use a Hash
path_length: {min: 1}
path_length: {max: 2}
```

Do you think it would be easy to add whatever options we end up choosing into association definition as well as the association method call?

(I'm agreed on the excellentness, by the way)

cheerfulstoic:
A default scope (I also don't remember if that's currently possible ;) would set the length for the whole model, I think.  Maybe we're thinking of different things, could you give code examples?

subvertallchris:
I think scopes would take care of it, we can wait to add support to
association definitions until we get some requests for it. Keep it simple
for now, we can always expand.

olance:
ah ah thanks, I'm flattered ^^

For the moment, `1..` is supported via the `:any` symbol and `X..` is supported via ranges, using: `(X..Float::INFINITY)`, which I agree is not very practical.  
Hashes are an interesting alternative :)

You're right about the scopes, that's not an option. What I was thinking was to have some ways to define sorts of named scopes, such as `friends_of_friends`, which would make use of the `friends` relationship with a fixed length. In this case I would not want to specify the length for each call indeed.  
However that's as easy as:

```ruby
def friends_of_friends
  self.friends(rel_length: 2)
end
```

So do we really need more?

olance:
@subvertallchris I think semantically speaking, this would not really be "scopes", are those would not add constraints on the queried node but traverse the graph from a (or many) given point(s).

But I agree with you, I think we should keep it simple.  
In Rails, they kinda "reverted" to simpler scopes when they documented that a named scope with arguments should in fact just be declared as a class method, instead of using the `scope` method and a lambda with arguments.  
That's less features supported (kind of) but in the end the code is cleaner and easier to read!

subvertallchris:
A `scope` method and a lambda with arguments was what I had in mind, but, yes, the point is the same: we can define a query around an existing association that makes it more specific.

cheerfulstoic:
The use case I'm thinking of would have the length defined in the association, but Chris is right, let's keep it simple and do one thing at a time.  I've been waiting a year or two to implement that anyway ;)

I see what you're saying now.  I had a lot of other comments that I just deleted because I need to think about them more.  Let's just go with this for now so we can keep moving forward ;).  Thanks!

olance:
@cheerfulstoic @subvertallchris I have updated the PR with following changes:

* `:none_or_more` Symbol value is not allowed anymore, it felt weird.
* `:rel_length` now accepts a Hash with `:min` and/or `:max` values. Using this, it's easy enough to get the same result as what `:none_or_more` provided: `rel_length: {min: 0}`
* `has_(one|many)` "getter" methods can now omit the node/rel names and pass a single Hash argument to be used as the options hash

Also, please note that as I said in a commit, 	I have taken the liberty to increase Rubocop's maximum allowed Cyclomatic complexity to 7 (default is 6), because I feel there's no satisfying way to reduce the Cyclomatic complexity of `AssociationCypherMethods#validate_rel_length` (it requires 4 `when` to check for allowed types and 2 additional `if`/`unless` for conditions on Symbol and Fixnum types).  
I could create two `validate_rel_length_fixnum` and `validate_rel_length_symbol` but they would be so trivial I think it would bloat the code more than make it clearer.

If you feel I'd should rather do that, just say the word ^^

Next step: docs!

subvertallchris:
This is great!

olance:
:)

I have added some docs in the "Querying" part of the tutorial-style docs, however I'm not sure where to document the option in the code itself. There's no real "public-facing" method as all `has_(one|many)` association methods are created dynamically.  
In fact, I haven't found any docs for the `:labels` option either!

I'll quote YARD's author on this [issue](https://github.com/lsegal/yard/issues/430#issuecomment-3237631):

> dynamic method definitions are inherently undocumentable in this form, and not recommended

So I guess there's no place for in-code YARD documentation about the options accepted by the associations getters?

I'll need to fix the `@option` doc I had started in the `QueryProxy` class though, it can still be useful.

cheerfulstoic:
Wow, this is great, thanks!  A few things:

 * You're right about the YARD docs.  Good point.  We're good as long as something is in the tutorial docs
 * With regard to the cyclomatic complexity I would suggest using a rubocop comment to disable that particular cop validation from checking that method.  See [this section of the README](https://github.com/bbatsov/rubocop#disabling-cops-within-source-code)
 * I don't know if this will help with cyclomatic complexity, but you're calling `fail` in a lot of places with the same error message so I think you could refactor doing something like:

```ruby
  def validate_rel_length!(length)
    message = validate_rel_length_message(length)

    fail(ArgumentError, "Invalid value for rel_length (#{length.inspect}): #{message}") if message

    true
  end

  def validate_rel_length_message(length)
    case length
    when Symbol then
      "expecting one of #{VALID_REL_LENGTH_SYMBOLS.keys.join(', ')})" if !VALID_REL_LENGTH_SYMBOLS.key?(length)
    when Fixnum then 'cannot be negative' if length < 0
    when Range
      if length.begin > length.end
        'cannot be a decreasing Range'
      elsif length.begin < 0
        'cannot include negative values'
      end
    when Hash
      'Hash keys should be a subset of [:min, :max]' if (length.keys & [:min, :max]) != length.keys
    else
      'should be a Symbol, Fixnum, Range or Hash'
    end
  end
```

There are a number of things going on here ;)  Feel free to disagree with any of them

 * Changed the method name to be `bang!` syntax to indicate that it may return an exception
 * We define one place where `fail` is called which simplifies our code but also gives us output which is always the same and which always shows the inspected value which was invalid
 * I changed the `unless` to an `if !`.  It's a bit of a pet peeve of mine and I've been part of a [long standing discussion](https://github.com/bbatsov/ruby-style-guide/issues/329) on the Ruby style guide repo about it ;)
 * I changed the validation of the `Hash` to exclude any invalid keys as well (by checking for subsetness)
 * You may need to factor out parts of the `case`, but I think that now that there aren't any method calls that the complexity is perhaps lower

olance:
You are mistaking the cyclomatic complexity with the ABC size of the method :)
(I've got no wish to sound pedantic about it, I learned about them because of Neo4j.rb using Rubocop -- which I set up in all my new projects since ^_^)

The [cyclomatic complexity](www.rubydoc.info/github/bbatsov/rubocop/Rubocop/Cop/Style/CyclomaticComplexity) counts the number of "decision points" in a method, that is: how many `if`/`unless`, `when`, `and`/`or` and some other (full list in the docs); whereas [ABC size](http://docs.codeclimate.com/article/148-glossary-abc-metric) counts **A**ssignments, **B**ranches and **C**onditionals and is another form of complexity indicator (perhaps more global/complete). "Branches" here refer to "change of context", so any method call increases the counter. So reducing the number of calls to `fail` actually decreases the ABC size of the method ^^

However I can already foresee an "offense" in your `validate_rel_length_message` method: Cyclomatic complexity is 9, higher than the current code. ABC size is 12.05 though, if I am not mistaken, so that's ok.

So, all this being said, I like the way you refactored the `fail` messages and I'll just keep the cyclomatic complexity lower by extracting the range-related checks to another method just as it's already been done.  
I am not much of a sectarian when it comes to `unless`/`if !`, so I'll keep your code that way, as nothing strikes me as being especially difficult to understand.

I must say I love those kind of discussions :)

cheerfulstoic:
We love these kind of discussions as well ;)

Thanks for the clarification, I normally just try to keep refactoring to make things simpler under rubocop stops yelling at me, but I knew the difference between them in the back of my head ;)

Otherwise I'm happy with the PR, thanks again!

subvertallchris:
Indeed. As evidence of love discussion, I present exhibit A: https://github.com/neo4jrb/neo4j/issues/380.

olance:
Well that's a lengthy discussion :P 
I haven't read it all, although it's interesting to see how ActiveQuery was born ^^

olance:
@cheerfulstoic / @subvertallchris The PR should be good to merge!

I have updated doc files I hadn't committed and rebased my branch on top of the master branch.

cheerfulstoic:
Yeah, that was a long discussion ;)  Crazy to think that I've been working on this project for over a year!

I'm happy if @subvertallchris is

subvertallchris:
Looks great to me as long as Travis doesn't have anything to say about it.

olance:
Ah, in the end, it always come down to Travis!
Wait & see ^^

olance:
:metal:

subvertallchris:
:metal: !?
This changes everything!

olance:
?

subvertallchris:
Haha, I'm just excited that there's a :metal:.

olance:
Okay, you've got me worried for a minute ^_^

PR #854, opened 2015-07-02
`where` can use model type converters, more type conversion logic

subvertallchris:
Part A makes QueryProxy's `where` method inspect the in-scope model for a matching key. If one is found, it will send the value given to the type converter. It means something like this is now valid:

```
class Student
  include Neo4j::ActiveNode
  property :created_at, type: DateTime
end

# Will run the value through `created_at`'s converter
student.where(created_at: DateTime.now - 10).first
```

If you match against an undeclared property, that's on you, you should convert it before performing the match. If you need custom conversion, there's a process for adding custom converters already in place.

"But Chris," you might be saying, "I often take a timestamp from a controller and immediately perform a match on a declared property. I don't want to have to convert it to a DateTime so it can just convert it back to an integer!" If so, you're in luck!

Part B makes type converters a bit smarter. Converters can now have a `db_type` method that returns the expected output type of conversion for interaction with Neo4j. If your `where` is an instance of this class, it will bypass conversion and just spit it back out.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,7 @@ source 'http://rubygems.org'
 
 gemspec
 
-# gem 'neo4j-core', github: 'neo4jrb/neo4j-core', branch: 'master'
+gem 'neo4j-core', github: 'neo4jrb/neo4j-core', branch: 'master'
 # gem 'neo4j-core', path: '../neo4j-core'
 
 # gem 'active_attr', github: 'neo4jrb/active_attr', branch: 'performance'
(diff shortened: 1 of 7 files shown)


subvertallchris:
ping @olance

subvertallchris:
Some Embedded tests are broken upstream but the code they're testing is fine and unrelated to what I'm doing here, so I'm skipping them for now to keep moving.. We can uncomment when they're fixed.

olance:
Great @subvertallchris, I'll have a thorough look at this ASAP!

olance:
Hey @subvertallchris sorry for not chiming in earlier!
That's quite a simple bunch of changes, I guess there's nothing to say but "good job"! :)

I like the "part B" part, but knowing that you're the one who actually told me you usually get a timestamp from a controller, I'm afraid you're a bit schizophrenic :P

PR #855, opened 2015-07-03
methods to clear cached classes after `reload!` is called

subvertallchris:
This appears to correct the root cause of #717 and possibly https://github.com/neo4jrb/devise-neo4j/issues/16. It uses the `before_remove_const` method fired by ActiveSupport's `reload!` to wipe cached references to classes. It's possible others are still lurking but this will be a start.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,7 @@ source 'http://rubygems.org'
 
 gemspec
 
-# gem 'neo4j-core', github: 'neo4jrb/neo4j-core', branch: 'master'
+gem 'neo4j-core', github: 'neo4jrb/neo4j-core', branch: 'master'
 # gem 'neo4j-core', path: '../neo4j-core'
 
 # gem 'active_attr', github: 'neo4jrb/active_attr', branch: 'performance'
--- a/lib/neo4j/active_node/has_n/association.rb
+++ b/lib/neo4j/active_node/has_n/association.rb
@@ -20,6 +20,7 @@ module Neo4j
         end
 
         def derive_model_class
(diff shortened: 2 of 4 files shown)


jacobwgillespie:
Any news on when this PR might get released on Rubygems?  Right now, using `reload!` in the console isn't working with the current version (5.0.14) and it looks like it's due to this not yet being released.

cheerfulstoic:
We definitely want to release a 5.1 soon.  I've been producing some screencasts for the gems and the latest screencasts require some features in master, so those will need to be made publicly accessible.

cheerfulstoic:
To give a clearer answer: I expect we'll at least do an RC that'll be in the next week or two, one way or another.

jacobwgillespie:
Ah, gotcha, so this is going out as a point release and not a patch.  Thank you for the work, I'll watch for the RC!

cheerfulstoic:
Ah, right.  I'm not sure if @subvertallchris planned to do a patch (I know he's been busy with work lately), but obviously it will go out with a point release ;)  Thanks for the patience

subvertallchris:
I think a patch release would be a good idea. A few people are locked to master because they need what it provides, I think it would help them out. I'll take care of it when I have a few minutes today and will post here when it's done.

jacobwgillespie:
:+1: Thank you!

subvertallchris:
Just released 5.0.15 with this. Let us know if you have any trouble.

cheerfulstoic:
@subvertallchris: CHANGELOG? ;)

jacobwgillespie:
It works great - thanks!  :tada:

PR #878, opened 2015-07-21
Support outputting pretty cypher queries

cheerfulstoic:
With `pretty_logged_cypher_queries` config option


cheerfulstoic:
@subvertallchris Could you try this out when you get a chance?  You should just need to use the `pretty_cypher` branches for both `neo4j` and `neo4j-core` as well as setting `config.neo4j.pretty_logged_cypher_queries = true`

You might also need to set `config.neo4j.log_cypher_queries = true`

PR #879, opened 2015-07-21
Association ids

cheerfulstoic:
Check the specs


cheerfulstoic:
Oops, didn't mean to branch this off the `pretty_cypher` branch.  Oh well ;)

subvertallchris:
What would you think about only defining these methods if they aren't already defined in the class?

cheerfulstoic:
I'm not sure I understand...  They should be defined when the `include Neo4j::ActiveNode` happens which means that anything defined after that should overwrite.  Right?

subvertallchris:
It'd happen when they call `has_many`, so they're fine if they do that first, but if they include a module that includes the methods, they'd be overwritten.

cheerfulstoic:
It looks like in ActiveRecord that it's sometimes desirable to be able to call `super` to extend the functionality.  I just realized, though that this would still be fine to not define the methods if they already exist.  Sorry if I don't make sense, I only got a couple of hours of airplane sleep last night (I'm surprisingly functional now, but maybe that's because Costa Rica time is three hours behind me)

http://apidock.com/rails/ActiveRecord/Associations/ClassMethods

It's worth noting all of the methods that you can call on associations in those docs.  We might need to modify some (like `create` for example we've talked about what methods should create relationships and which should create nodes)

Issue #889, opened 2015-07-28
Should we break before every association in a proxy chain?


cheerfulstoic:
So I have the following Ruby code:

```ruby
  user.created_assets(:asset).optional(:categories, :category).pluck('asset.title', 'collect(category.name)')
```

Which generates the following Cypher:

```cypher
  MATCH
    user4865,
    user4865-[rel1:`CREATED`]->(asset:`Asset`)
  OPTIONAL MATCH asset-[has_category_rel:`HAS_CATEGORY`]->(category:`Category`)
  WHERE (ID(user4865) = {ID_user4865})
  RETURN
    asset.title,
    collect(category.name)
```

Neo4j actually returns all assets and their categories in this case rather than limiting by the user.  I'm not sure if this is a bug in Neo4j, but maybe we should `break` between each association?  I've made the two small changes and it seems to fix the problem

cheerfulstoic:
Closing now that we have #890 890

PR #890, opened 2015-07-29
Always break between associations to fix issue when OPTIONAL MATCHes get between MATCHes and their WHEREs (#889)

cheerfulstoic:
For issue https://github.com/neo4jrb/neo4j/issues/889


cheerfulstoic:
Grrr, this is frustrating.  So you can't have a `MATCH` after an `OPTIONAL MATCH` apparently, but you need to group the right `WHERE` clauses... I think that Neo4j should just let you group it all together and then put a `WHERE` after both and apply the `WHERE` to the `MATCH` even if there's an `OPTIONAL MATCH` in between...

PR #892, opened 2015-07-30
Id prop index calls

subvertallchris:
Just gonna copy and paste pieces from a conversation with @cheerfulstoic just now:

"I’ve gotta merge something that changes some behavior around uuids and id_property. It was creating a constraint on uuid every time the server started, which wasn’t a problem unless you try setting a custom generation method for uuid and want an index instead of a constraint.

So you’d do this:
```
class MyModel
  include Neo4j::ActiveNode
  id_property :uuid, auto: :uuid, constraint: false
  index :uuid
end
```
Or maybe you’d use a custom method, it doesn’t matter. What happens now is session starts, default behavior creates a constraint, then the index call drops the constraint and replaces it with an index. Next time the app starts, it repeats but it sees you already have an index, so it drops it to create the constraint, then drops the constraint to create the index.
It’s the ciiiiircle of liiiiife..."

Brian described it as the "circle of death of performance," which is both true and metal. :metal:

The change is two-fold:

* Allow manual calls to `id_property` register as pending events.
* Make `id_property` aware of Neo4j session status. It will give you "last-call-wins" behavior.

A few tests were updated to help confirm some of this behavior. It's a little tricky because it relies on interaction with new sessions, so there's a new helper method in `spec_helper` to make it a little less painful.

### DIFF
--- a/lib/neo4j/active_node.rb
+++ b/lib/neo4j/active_node.rb
@@ -76,7 +76,7 @@ module Neo4j
 
       def self.inherit_id_property(other)
         Neo4j::Session.on_session_available do |_|
-          return unless self.id_property?
+          next if other.manual_id_property? || !self.id_property?
           id_prop = self.id_property_info
           conf = id_prop[:type].empty? ? {auto: :uuid} : id_prop[:type]
           other.id_property id_prop[:name], conf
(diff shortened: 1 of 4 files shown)


cheerfulstoic:
:metal:

PR #895, opened 2015-08-02
Object allocation fixes 080215

subvertallchris:
Skipping those hashes during init can save object allocations and speed things up. Moving the escape logic out of `initialize` and into the methods being called helps keep things a bit cleaner.


cheerfulstoic:
This might give you some more ideas:

https://github.com/rails/rails/pull/21057

lib/neo4j/shared/property.rb
```diff
@@ -17,15 +17,14 @@ class MultiparameterAssignmentError < StandardError; end
 
     # TODO: Remove the commented :super entirely once this code is part of a release.
     # It calls an init method in active_attr that has a very negative impact on performance.
-    def initialize(attributes = {}, _options = nil)
-      attributes = process_attributes(attributes) unless attributes.empty?
+    def initialize(attributes = nil, _options = nil)
```
cheerfulstoic: Why change this to `nil`?  I could see changing the `_options` to default to `{}` (also it seems like `_options` isn't used...)

lib/neo4j/active_rel/property.rb
```diff
@@ -24,6 +24,11 @@ def initialize(attributes = {}, options = {})
       send_props(@relationship_props) unless @relationship_props.nil?
     end
 
+    # Satisfies a requirement of Neo4j::Shared::Initialize
+    def default_properties=(properties)
```
cheerfulstoic: Seems odd that this doesn't do anything...

PR #903, opened 2015-08-07
reuse existing node identity if set when calling :all

subvertallchris:


cheerfulstoic:
Do you know if this is a problem with v5.0.12?  If so, we should cherry pick these changes into a new patch release

subvertallchris:
I think it's been a problem since November, so yes, definitely.

cheerfulstoic:
Awesome, thanks ;)  Could be worse, I suppose: http://hothardware.com/news/researchers-discover-rootkit-exploit-in-intel-processors-that-dates-back-to-1997?utm_content=bufferd726d&utm_medium=social&utm_source=twitter.com&utm_campaign=buffer ;)

PR #916, opened 2015-08-15
Rel uniqueness sync

subvertallchris:
A few small changes. It...

* makes associations `:unique?` defer to rel_class if set
* adds `rel_class?` and `relationship_class?` methods to associations
* adds `create_method` class method to ActiveRel
* adds `#{related_node}_neo_id` method to ActiveRel instances to match CypherRelationship instances
* adds `unique?` to ActiveRel to match association method

I also removed some instance variables in ActiveRel tests.


cheerfulstoic:
:metal:

PR #915, opened 2015-08-14
Add Timestamps mixin

jacobwgillespie:
This adds the following mixins (a la mongoid) for including timestamp properties in an easy way:

```ruby
class Student
  include Neo4j::ActiveNode
  include Neo4j::Timestamps
end

class Student
  include Neo4j::ActiveNode
  include Neo4j::Timestamps::Created
end

class Student
  include Neo4j::ActiveNode
  include Neo4j::Timestamps::Updated
end

class EnrolledIn
  include Neo4j::ActiveRel
  include Neo4j::Timestamps
end

class EnrolledIn
  include Neo4j::ActiveRel
  include Neo4j::Timestamps::Created
end

class EnrolledIn
  include Neo4j::ActiveRel
  include Neo4j::Timestamps::Updated
end
```

`Neo4j::Timestamps` includes both `Created` and `Updated`.

Closes #861.

PR #925, opened 2015-08-18
Actually require the Neo4j::Timestamps mixin

jacobwgillespie:
So, I was stupid and forgot to require the timestamps mixin classes in my PR #915 - this fixes that.


cheerfulstoic:
Thanks!  I changed the `require`s a bit so that `neo4j.rb` requires `timestamps.rb` which in turn requires the other two.  I don't know that anybody would require `timestamps.rb` directly, but it seemed like good practice.

Released as 5.1.1

jacobwgillespie:
Thanks!

PR #931, opened 2015-08-20
simplifies node and rel persistence methods

subvertallchris:
I think this is kind of nice. It breaks up the create and update processes in such a way that it's easy to get your hands on the node or rel's properties as they would be if sent to the database for each persistence event. It's useful if you want to instantiate an instance of a class, manipulate it within the app, but then handle the DB interaction yourself.

```ruby
props = User.new(name: 'Chris', age: 31, occupation: 'Developer', location: 'Brooklyn').props_for_create

Neo4j::Session.current.query
  .match(i: 'Instrument', l: 'ProgrammingLanguage')
  .where(i: { name: 'Guitar'}, l: { name: 'Ruby' })
  .break
  .create('(i)<-[:PLAYS]-(u:User {user_props})-[:WORKS_WITH]->(l)')
  .params(user_props: props)
  .pluck(:u)
```

Out of curiosity, I compared the speed of this VS multiple queries.

```
Comparison:
                 one:      248.4 i/s
                many:       64.8 i/s - 3.84x slower
```

If you find some of the method names a little weird, I'm with you. `props_for_create` started as `creation_props`, but the parallel for update would just be `update_props`, and `update_props` in Neo4j-core is meant like a verb, "update these props." Calling `_persisted_obj.update_props(update_props)` just felt wrong. This was the best I could think to do in a way that keeps things consistent for both operations. Another method, `props_for_persistence`, will call the method that's appropriate for the given state of the node or rel.

Gonna write a few tests before merging but wanted to get it up here so I'd stop forgetting to finish it.

### DIFF
--- a/lib/neo4j/active_node/persistence.rb
+++ b/lib/neo4j/active_node/persistence.rb
@@ -47,25 +47,29 @@ module Neo4j::ActiveNode
     # Creates a model with values matching those of the instance attributes and returns its id.
     # @private
     # @return true
-    def create_model(*)
-      create_magic_properties
-      set_timestamps
-      create_magic_properties
-      properties = self.class.declared_property_manager.convert_properties_to(self, :db, props)
-      node = _create_node(properties)
+    def create_model
+      node = _create_node(props_for_create)
       init_on_load(node, node.props)
       send_props(@relationship_props) if @relationship_props
       @relationship_props = @deferred_nodes = nil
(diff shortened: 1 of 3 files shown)


cheerfulstoic:
:metal:

subvertallchris:
Damn it. Closing this to PR against 5.1.x.

PR #932, opened 2015-08-20
Config methods for timestamps

rebecca-eakins:
Hey there! In working with @subvertallchris at our day job, we realized one should be able to configure default timestamps globally from the config file. This PR makes it happen.

There are actually very few changes, since `DeclaredProperty#register_magic_property` already has a bit of timestamp-y logic. The default action is as it was before; timestamps will be `DateTime` unless told otherwise.

### DIFF
--- a/lib/neo4j/active_node.rb
+++ b/lib/neo4j/active_node.rb
@@ -41,6 +41,8 @@ module Neo4j
     include Neo4j::ActiveNode::Scope
     include Neo4j::ActiveNode::Dependent
 
+    include Neo4j::Timestamps if Neo4j::Config.record_timestamps
+
     def neo4j_obj
       _persisted_obj || fail('Tried to access native neo4j object on a non persisted object')
     end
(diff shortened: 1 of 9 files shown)


cheerfulstoic:
This is great, thanks!

What do you think about the idea of having a configuration variable to enable `created_at` / `updated_at` properties on all models?  I honestly feel like this should be the default (though it would be a 6.0.0 change, if so).  It might be good to have a method which disables timestamps as well, if this was implemented.

rebecca-eakins:
That's such great feedback! I'll get cracking on these points this weekend and add it to the PR.

Just note, I'm traveling for family reasons so I might not be able to submit until Sunday evening. 

Thank you for your thoughts!

cheerfulstoic:
Awesome, thanks!  And no rush, certainly!

rebecca-eakins:
Hi @cheerfulstoic! Since I ended up adding new functionality, I'm going to close this PR and open a new one comparing against master.

PR #933, opened 2015-08-21
Return cached result for has_one relationships

jacobwgillespie:
We'll see if Travis tests pass - ideally I'd like to add a test to check if cached results are used, but I haven't figured out the test setup yet.

Before this PR:

```ruby
b = Book.find(book_id)
b.author # queries the database
b.author # queries the database again
```

After this PR:

```ruby
b = Book.find(book_id)
b.author # queries the database
b.author # returns the cached copy of the author

b.reload # reloads the book from the database
b.author # queries the database
b.author # returns the cached copy of the author
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased][unreleased]
 
+### Fixed
+- `has_one` associations are now properly cached (like `has_many` associations)
+
 ## [5.1.2] - 08-20-2015
 
 ### Fixed
--- a/lib/neo4j/active_node/has_n.rb
+++ b/lib/neo4j/active_node/has_n.rb
@@ -380,7 +380,7 @@ module Neo4j::ActiveNode
           if options[:rel_length] && !options[:rel_length].is_a?(Fixnum)
             results
           else
-            results.first
+            results.result.first
           end
         end
       end
--- a/spec/e2e/has_one_spec.rb
+++ b/spec/e2e/has_one_spec.rb
@@ -40,6 +40,39 @@ describe 'has_one' do
       a.children.to_a.should =~ [b, c]
     end
 
+    it 'caches the result of has_one accessor' do
+      a = HasOneA.create(name: 'a')
+      b = HasOneB.create(name: 'b')
+      a.children << b
+
+      b = HasOneB.find(b.id)
+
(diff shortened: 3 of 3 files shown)


cheerfulstoic:
This is great, thanks!

If you look at `asociation_proxy_spec.rb` you'll see that there is a `expect_queries` helper method.  That's defined in the `spec_helper` which uses a `$expect_queries_count` global variable.  Not great certainly, but only used in the specs (and it's used to track a global count of queries, so maybe not the worst usage).

Anyway, you could use that to make the spec(s) if you want.  If that's good them I'm happy to bring this into the `5.1.x` branch and release a patch

cheerfulstoic:
(oh, and `association_proxy_spec` isn't the best name anymore because association proxies aren't always used there.  For later refactoring ;)

jacobwgillespie:
Sweet, added!  That was fairly painless :)

jacobwgillespie:
They're failing, but it looks like it's not related to the added tests...

cheerfulstoic:
Ah, fiddly-widdly-timey-wimey issues.  That's happened before.  I'll try to fix it for good now

cheerfulstoic:
Ah, interesting.  So there was one timey-wimey issue, but it looks like the JRuby specs are failing because in `./spec/e2e/has_one_spec.rb:63` there were 3 queries expected but only 2 happened.  That seems more related to your change

cheerfulstoic:
Weird... that line has nothing to do with `expect_queries`....  I've tried running it locally with JRuby with no issues...

cheerfulstoic:
I've got to head out for the night, but I tried re-running the builds and they're still failing, so it's reproducible on Travis, at least.  I won't be around tomorrow but I'll take a look again soon.  You're welcome to figure it out as well, of course

jacobwgillespie:
It's passing!  Realized I accidentally had the `b.reload` in the `expect_queries` - was testing the wrong thing.  :)  It was failing on my local machine too under jruby, but this fixes everywhere.  Thanks for the help!

cheerfulstoic:
Awesome!  Thanks!  I'll merge and cut a patch shortly

jacobwgillespie:
Thank you!  :cake:

cheerfulstoic:
Released as 5.1.3.  Thanks again!

cheerfulstoic:
Thanks coveralls... :-/

PR #935, opened 2015-08-22
simplifies node and rel persistence methods

subvertallchris:
I think this is kind of nice. It breaks up the create and update processes in such a way that it's easy to get your hands on the node or rel's properties as they would be if sent to the database for each persistence event. It's useful if you want to instantiate an instance of a class, manipulate it within the app, but then handle the DB interaction yourself.

```ruby
props = User.new(name: 'Chris', age: 31, occupation: 'Developer', location: 'Brooklyn').props_for_create

Neo4j::Session.current.query
  .match(i: 'Instrument', l: 'ProgrammingLanguage')
  .where(i: { name: 'Guitar'}, l: { name: 'Ruby' })
  .break
  .create('(i)<-[:PLAYS]-(u:User {user_props})-[:WORKS_WITH]->(l)')
  .params(user_props: props)
  .pluck(:u)
```

Out of curiosity, I compared the speed of this VS multiple queries.

```
Comparison:
                 one:      248.4 i/s
                many:       64.8 i/s - 3.84x slower
```

If you find some of the method names a little weird, I'm with you. `props_for_create` started as `creation_props`, but the parallel for update would just be `update_props`, and `update_props` in Neo4j-core is meant like a verb, "update these props." Calling `_persisted_obj.update_props(update_props)` just felt wrong. This was the best I could think to do in a way that keeps things consistent for both operations. Another method, `props_for_persistence`, will call the method that's appropriate for the given state of the node or rel.

Gonna write a few tests before merging but wanted to get it up here so I'd stop forgetting to finish it.

Itemized changes:

* New `props_for_create`, `labels_for_create`, `props_for_update`, and `props_for_persistence` methods
* Refactored `create_model` and `update_model` methods to be simpler, more focused. Some code was moved into the `Shared::Persistence` module to keep things DRY.
* Reorganized the unit tests so the paths match `lib`. It's easier to find what you're looking for.
* Removed a few splat arguments to cut allocation during persistence. Negligible performance improvement but it has to start somewhere.


subvertallchris:
I reorganized the unit tests so directory structure mirrors `lib`. It makes it easier to find the tests responsible for the file you're working with.

cheerfulstoic:
:metal: re: the specs ;)  The other stuff seems great too!

PR #939, opened 2015-08-25
Timestamps Type and Global Inclusion

rebecca-eakins:
This PR does a few things with `Neo4j::Config` and timestamps.

First, I added a method `timestamp_type` to allow users to configure their default timestamps globally. Currently, it defaults to `DateTime`, but users can easily specify `Integer` or `Date` as they please. This involved a small refactor in `DeclaredProperty#register_magic_property`, but all the logic is pretty straightforward.

Second, I added `Neo4j::Config.record_timestamps` for Rails-inspired ease. It defaults to false, but if set to true, all `ActiveNode` and `ActiveRel` models will include the timestamps module.

Thanks for all the help! Please let me know if you have any questions/thoughts.


rebecca-eakins:
@subvertallchris suggested I include documentation for this PR. To resolve merge conflicts, I added `docs/_exts/hidden_code_block.pyc` to the gitignore file.

docs/Configuration.rst
```diff
@@ -59,3 +59,13 @@ Configuration
     **Default:** ``nil``
 
     If true, format outputted queries with newlines and colors to be more easily readable by humans
+
+  **record_timestamps**
+    **Default:** ``false``
+    
+    A Rails-inspired configuration to manage inclusion of the Timestamps module. If set to true, all ActiveNode and ActiveRel models will include the Timestamps module and have ``:created_at`` and ``:updated_at`` properties. For example, include the following in your config: ``Neo4j::Config.record_timestamps = true``.
```
cheerfulstoic: The config (in `application.rb` or `environments/(development|test|production).rb` would look like `neo4j.config.record_timestamps = true`.

Probably this should be covered at the top of this page

docs/Configuration.rst
```diff
@@ -59,3 +59,13 @@ Configuration
...
+
+  **timestamp_type**
+    **Default:** ``DateTime``
+
+    This method returns the specified default type for the ``:created_at`` and ``:updated_at`` timestamps. Specify another type (e.g. ``Integer``) by declaring it in your config (``Neo4j::Config[:timestamp_type] = Date``).
```
cheerfulstoic: What would happen if you set this to `Boolean` or some other type which couldn't represent dates?  I imagine that valid value for this could be `Date`/`DateTime`/`Time`/`Integer`/`String`.

Also, the same comment as above.  We should specify at the top how to set these configuration variables

cheerfulstoic:
I made a couple of comments, but this all seems straightforward!  Thanks for the docs!  I won't be around tonight but I'd be happy to merge this tomorrow.

The only other thing that I can think of is to have a `CHANGELOG.md` entry (according to the http://keepachangelog.com/ standard).  If you don't have time I'll throw it in after merging.

Thanks!

cheerfulstoic:
Actually, since I already looked it over I'm not sure why I didn't just merge this... ;)  It's in master and should come out in 5.2.0 on Friday.  Thanks again!

docs/Configuration.rst
cheerfulstoic: Actually, I just realized that my suggestion is Rails-specific.  I'll put in some docs at the top which make it clear.  Thanks!

PR #941, opened 2015-08-25
fix association setting with node ids

subvertallchris:
@jacobwgillespie pointed out that unpersisted relationship setting wasn't working with ids. This should fix that.

### DIFF
--- a/lib/neo4j/active_node/unpersisted.rb
+++ b/lib/neo4j/active_node/unpersisted.rb
@@ -40,8 +40,10 @@ module Neo4j
       end
 
       def save_and_associate_node(association_name, node, operator)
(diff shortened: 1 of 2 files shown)


jacobwgillespie:
Works for me - thanks!  :+1:

subvertallchris:
Great! Can you wait for a general release of this or do you need a patch ASAP? I'm going out to lunch for real now but we can do a patch when I get back if you need it.

jacobwgillespie:
Patch would be awesome if it's not too much trouble - thanks!

subvertallchris:
Let me see if I can modify that query that clears out the old rel ids first...

Issue #948, opened 2015-08-31
Boolean Defaults Not Saved


jacobwgillespie:
Boolean default values do not seem to be saved into the db when calling `Model.new(params).save` on 5.1.5.  Here's an example model:

```ruby
class User
  include Neo4j::ActiveNode
  include GlobalID::Identification

  # Properties

  property :username, type: String
  property :login, type: String
  property :email, type: String
  property :hashed_password, type: String
  property :name, type: String

  property :created_at, type: DateTime
  property :updated_at, type: DateTime

  property :is_verified, type: Boolean, default: false
  property :is_site_admin, type: Boolean, default: false
  property :is_private, type: Boolean, default: false

  property :test_default, type: String, default: 'works'
  ...
end
```

This saves the values of the boolean `is_verified`, etc., properties:

```ruby
u = User.new
u.assign_attributes(params)
u.save
```

```ruby
CYPHER 52ms CREATE (n:`User` {props}) RETURN ID(n) | {:props=>{:uuid=>"2841bba0-c4d2-4a65-864a-71e7606a27d4", :username=>"test", :login=>"test", :email=>"test@example.com", :hashed_password=>"$2a$10$xpiWg6pZabwnSAL66oxGT..zUthcRTIX8sBlQev4cphJd09euetZy", :name=>"test", :created_at=>1441043306, :updated_at=>1441043306, :is_verified=>false, :is_site_admin=>false, :is_private=>false, :test_default=>"works"}}
```

And this does not:

```ruby
u = User.new(params)
u.save
```

```ruby
CYPHER 53ms CREATE (n:`User` {props}) RETURN ID(n) | {:props=>{:uuid=>"d7fcd455-a01e-4475-84e5-a7ce0bfae36e", :username=>"test", :login=>"test", :email=>"test@example.com", :hashed_password=>"$2a$10$r6mOaJNlqez25.MG01fuuuIqR07sQXiTHY9UWEjmRQ46YaN/u8iGW", :name=>"test", :created_at=>1441043368, :updated_at=>1441043368, :test_default=>"works"}}
```

The default value of `test_default` will be saved in the DB with either.

Not sure if it matters, but `params` is the return value from `params.permit(:name, :username, ...)`.

subvertallchris:
Good catch. Are you sure it's just Boolean defaults?

The master branch, targeting 5.2.x, includes a heavy refactor of the
persistence process. Maybe try with that to see if it got corrected along
the way.

[...]

PR #957, opened 2015-09-08
Boolean fix

subvertallchris:
This fixes problems related to default values of `ActiveAttr::Typecasting::Boolean` properties. Along the way, I found that `updated_at` was not being added to nodes outside of `create` actions and `props_for_update` was using String keys while `props_for_create` had symbol keys.


subvertallchris:
Fixes #948.

PR #968, opened 2015-09-18
Activerel class array

subvertallchris:
* Brings array options to `ActiveRel`'s `(from|to)_class`
* Deprecates all `ActiveRel` query methods.

Because query methods are deprecated, I did not modify them to support Array from/to. 

If you are using these methods and want to disable the deprecation warning, set `ActiveSupport::Deprecation.silenced = true`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased][unreleased]
 
+### Added
+
+- Support for Array arguments to ActiveRel's `from_class` and `to_class`.
+
+### Changed
+
+- Deprecated all methods in ActiveRel's Query module except for those that allow finding by id.
+
 ## [5.2.6] - 09-16-2015
 
 ### Fixed
--- a/lib/neo4j/active_rel/persistence.rb
+++ b/lib/neo4j/active_rel/persistence.rb
@@ -60,11 +60,21 @@ module Neo4j::ActiveRel
         type = from_node == node ? :_from_class : :_to_class
         type_class = self.class.send(type)
 
-        next if [:any, false].include?(type_class)
-
-        unless node.class.mapped_label_names.include?(type_class.to_s.constantize.mapped_label_name)
+        unless valid_type?(type_class, node)
           fail ModelClassInvalidError, type_validation_error_message(node, type_class)
         end
+      end
+    end
+
+    def valid_type?(type_object, node)
+      case type_object
+      when false, :any
+        true
+      when Array
+        type_object.each { |c| return false unless valid_type?(c, node) }
+        true
+      else
+        node.class.mapped_label_names.include?(type_object.to_s.constantize.mapped_label_name)
       end
     end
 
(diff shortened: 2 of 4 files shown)


janpieper:
Hi @subvertallchris! Maybe I am wrong, but this implementation feels broken...

As I understand, the Array will allow something like this:

```
                                -----> Apple
                               /
                              /
Basket ----- [:CONTAINS]-----
                              \
                               \
                                -----> Banana
```

I think, the matching ruby code would look like this:

```ruby
class Basket
  include Neo4j::ActiveNode
  has_many :out, :apples, type: CONTAINS, model_class: Apple
  has_many :out, :bananas, type: CONTAINS, model_class: Banana
end

class Apple
  include Neo4j::ActiveNode
  property :weight, type: Integer
  has_one :in, :basekt, origin: :apples, model_class: Basket
end

class Banana
  include Neo4j::ActiveNode
  property :weight, type: Integer
  has_one :in, :basekt, origin: :apples, model_class: Basket
end

class ContainsRelation
  include Neo4j::ActiveRel
  from_class Basket
  to_class [ Apple, Banana ]
  type :CONTAINS
end
```

Now, i want to use the relation to put one apple and one banana into the basket.

```ruby
basekt = Basket.create

apple = Apple.create(weight: 10)
banana = Banana.create(weight: 20)

ContainsRelation.create(from_node: basket, to_node: apple)
ContainsRelation.create(from_node: basket, to_node: banana)
```

But I get an error :open_mouth:

```
Neo4j::ActiveRel::Persistence::ModelClassInvalidError:
  Node class was Apple (54190140), expected [
    Apple(weight: Integer),
    Banana(weight: Integer)
  ] (55617160)
```

Tested with 67e11bd2.

If I take a look into `#valid_type?` in `lib/neo4j/active_rel/persistence.rb` i see this:

```ruby
def valid_type?(type_object, node)
  case type_object
  when false, :any
    true
  when Array
    type_object.each { |c| return false unless valid_type?(c, node) }
    true
  else
    node.class.mapped_label_names.include?(type_object.to_s.constantize.mapped_label_name)
  end
end
```

For me, there are at least two bugs in this method...

1. I don't get how the else-part should work... It checks if an array contains another array. This won't work in this case: `[:Apple].include?([:Apple]) # false`
2. Shouldn't the Array-part use `#any?` to check if any specified class matches the given node!?

I think, the Array-part should look like this:

```ruby
def valid_type?(type_object, node)
  # ...
  when Array
    type_object.any? { |c| valid_type?(c, node) }
  else
  # ...
end
```

subvertallchris:
Yup, looks like my tests didn't go far enough. I get _really_ distracted by trying to write tests that didn't duplicate and clearly lost focus a little. I'll be home in three hours and will fix it first thing. Thanks for tracking that down.

cheerfulstoic:
Also not a huge deal but we normally recommend using symbols rather than the class constants just in case there's issues with the load order of your classes.

subvertallchris:
Huge +1 to that. I'd love to find a better way of handling that.

janpieper:
@cheerfulstoic: Yeah, i know! With rails, it works fine, without --> chicken-and-egg problem :smile:

subvertallchris:
@janpieper That fix will be merged in a sec, I had a lull at work so just popped it in and fixed the test.

PR #974, opened 2015-09-24
Return true on successful save! calls

jmdeldin:
For [consistency with ActiveRecord](http://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-save-21), I've made it so ActiveNode and ActiveRel return `true` when `save!` does not error out.

## TODO

- [x] Not sure if it's worth fighting Rubocop for `or` instead of `||` -- is there a magic cookie?

### DIFF
--- a/lib/neo4j/active_node/persistence.rb
+++ b/lib/neo4j/active_node/persistence.rb
@@ -40,7 +40,7 @@ module Neo4j::ActiveNode
     # @see Neo4j::Rails::Validations Neo4j::Rails::Validations - for the :validate parameter
     # @see Neo4j::Rails::Callbacks Neo4j::Rails::Callbacks - for callbacks
     def save!(*args)
-      fail RecordInvalidError, self unless save(*args)
+      save(*args) || fail(RecordInvalidError, self)
     end
 
     # Creates a model with values matching those of the instance attributes and returns its id.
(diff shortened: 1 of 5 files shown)


cheerfulstoic:
This is great, thanks so much!

A bit more subtlety would be nice from rubocop regarding "or".  Rubocop goes mainly off of the ruby-style-guide, though which says "The and and or keywords are banned. It's just not worth it. Always use && and || instead."

Avdi Grimm has an excellent free RubyTapas on the topic, though:

http://devblog.avdi.org/2014/08/26/how-to-use-rubys-english-andor-operators-without-going-nuts/

I'm going to merge in your changes and test something in Rubocop...

cheerfulstoic:
Ok, it looks like Rubocop doesn't allow for any configuration of this cop.  You can disable the cop for specific files, but I definitely like making sure that we use the symbolic versions for boolean expressions.  So for now let's keep the line comments, but I've made a comment over here:

https://github.com/bbatsov/ruby-style-guide/issues/415

Thanks again!

jmdeldin:
Thanks, @cheerfulstoic! Agreed on the rubocop/style guide as well -- it'd be great if `or` and `and` were allowed for control flow. Maybe rubocop could look for lines containing `or`/`and` and `if`/`unless`, and flag those as no good, terrible Rubies. Otherwise, we would get a free pass for control flow.

cheerfulstoic:
One other issue is the lower precedence of `=` against the english operators.  There are cases where you might want it to go both ways, though (I think both are demoed in the screencast).

We'll see how it goes down with the style guide repo.  Though yeah, even if that gets changed it might take some clever hacking to find a way to have rules like you'd want.  I know Rubocop will actually tell you when your code is ambiguous, though, so maybe there's a shot...

PR #977, opened 2015-09-28
Index property management

subvertallchris:
A lot of breaking changes here:

* Changes `DeclaredPropertyManager` to `DeclaredProperties`
* Prevents users from setting both an exact index and unique constraint. It covers options given to `properties` as well as the `index` and `constraint` methods and combinations thereof.
* Removes the optional arguments to `index` that would allow you to use that method to set constraints.
* Replaces all calls to `on_session_available` with `on_next_session_available`, per change in Neo4j-core.

Non-breaking changes:

* Some convenience methods to get your hands on declared property information. `MyModel.declared_properties[key]`, `.declared_properties.each`, and a few others.
* `DeclaredProperty` instances have a few methods to get and set indexing/constraint info. Note that these methods do not actually set properties in the database, you should use the class methods on the ActiveNode models for that.
* I moved all of ActiveNode's index-related code into an `Indexing` module.

Fixes:

* Properties are now properly declared when inheriting. `MySubclass.declared_properties` will include properties from the parent. 

To-do:

* This could use some DRY help. I'm still in my "just-make-it-work" first round.
* I might rework Neo4j-core's `Label` class. The API for index and constraints is more complicated than it needs to be, which leads to a lot of the complication in our `Indexing` module. The methods to add or remove indexes and constraints should be able to share code but the Neo4j-core APIs are juuuust different enough that they can't.

**Travis will fail until https://github.com/neo4jrb/neo4j-core/pull/226 is merged in. This branch references my local paths at the moment and is not ready for merge.**


subvertallchris:
This is mostly good but I want to write a few more tests to cover the new APIs directly. I also want to run it through its paces with some of my local apps. There might be a bug, too... It's still possible to set an index through an option to `property` and then call the `constraint` class method. Just need to add a check to prevent that.

I'm happy with this overall, though. It removes a lot of responsibility from ActiveNode.

Issue #951, opened 2015-09-03
Stringified booleans not being cast as booleans in rel_where


jmdeldin:
Following up from Gitter, I'm finding boolean values (when passed in as strings, e.g., `thing.rel_where(other_thing: 'false')`) are not cast to bools in `rel_where`. From [this example](https://gist.github.com/jmdeldin/2d67e63e84021426d669): 

```ruby
# 1. this works
rs = a1.books.rel_where(enjoyed_writing: false).to_a
puts "books: #{rs.length} (should be 1)"

# 2. this does not cast 'false' into false
rs = a1.books.rel_where(enjoyed_writing: 'false').to_a
puts "books: #{rs.length} (should be 1)"
```
I worked around this in my own code, but it _seems_ like this should work.

subvertallchris:
Good catch, I can find and fix this when I return in a few days.

subvertallchris:
`rel_where` doesn't do any typecasting, actually. It's going to take a little work to fix it, I'm not sure it's going to happen very quickly. I also discovered that QueryProxy's typecasting only works with those converters defined within this gem, not those provided by ActiveAttr. I have some ideas for fixing it but I'm not sure I like any of them. Will post some updates when I have them but in the meantime, if you or anyone wants to take a stab at correcting this, PRs are welcome.

jmdeldin:
Ah, thanks for the update! I'm not sure if I'll have anytime soon, but if I were to, would I start somewhere in the [Query::QueryProxy::Link](https://github.com/neo4jrb/neo4j/blob/b77ae9d24853705eaa437302bbdf8128a0dfd784/lib/neo4j/active_node/query/query_proxy_link.rb) class?

subvertallchris:
Sort of. I think I actually managed to get it working, kind of, but here's some background:

Working backwards, the ultimate goal is `for_rel_where_clause` needs to get the ActiveRel model as an argument: https://github.com/neo4jrb/neo4j/blob/b77ae9d24853705eaa437302bbdf8128a0dfd784/lib/neo4j/active_node/query/query_proxy_link.rb#L75. If that happens, you can use https://github.com/neo4jrb/neo4j/blob/b77ae9d24853705eaa437302bbdf8128a0dfd784/lib/neo4j/active_node/query/query_proxy_link.rb#L58 or a variation thereof to perform the conversion.

[...]

PR #986, opened 2015-10-14
Fix String->Boolean type conversion

jmdeldin:
Following from #951, String to Boolean conversion is not working. After prodding around, I think the issue is that `Neo4j::Shared::TypeConverters#converters` has all class keys except `[true, false]` for bools. When you declare a property's `type` as `Boolean`, the `converted_property` method cannot locate the `BooleanConverter`.

This PR makes it so the `ActiveAttr::TypeCasting::Boolean` is registered as a converter by making `BooleanConverter#convert_type` return that class.

- [x] add failing spec
- [x] fix converter

### DIFF
--- a/lib/neo4j/shared/type_converters.rb
+++ b/lib/neo4j/shared/type_converters.rb
@@ -97,13 +97,11 @@ module Neo4j::Shared
           convert_type.include?(value)
         end
 
-        def convert_type
-          [true, false]
-        end
-
         def db_type
           ActiveAttr::Typecasting::Boolean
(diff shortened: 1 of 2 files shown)


subvertallchris:
This is good! I branched your branch to make some slight changes -- we need `converted?` to inspect an array, so I fixed that and added a test. It'll be in another PR in a moment. Great catch, thanks for taking care of this.

jmdeldin:
Awesome! Thanks for the final work and merging!

PR #987, opened 2015-10-15
Where conversion

subvertallchris:
See #986. Adds one extra method and test. cc @jmdeldin


subvertallchris:
That one failure is due to memory issues in JRuby, not a reflection of any changes in the code, so I'm going to merge this. @jmdeldin, thanks for the help with this!

PR #1012, opened 2015-10-26
Unpersisted node handling for ActiveRel

subvertallchris:
This gives ActiveRel models the ability to create a relationship and one or both nodes in a single Cypher query, _with_ all validations and callbacks on all objects.

```ruby
student = Student.new
lesson = Lesson.new
rel = EnrolledIn.create(from_node: student, to_node: lesson)
 # CYPHER 6ms CREATE (from_node:`Student` {uuid: {from_node_Student_uuid}, l: {from_node_Student_l}}), (to_node:`Lesson` {uuid: {to_node_Lesson_uuid}}), from_node-[rel:ENROLLED_IN {rel_params}]->to_node RETURN rel, from_node, to_node | {:from_node_Student_uuid=>"9a088b89-5906-4db0-8d8c-04fdcf7aaa43", :from_node_Student_l=>false, :to_node_Lesson_uuid=>"2e1338a3-0b91-4675-8e0b-ec66a7078d0f", :rel_params=>{:created_at=>1445878151, :updated_at=>1445878151, :active=>false}}
# => #<EnrolledIn (:Student {uuid: "9a088b89-5906-4db0-8d8c-04fdcf7aaa43"})-[:ENROLLED_IN]->(:Lesson {uuid: "2e1338a3-0b91-4675-8e0b-ec66a7078d0f"}) active: false, created_at: Mon, 26 Oct 2015 16:49:11 +0000, grade: nil, student_notes: nil, teacher_notes: nil, thing: nil, updated_at: Mon, 26 Oct 2015 16:49:11 +0000> 
student.persisted?
#  => true 
lesson.persisted?
# => true 
student.uuid
# => "9a088b89-5906-4db0-8d8c-04fdcf7aaa43" 
lesson.uuid
 # => "2e1338a3-0b91-4675-8e0b-ec66a7078d0f" 
```

It takes advantage of some new Factory classes that, given a node or rel, will build a Cypher query to match or create based on the state of the object. Factory instances can be connected to others, allowing you to progressively build a query.

When creating two nodes and a rel at the same time, my benchmarks indicate a 15% performance improvement over three separate queries. I _think_ the factory code could be reused in some other classes to separate concerns a bit and take advantage of the one-query-to-rule-them-all feature, but that's TBD.

There are still a few tests I want to add to for the factories but this is stable as far as I've seen. I'm opening the PR now in case anyone wants to give it a shot or comment and so I don't forget to finish it up in the next few days.


review-ninja:
<a href="https://app.review.ninja/neo4jrb/neo4j/pull/1012" target="_blank">ReviewNinja

cheerfulstoic:
Awesome!  I imagine this could perhaps be done with the batch query functionality in the new CypherSession API when that comes along, yes?

One thing that I've been meaning to bring up, what do you think about supporting the syntax `ActiveRelModel.create(from_node_object, to_node_object, property: value)`?

subvertallchris:
I don't think so, since you'd still need to get the results from the node creations before you can create the rel, so they either need to be completely separate or all at the same time.

I like the flexibility of the current method signature. The `from_node=`/`to_node=` methods give you the option of setting properties and dealing with the nodes separately. Keeping the signature between the modules lets us reuse a lot of code and keeps things simpler, too.

subvertallchris:
The basic association creation code might be able to take advantage of this, though, and then they'd see the performance benefit. We'd also be able to simplify https://github.com/neo4jrb/neo4j/blob/master/lib/neo4j/active_node/has_n/association.rb, since it wouldn't need to be responsible for rel creation.

If we do that, it might also be a good time to add the "associations-with-rel-class-specified-use-that-rel-class-when-using-the-<<-operator" feature.

PR #1034, opened 2015-11-03
_classname removal

subvertallchris:
Woooooooooo...

For those just joining us, `_classname` was a legacy feature that we used to improve performance prior to DB v2.1.5. Nodes and rels came back from the DB without labels and types, so a second query was necessary to find them. Removing this code helps keep things cleaner and simpler.

This means that < 2.1.4 is completely unsupported. I think it will still probably work -- code exists for it to find type/label when missing -- but it will be extremely slow and we cannot make any promises.


cheerfulstoic:
:dance:

FYI to all: We generally like to keep backwards compatibility as much as possible, but this is definitely worth it. The upgrade from 2.0.0 - 2.1.5 shouldn't be too bad and we haven't supported 1.x for a long time)

cheerfulstoic:
Also, yay for:

11 additions and 536 deletions. Coverage 97.19% (Diff 96.91%)

PR #1036, opened 2015-11-04
ActiveRel#create and #create! three-arg syntax

subvertallchris:
Just like `initialize`, only a little less efficient (one allocation) because it's not often that someone needs to _create_ enough rels at once that it becomes an issue.


subvertallchris:
Failing from a single timestamp mismatch, so merging.

Issue #1035, opened 2015-11-04
Arguments for `creates_unique`?


cheerfulstoic:
I just started using `ActiveRel.creates_unique` and I thought it would always only result in one relationship, but it only does that if all of the properties that I'm giving are equal to another relationship's properties.  It would be nice to be able to say which properties it should match on.  Then the `ActiveRel` code can pull all of the rest out into a `ON CREATE SET`

subvertallchris:
As a wise man once said... https://www.youtube.com/watch?v=UaU6NQzEGcg

That's really unfortunate.

subvertallchris:
Is this worth backporting into 5.2.x? It reminds of me that `find_or_create` issue, where we're changing something that's not technically a bug but it defies expectations.

cheerfulstoic:
It's a funny one because I don't think it would defy expectations if you know what `CREATES UNIQUE` is doing.  Though if we don't, what are the chances that somebody else will?

Regarding backporting, it's tempting, but I'd also like to have reasons for people to want to update to 6.0 ;)

subvertallchris:
Yeah, for sure. It makes sense once you know about it...

Since there's so much room for confusion, I think should make the syntax more explicit so you can't accidentally end up with unwanted behavior. In 6.0, maybe we **require** an argument of `nil` or `:none`, `:all`, `on: [keys]`, or `except: [keys]`. `nil` would give the behavior we expected now, the others would do as expected?

I'm going to do a quick spike of this -- already have `nil` behavior working -- and we can tighten up the options tonight and tomorrow.

subvertallchris:
That PR is a proposed implementation of the comment above.

subvertallchris:
Ooh, we can close this.

PR #1038, opened 2015-11-06
Arguments for ActiveRel.creates_unique, new Association `unique` options

subvertallchris:
This is RE: https://github.com/neo4jrb/neo4j/issues/1035. It forces ActiveRel's `creates_unique` and `Association`'s `unique` option to take an argument.

* `:none` means that all properties are ignored where `CREATE UNIQUE` is concerned, properties will be set after the rel is created, like our current expectations
* `:all` means that any property's change will result in a new rel
* `{on: [keys]}` defines which properties it uses for the match

Called with no arguments or `true`, it will default to `:none`.

(This has been edited to reflect the end results. Comments below reflect some earlier content here.)


subvertallchris:
Just pushed an update to this that makes plain ole' associations' `unique` option require one of these same arguments. I'm trying to massage things into place so we can reuse more code, so there will be a few more commits.

Rubocop is barking and it needs a little refactoring. I'm going to extract the specifics of how a relationship is built from `QueryProxy` into a new class tomorrow.

cheerfulstoic:
Awesome, thanks ;)

I've been thinking about the API and I'm honestly not sure how it will be used, but my thought was that by default you might call `creates_unique` and it should work like your `:none` case and then if you specify some arguments those would be equivalent to the `{on: [keys]}` case.  I wonder if the `except` or `:all` options would be used much, if at all.

subvertallchris:
I'm ok with it defaulting to `:none` but I worry that someone who's used to
Cypher's behavior would expect it to behave like `:all`.

 `none` and `on` with keys are ones I expect to be the most useful, but I
added the others since it just felt like they should be there and the code
made it easy enough. ;-)

cheerfulstoic:
Regarding people who are used to Cypher, I suspect they're generally going to `CREATE UNIQUE` with either no or only a small set of properties.  The problem is that our API automatically puts in all of the properties/values for you.  So thinking about it, I think that even people experienced with Cypher would be surprised at our implementation of `creates_unique`.

I'm trying to think of what bothers me about the `except` cases, and I think the problem is that you can have surprising behavior on a team if somebody adds a property.  For something that seems like it should be an innocent action, it could have larger consequences.  Allowing people to just specify the fields that they want to be used seems safer.  Plus it's a cleaner interface because `creates_unique` without arguments becomes what people would expect it to be (only ever create one relationship between two given nodes) and then you can list out the properties that are important otherwise.

subvertallchris:
That makes sense. I can remove the Hash aspect of the parameter entirely
and just have it be the Symbol or list of keys. If no argument is given, it
can default to `:none`, but I'm still not sure it's safe to assume that
everyone will interpret it the way we did. I think we should keep `:all` as
an option.

cheerfulstoic:
Sounds good!  The only other thing is that I have the slightest of worries that somebody might have a property called `all`.  So maybe `all_properties`?  I dunno, up to you ;)

subvertallchris:
Let's just stick with a hash, `{on: [keys]}`, to keep things simple and prevent conflicts. It also gives a path for future filter types if something becomes necessary.

subvertallchris:
Snuck that in during lunch, will set the default `:none` tonight.

cheerfulstoic:
Cool ;p

subvertallchris:
Closed by mistake?

cheerfulstoic:
Yup, sorry about that!

subvertallchris:
Ugh, failing in JRuby due to dependency issue. https://github.com/neo4jrb/neo4j/issues/1040

PR #1047, opened 2015-11-15
QueryProxy#<< w/ActiveRel set will use create!

### LINKED ISSUE #1045: `QueryProxy#<<` should use `create!` when rel_class is present
This should occur at https://github.com/neo4jrb/neo4j/blob/master/lib/neo4j/active_node/has_n/association/rel_factory.rb#L31

As an alternative, https://github.com/neo4jrb/neo4j/blob/master/lib/neo4j/active_node/query/query_proxy.rb#L189 could look at the response and, if false, it should raise an error there.

subvertallchris:
Closes #1045 but I'm not sure that this is the way to go. Users need to know if their relationships are failing, but it should probably return a Boolean so you can look for it instead of having to rescue from an error. Maybe we should have a config setting that will use `create!` every time, if that's what the user wants. @cheerfulstoic, thoughts?

### DIFF
--- a/lib/neo4j/active_node/has_n/association/rel_factory.rb
+++ b/lib/neo4j/active_node/has_n/association/rel_factory.rb
@@ -28,7 +28,7 @@ module Neo4j::ActiveNode::HasN
       def _create_relationship_with_rel_class
         Array(other_node_or_nodes).each do |other_node|
           node_props = _nodes_for_create(other_node, :from_node, :to_node)
-          association.relationship_class.create(properties.merge(node_props))
+          association.relationship_class.create!(properties.merge(node_props))
         end
       end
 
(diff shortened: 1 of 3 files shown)


cheerfulstoic:
In Ruby `(array << object)` returns the array after the object has been added.  I've sometimes seen people do `(array << object).method`.

People definitely need to know, though, if there are errors in saving due to validations or whatever, since we're wrapping the `ActiveRel.create`.  Probably we should let any exceptions raise up and create a special exception for validations...  I don't like having to create too many configuration settings about this kind of thing.  We should have a well-thought through API that gives them the ability to choose one method or another to get the behavior that they want.

subvertallchris:
I hadn't given much thought about returning a Boolean until after I threw
this up here, it was very off the cuff.

I think it's fine to raise an error on failure in this case. If they have
the expectation that a validation might fail and they want a want
a Boolean, they can use `save` on the rel class directly. `create!` raises
`ActiveRel::Persistence::InvalidRel` or something inspired by
`ActiveRecord;:RecordInvalid`, which should be fine for now as far as the
exception class goes.

PR #1059, opened 2015-11-29
Fix spec on find_first method

takiy33:


takiy33:
I close this PR because test is not wrong.
I misunderstood it.

cheerfulstoic:
I was just having a look at this and it still seems broken to me.  Does rspec support `should.nil?` to test for `nil` values?  If so, I don't really like that syntax as it's obviously caused confusion for both of us ;)

cheerfulstoic:
It seems like in the original project the line is `should == nil`:

https://github.com/ianwhite/orm_adapter/blob/f8c570a60d9b7a239eca721bb9504a9e3dec48c5/spec/orm_adapter/example_app_shared.rb#L93

Though I think `should be_nil` like your change is recommended by the RSpec team.  I still think this should be fixed, but I'm wondering if I'm missing something

takiy33:
@cheerfulstoic Thank you for reply.
I restore the branch.

cheerfulstoic:
Cool, thanks!  I merged it before Travis CI was done because it passed last time.

Let us know if you have any other questions!  We're also on Gitter: https://gitter.im/neo4jrb/neo4j

PR #1060, opened 2015-11-30
Support ERB in 'neo4j.yml' config file

mrstif:
Support ERB in the `neo4j.yml` config file:

    development:
      url: http://localhost:7474
    production:
      url: <%= ENV['NEO4J_URL'] %>

### DIFF
--- a/lib/neo4j/railtie.rb
+++ b/lib/neo4j/railtie.rb
@@ -42,7 +42,7 @@ module Neo4j
 
       def config_data
         @config_data ||= if yaml_path
-                           HashWithIndifferentAccess.new(YAML.load(yaml_path.read)[Rails.env])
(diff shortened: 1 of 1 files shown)


cheerfulstoic:
I actually considered this, but I think the `railtie.rb` automatically looks for and prioritizes the `NEO4J_URL` environment variable.  Is that what you're experiencing?

mrstif:
Sorry @cheerfulstoic ! This was a bad example... I didn't remember that that variable is actually the reserved one in the railtie... A proper example would be `<%= ENV['GRAPHSTORY_URL'] %>`

Just think it's a bit messy to scatter the configuration all around. It's just a matter of keeping it all in one place :)

cheerfulstoic:
Right, good point.  Let's do it!

mrstif:
Cheers @cheerfulstoic ! :D

PR #1062, opened 2015-12-09
Use `--require spec_helper` option

koic:
This is how I go about it. I thought [Don't Repeat Yourself](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).


cheerfulstoic:
Love it, thanks!

koic:
Merged, thanks :smile:

Issue #1073, opened 2015-12-18
Neo4j::ActiveRel #reload not working properly?


ProGM:
I tried a code like this, but I'm getting strange results. Is `reload` broken?
```bash
irb(main):003:0> v = MyRel.create(from_node: a, to_node: b)
 CYPHER 3ms MATCH (from_node), (to_node) WHERE (ID(from_node) = {from_node_id}) AND (ID(to_node) = {to_node_id}) CREATE UNIQUE from_node-[rel:has_visited {rel_create_props}]->to_node SET rel.`created_at` = {setter_rel_created_at}, rel.`updated_at` = {setter_rel_updated_at}, rel.`times` = {setter_rel_times} RETURN rel | {:from_node_id=>19, :to_node_id=>49, :setter_rel_created_at=>1450453432, :setter_rel_updated_at=>1450453432, :setter_rel_times=>1, :rel_create_props=>{}}
=> #<MyRel created_at: Fri, 18 Dec 2015 15:43:52 +0000, updated_at: Fri, 18 Dec 2015 15:43:52 +0000, times: 1>
irb(main):004:0> v2 = VisitRel.last
DEPRECATION WARNING: The Neo4j::ActiveRel::Query module has been deprecated and will be removed in a future version of the gem. (called from irb_binding at (irb):4)
 CYPHER 24ms MATCH (n1:`Person`)-[r1:`has_visited`]->(n2) RETURN r1 ORDER BY ID(r1) DESC LIMIT {limit_1} | {:limit_1=>1}
=> #<VisitRel created_at: Fri, 18 Dec 2015 15:43:52 +0000, updated_at: Fri, 18 Dec 2015 15:43:52 +0000, times: 1>
irb(main):009:0> v.update_attribute(:times, 7)
CYPHER 67ms MATCH (node)-[n]-() WHERE ID(n) = {neo_id} SET n.`updated_at`= {updated_at},n.`times`= {times} | {"updated_at"=>1450453502, "times"=>7, :neo_id=>1621}
irb(main):017:0> v2.reload.times
 CYPHER 3ms MATCH n WHERE (ID(n) = {ID_n}) RETURN n | {:ID_n=>1621}
=> 1
```

cheerfulstoic:
Looks like @subvertallchris has been working on this in #1076

subvertallchris:
[...]

Issue #1067, opened 2015-12-16
Any planned support for ActiveRecord-like #increment! method?


ProGM:
Do you plan to add support for `#increment!` to both `ActiveNode` and `ActiveRel`?

It could be useful for keeping visits count.

http://apidock.com/rails/ActiveRecord/Base/increment!

cheerfulstoic:
Interesting, I'm a long time AR user but I hadn't heard of `#increment!` before.  Definitely something we could implement.  I can't tonight but I'll give it a shot tomorrow (a PR is also welcome in the meantime.  We're happy to help on https://gitter.im/neo4jrb/neo4j if you have any issues / questions)

ProGM:
Uhm... I checked how it's implemented in ActiveRecord. It seems it simply does something like:

```
def increment(attribute, by = 1)
    self[attribute] ||= 0
    self[attribute] += by
    self
end

def increment!(attribute, by = 1)
    increment(attribute, by).update_attribute(attribute, self[attribute])
end
```

This is ok, but it's not concurrent (as stated in the docs). Do you think it's better to implement it in the same way or do some custom query like `SET attribute = attribute + 1`?

cheerfulstoic:
I think it would be good to mock the API, but then if we want a concurrent version to create a separate method like `concurrent_increment`.

ProGM:
@cheerfulstoic @subvertallchris Shouldn't this be closed?
However, I discovered the rails way for `concurrent_increment`, probably we should fix this:
http://apidock.com/rails/v4.2.1/ActiveRecord/CounterCache/ClassMethods/update_counters

cheerfulstoic:
Yup, good catch!  Tagging with milestone `6.1.x` in case somebody else comes along looking for this

Happy to have improvements, but I think we can close if it's working

PR #1074, opened 2015-12-18
#increment method & friends

ProGM:
I tried to implement #1067, but I'm having some issues with ActiveRel version of `concurrent_increment`, maybe because of #1073 

Any feedback is appreciated :)

### DIFF
--- a/lib/neo4j/active_node/persistence.rb
+++ b/lib/neo4j/active_node/persistence.rb
@@ -33,15 +33,8 @@ module Neo4j::ActiveNode
     # @param [Symbol, String] name of the attribute to increment
     # @param [Integer, Float] amount to increment
     def concurrent_increment!(attribute, by = 1)
-      new_attribute = Neo4j::Session.current!.query
-                      .match_nodes(n: neo_id).with(:n)
-                      .set("n.`#{attribute}` = COALESCE(n.`#{attribute}`, 0) + {by}")
-                      .params(by: by).limit(1)
-                      .pluck("n.`#{attribute}`").first
-      return false unless new_attribute
-      self[attribute] = new_attribute
-      changed_attributes.delete(attribute)
-      true
+      query_node = Neo4j::Session.query.match_nodes(n: neo_id)
+      increment_by_query! query_node, attribute, by
     end
 
     # Persist the object to the database.  Validations and Callbacks are included
--- a/lib/neo4j/active_node/persistence.rb
+++ b/lib/neo4j/active_node/persistence.rb
@@ -29,6 +29,20 @@ module Neo4j::ActiveNode
       end
     end
 
+    # Increments concurrently a numeric attribute by a centain amount
+    # @param [Symbol, String] name of the attribute to increment
+    # @param [Integer, Float] amount to increment
+    def concurrent_increment!(attribute, by = 1)
+      new_attribute = Neo4j::Session.current!.query
+                      .match(:n).where(n: {neo_id: neo_id}).with(:n)
+                      .set("n.#{attribute} = COALESCE(n.#{attribute}, 0) + #{by}")
+                      .pluck("n.#{attribute}").first
+      return false unless new_attribute
+      self[attribute] = new_attribute
+      changed_attributes.clear
+      true
+    end
+
     # Persist the object to the database.  Validations and Callbacks are included
     # by default but validation can be disabled by passing :validate => false
     # to #save!  Creates a new transaction.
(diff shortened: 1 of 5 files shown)


spec/e2e/active_model_spec.rb
```diff
@@ -436,6 +436,27 @@ def true_results?(node, verb)
       expect(person.name).to eq 'Wilson'
     end
 
+    it 'can increment an attribute' do
+      person = Person.create(name: 'andreas', age: 21)
+      person.increment(:age)
+      expect(person.age).to eq(22)
```
cheerfulstoic: I think that this could be:

    expect { person.increment(:age) }.to change { person.age }.from(22).to(21)

Though if you're not using `age_was` then I don't think the dirty tracking stuff that you implemented will be tested.  It might be better to test that separately anyway.

cheerfulstoic:
This is great, thanks!

I think that @subvertallchris fixed #1073 in master, so you could merge that in and give it another try.

I have a few other thoughts:

 * In `ActiveRel::Query` there is a `all_query` method which could replace the first part of your `ActiveRel#concurrent_increment`.  
 * Since the two concurrent methods have a lot of repeated code, I could imagine a method in `Neo4j::Shared::Persistence` which a lot of the work.  It could take a query and perhaps the variable in question (`n` in both cases, maybe we should just assume `n` for now?).  Maybe there could be an `increment` method on `Query` in `neo4j-core` that takes a node and an increment amount?

spec/e2e/active_model_spec.rb
ProGM: It was kind a way to check check that `increment` is not saving, `increment!` is saving.
What about something like:

```ruby
    it 'can increment an attribute' do
      person = Person.create(name: 'andreas', age: 21)
      expect { person.increment(:age) }.to change { person.age }.from(21).to(22)
    end

    it 'can increment an attribute and save' do
      person = Person.create(name: 'andreas', age: 21)
      expect { person.increment!(:age) }.to change { person.age }.from(21).to(22)
      expect(person).not_to be_changed
    end
```

ProGM:
I already merged #1073 , in fact tests are now passing!

* Oh, I missed it
* Yep, you're right, I have noticed that there was a lot of repetition. I'm fixing.

About the `increment` method on `neo4j-core`, it would be great!
Do you mean something like this?

```ruby
query.match_nodes(n: some_id).increment(:some_attribute, amount)
```

ProGM:
@cheerfulstoic  I improved a bit the code, following your suggestions :+1:

cheerfulstoic:
This looks great to me!  Could I get a second sign off from @subvertallchris ?

And yeah, the bit that you posted about an `increment` method in `neo4j-core` is the sort of thing I was thinking about.  The second argument could default to `1`.  Also, I think that it would be an "end cap" type method like `pluck` and `exec` which makes the query execute.  I could even imagine there would be room for other helpers.

That said, the more I think about it, the more that I think it might make sense as another gem.  It might be good to keep `neo4j-core` relatively focused on being a driver and a marshaller of objects.

cheerfulstoic:
Going to just merge and if @subvertallchris objects we can fix in master ;)

Thanks again for the contribution!

subvertallchris:
No objections!

subvertallchris:
My PR will have fixed the issue preventing that one feature's
implementation. On phone so not sure if that was included here and
failing... If no, give it a shot and it should work fine.

cheerfulstoic:
Not sure what you mean ;)

BTW @ProGM this feature has been just been released as part of 6.1.0.  Thanks!

lib/neo4j/active_node/persistence.rb
```diff
@@ -29,6 +29,14 @@ def save(*)
       end
     end
 
+    # Increments concurrently a numeric attribute by a centain amount
+    # @param [Symbol, String] name of the attribute to increment
+    # @param [Integer, Float] amount to increment
+    def concurrent_increment!(attribute, by = 1)
+      query_node = Neo4j::Session.query.match_nodes(n: neo_id)
+      increment_by_query! query_node, attribute, by
```
cheerfulstoic: I'm refactoring this in a branch, but I wanted to FYI you: When you have an object you can simply do `query_as(:n)`, so this method can just be:

```ruby
increment_by_query! query_as(:n), attribute, by
```

In my case the reason that's useful is that there are a limited number of places that `Neo4j::Session` is mentioned to reduce coupling

(Thanks again ;)

ProGM: Oh, very useful, thanks!

Issue #1085, opened 2015-12-24
undefined method `to_a' for Group(created_at: DateTime, name: String, updated_at: DateTime):Class


danpe:
After updating to Neo4jrb 6.0.4 I get the following:

```ruby
class Group 
  include Neo4j::ActiveNode
  include Neo4j::Timestamps # will give model created_at and updated_at timestamps

  property :name, type: String

  has_many :in, :subsidiaries, origin: :parent, model_class: Group
  has_one :out, :parent, type: :SUBSIDIARY_OF, model_class: Group 
  
  has_many :out, :investments, type: :INVESTED_IN, model_class: Investment
end
```

cheerfulstoic:
It looks like the line from your model is this one:

```ruby
has_many :out, :investments, type: :INVESTED_IN, model_class: Investment
```

Which makes me think that it has to do with the `Investment` model (unless the line numbers aren't quite the same).  What if you passed in a symbol for all of the `model_class` options (like `:Group` and `:Investment`)?

danpe:
You were absolutely right. it was this line.
Passing symbols solved the issue.

I think we should update the documentation: https://github.com/neo4jrb/neo4j/wiki/Neo4j%3A%3AActiveNode

cheerfulstoic:
Just updated that page, thanks!

I wonder if we should also require that `model_class` is always either a Symbol or a String.  @subvertallchris ?

subvertallchris:
I've always found the use of String or Symbol model names to be a best practice to avoid circular or not-yet-loaded dependency issues in Rails and Sinatra, respectively. Codifying it as the law of the land would fix that and help with things like this, so... Yeah, I'm ok with that. We'd probably want to enforce it across the board, so `rel_class` and `from_class`/`to_class` in ActiveRel, too.

cheerfulstoic:
Just made PR #1086

Works for `model_class` and `rel_class`.  Working on `from_class` and `to_class` now

PR #1086, opened 2015-12-25
No class arguments

cheerfulstoic:
See #1085


cheerfulstoic:
I think that'll do it!  Feel free to have a look.  I'll merge it in in a couple of days

Issue #444, opened 2014-08-28
Support for ActiveRecord style scopes


andreasronge:
http://guides.rubyonrails.org/active_record_querying.html#scopes

PR #1088, opened 2015-12-27
Allow same name for property and scope

cheerfulstoic:
Ran across this issue in the GraphGist portal app.  My code:

```ruby
class UseCase < GraphStarter::Asset
  property :name, type: String
  property :for_challenge, type: Boolean

  scope :for_challenge, -> { where(for_challenge: true) }

  has_many :in, :graph_gists, origin: :use_cases
end
```

Whenever I would try to use the `UseCase` model (or an association to it), I would get a "stack level to deep" message.  Tracked it down and it seems like we're creating an instance method for scopes, though I don't think we ever need an instance method for scopes, so I killed it.

@subvertallchris seem good?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@ All notable changes to this project will be documented in this file.
 This file should follow the standards specified on [http://keepachangelog.com/]
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Fixed
+
+- `ActiveRel#reload` relied on code written explicitly for loading nodes by ID, not rels.
+
 ## [6.0.3] - 12-18-2015
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


cheerfulstoic:
Gah, I think I branched from the wrong point.  This is supposed to go into 6.0.x but it looks like some of my changes from my other PR are there.  I can merge both in at once.  The only file that needs a look here is `scope.rb`

subvertallchris:
Looks good to me. Any reason you can think of that it was like that in the
first place? Not saying that in an accusatory way, (I don't even know who
built that... Was it part of the original implementation?) I'm just curious.

cheerfulstoic:
Looks like I refactored the code in 670a35305af68c5239525b3bda02272182d7c8b2 but it's been there since the implementation of scopes my @andreasronge in c14fec230405f78548b122702089f89547711eb5

cheerfulstoic:
The original issue was #444 which has a link to the docs for scopes.  I don't think that the docs show a method on instances, so I think we're good

cheerfulstoic:
Since this PR is conflicted I made the change manually to `6.0.x` and `master`

PR #1089, opened 2015-12-28
Run tests by Ruby 2.2.4 and 2.3.0

koic:
Ruby [2.2.4](https://www.ruby-lang.org/en/news/2015/12/16/ruby-2-2-4-released/) and [2.3.0](https://www.ruby-lang.org/en/news/2015/12/25/ruby-2-3-0-released/) has been released :santa:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ language: ruby
 cache: bundler
 sudo: false
 rvm:
+  - 2.3.0
   - 2.2.4
   - 2.0.0
   - jruby-1.7.23
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ language: ruby
 cache: bundler
 sudo: false
 rvm:
-  - 2.2.3
+  - 2.2.4
   - 2.0.0
   - jruby-1.7.23
   - jruby-9.0.4.0


cheerfulstoic:
2.3 seems like a nice minor release, thanks!

Looking at Travis, I see we're not covering 2.1.x.  Each new Ruby version we add adds three builds to Travis, but maybe we don't need to test all versions of Neo4j against each version of Ruby...

koic:
Thank you for merging :snowman: 

[IMO] It would be better to covering 2.1.x on Travis. It is true that increase run-time of Travis, but a sense of relief can be imparted to 2.1.x users.
Of course, I thought NOT necessary to non-supported Ruby versions (e.g. [EOL](https://www.ruby-lang.org/en/news/2014/07/01/eol-for-1-8-7-and-1-9-2/) Ruby versions) .

subvertallchris:
I'd be ok only testing against the oldest and newest versions that we
support. If it works against 2.0.x and 2.3.x, it should work against
everything in between. For now, I'd also like to continue testing against
2.2.x until 2.3.1 is released, too.

We had a lot of reports of terrible performance in 2.1.0 but were never
able to replicate it. I don't think we ever took the position of it being
officially unsupported but we do not encourage its use, especially when
upgrading from 2.1.x to 2.2.x comes without issues for most users.

koic:
I got it. Thank you for the thorough explanation.

PR #1094, opened 2015-12-30
Supporting Ruby marshaling for ActiveNode and ActiveRel

cheerfulstoic:


subvertallchris:
Can you throw a test in? Our untested stuff has a habit of getting refactored out or broken down the road.

cheerfulstoic:
Good idea!  I put in tests for `neo4j-core` and `neo4j` and even found some bugs along the way.  I think it's pretty solid now, but going to wait for Mssr. Travis

cheerfulstoic:
This is failing because it's not running against the `marshal` branch of `neo4j-core`.

We should be running Travis of `neo4j` against `master` of `neo4j-core` right?  Often we need to make changes to both at the same time before we release

cheerfulstoic:
Maybe the `Gemfile` should be configured so that if the `CI` environment variable is set that it runs against `neo4j-core` on GitHub master, but otherwise it doesn't specify, but we could have an `else` with the commented out line so that we could test locally

cheerfulstoic:
Just made that change, let me know what you think

subvertallchris:
We've had it setup a few ways in the past. I think it's best to leave
neo4j-core out of the Gemfile and test against a release whenever possible
because it will be a better representation of use in the wild, outside of
the gem dev process. We had a few cases in the past where Travis passed, we
released `neo4j`, but the release was broken for users because
`neo4j.gemspec` hadn't been updated to require a newer version of core. In
this case, of course, there's no way around adding it to the Gemfile and
testing against the core branch, but it'd probably be best to merge it,
release core, and update this gemspec to require that release as part of
this PR.

Sorry if that seems muddled, typing fast before bed!

subvertallchris:
Changes look great!

cheerfulstoic:
It seems like maybe we should have "Check if the `neo4j.gemspec` needs the `neo4j-core` pre-requisite updated" to our release checklist.

Maybe in Travis we could have a test against the currently released gem with `allow_failures`

PR #1095, opened 2015-12-30
Lots of changes to support the new session API in neo4j-core

cheerfulstoic:
Still have a ways to go...

Checklist of things to remember:

- [ ] Migrations / label objects should support different kinds of constraints (unique, property existance, etc...)


cheerfulstoic:
See https://github.com/neo4jrb/neo4j-core/pull/247 for some of what's going on here

ProGM:
Looking forward to see this complete 👍 👍 👍

cheerfulstoic:
Yay, this seems to be finally working!  Thanks a lot for help from @ProGM !  I think that it's reporting as not passing because of issues loading `neo4j-core` from the travis build which is what we'd get when we'd merge to master.  When we do that we can fix the `Gemfile`

@subvertallchris and @ProGM would you like to look over this PR before we merge it into `master`?  Don't forget that there is [this PR on neo4j-core](https://github.com/neo4jrb/neo4j-core/pull/247)

PR #1096, opened 2015-12-31
Remove dependency on ActiveAttr

subvertallchris:
[ActiveAttr](https://github.com/cgriego/active_attr) has been a crucial dependency since the 3.0 launch. It provided the following:

* Primitive `attributes` setters and getters, both class and instance
* Primitive type conversion
* Some compatibility with `ActiveModel`

This has been a huge help since it's obviously a lot of important, complicated stuff, but also a big problem because the library is no longer maintained. In particular, we have a lot of duplicate behavior, especially around type conversion and property management. There are also performance issues that will not be addressed.

Here, I copied/pasted/rubocopped all of the modules that we need into the gem. In the process, I made a few slight changes to remove some methods and behavior that I know we're not using and, in some cases, have been working around.

There are a few important changes immediately:

* Errors previously namespaced under `ActiveAttr` are now namespaced under `Neo4j`
* Classes namespaced under `ActiveAttr` are now namespaced under `Neo4j::Shared`
* All `ActiveAttr` typecasters were removed since they duplicated classes in `Neo4j::Shared::TypeConverters`. Our typecasters have to convert in different directions but operate at the class level, not instance, so they are faster.
* If you were testing `Boolean` and needed to call it through its namespace, `ActiveAttr::Typecasting::Boolean`, you should use `Neo4j::Shared::Boolean`.
* If you had a custom typecaster, rename its `:call` method to `:to_ruby`.

~~There is work to be done:~~

* ~~There are YARD comments referencing `ActiveAttr` all over the place. This will be an easy fix.~~
* ~~There is a ton of duplicate behavior. In particular, there's an `AttributeDefinition` class that _probably_ does most if not all of the same things as `DeclaredProperty`. There are also the primitive typecasters in the `Typecasting` namespace and then our more elaborate ones in `TypeConverters`. Oof.~~

Once this is all done, I think we can do a lot to improve performance and clarity within the gem.


cheerfulstoic:
:metal: :+1: :dancers: :beers:

subvertallchris:
I think this ready for merge. My summary in the first comment and the CHANGELOG have been updated to reflect the major changes.

raphi:
:+1:

lib/neo4j/shared/attributes.rb
```diff
@@ -0,0 +1,207 @@
+module Neo4j::Shared
```
cheerfulstoic: It would be good to have comments at the top with SOME CAPS to indicate where this file originally came from so that people can understand the history

lib/neo4j/shared/declared_property.rb
```diff
@@ -9,12 +11,46 @@ class IllegalPropertyError < StandardError; end
...
+    #   attribute_definition <=> other
+    #
+    # @param [Neo4j::Shared::AttributeDefinition, Object] other The other
+    #   attribute definition to compare with.
+    #
+    # @return [-1, 0, 1, nil]
+    def <=>(other)
+      return nil unless other.instance_of? self.class
+      return nil if name == other.name && options != other.options
+      self.name.to_s <=> other.name.to_s
```
cheerfulstoic: Maybe this should be `self.to_s <=> other.to_s` since there is a `to_s` on line 42?

lib/neo4j/shared/attributes.rb
cheerfulstoic: Oh, I see the "Originally part of ActiveAttr" bit now ;)

lib/neo4j/shared/type_converters.rb
```diff
@@ -102,7 +108,7 @@ def converted_values
         end
 
         def db_type
-          ActiveAttr::Typecasting::Boolean
+          Neo4j::Shared::Boolean
```
cheerfulstoic: I've always thought it was weird that we used these constants.  Ideally, I think, they would be symbols like `:boolean`.  But that's a discussion for later ;)

lib/neo4j/shared/attributes.rb
subvertallchris: ;-)

lib/neo4j/shared/declared_property.rb
subvertallchris: Sounds right, good point.

lib/neo4j/shared/type_converters.rb
subvertallchris: Using constants might keep some of the code a bit simpler where type conversion is concerned. It might be completely necessary or at least significantly easier to use them than symbols with QueryProxy, too, but I can't remember what all that code looks like offhand.

subvertallchris:
Let's do this.

cheerfulstoic:
http://breakouttheskinnygir.fatcow.com/wp-content/uploads/2015/12/lets-do-this-shit.jpg

PR #1102, opened 2016-01-13
Fix doc to render as code block

jessmartin:
Was rendering a a quote before:

image


cheerfulstoic:
Yay!  Awesome, thanks!

Issue #1107, opened 2016-01-18
Unable to find factory for #<Neo4j::ActiveRel::RelatedNode:0x0000000b8977a8>


djvs:
Trace:

    /usr/local/rvm/gems/ruby-2.2.0/gems/neo4j-6.1.0/lib/neo4j/shared/query_factory.rb:20:in `create'
    /usr/local/rvm/gems/ruby-2.2.0/gems/neo4j-6.1.0/lib/neo4j/active_rel/persistence/query_factory.rb:67:in `query_factory'
    /usr/local/rvm/gems/ruby-2.2.0/gems/neo4j-6.1.0/lib/neo4j/active_rel/persistence/query_factory.rb:60:in `block in iterative_query'
    /usr/local/rvm/gems/ruby-2.2.0/gems/neo4j-6.1.0/lib/neo4j/active_rel/persistence/query_factory.rb:58:in `each'
    /usr/local/rvm/gems/ruby-2.2.0/gems/neo4j-6.1.0/lib/neo4j/active_rel/persistence/query_factory.rb:58:in `inject'
    /usr/local/rvm/gems/ruby-2.2.0/gems/neo4j-6.1.0/lib/neo4j/active_rel/persistence/query_factory.rb:58:in `iterative_query'
    /usr/local/rvm/gems/ruby-2.2.0/gems/neo4j-6.1.0/lib/neo4j/active_rel/persistence/query_factory.rb:23:in `block in build!'
    /usr/local/rvm/gems/ruby-2.2.0/gems/neo4j-6.1.0/lib/neo4j/active_rel/persistence/query_factory.rb:41:in `block (2 levels) in node_before_callbacks!'
    /usr/local/rvm/gems/ruby-2.2.0/gems/neo4j-6.1.0/lib/neo4j/shared/callbacks.rb:41:in `conditional_callback'
    /usr/local/rvm/gems/ruby-2.2.0/gems/neo4j-6.1.0/lib/neo4j/active_rel/related_node.rb:62:in `method_missing'
    /usr/local/rvm/gems/ruby-2.2.0/gems/neo4j-6.1.0/lib/neo4j/active_rel/persistence/query_factory.rb:40:in `block in node_before_callbacks!'
    /usr/local/rvm/gems/ruby-2.2.0/gems/neo4j-6.1.0/lib/neo4j/shared/callbacks.rb:41:in `conditional_callback'
[...]

PR #1108, opened 2016-01-18
Fix for relationship from/to_node class specificity issues

djvs:
Re: https://github.com/neo4jrb/neo4j/issues/1107#issuecomment-172603751

Needs testing

### DIFF
--- a/lib/neo4j/shared/query_factory.rb
+++ b/lib/neo4j/shared/query_factory.rb
@@ -11,14 +11,13 @@ module Neo4j::Shared
     end
 
     def self.create(graph_object, identifier)
-      factory = case graph_object
-                when Neo4j::ActiveNode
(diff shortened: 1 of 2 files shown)


cheerfulstoic:
FYI this build is failing in rubocop because you have some double quoted strings when you don't need to.  You can either run `bundle exec rubocop` to see these places, or run `bundle exec rubocop -a` to just have them be fixed by rubocop

@subvertallchris is  more familiar with the issue so I'll let him comment on the substance, but on the style I like the syntax of `factory = if ...` so that you don't need to have multiple `factory = ` statements.

cheerfulstoic:
(And thanks for the PR!)

subvertallchris:
Gonna need tests in there for the bug you're trying to fix, too.

Issue #926, opened 2015-08-18
`update_all` for models / scopes


cheerfulstoic:

ProGM:
@cheerfulstoic I'm trying to implement a temporary replacement. What could be a good implementation for this?

I thought about something like:
```ruby
def update_all(params = {})
  query_as(:c).set(c: params).exec
end
```

or maybe:

```ruby
def update_all(updates, bindings = {})
  if updates.is_a? Hash
    query_as(:n).set(n: values).exec
  elsif updates.is_a? String
    query_as(:n).set(values).params(bindings).exec
  else
    raise ArgumentError, 'Invalid parameters.'
  end
end
```

cheerfulstoic:
I think you're on a good path.  If you come up with a good final solution I'm happy to throw it in the gem (or you can make a PR and get credit ;)  I have a few thoughts:

 * First, I would call `bindings` `parameters` or `params` to be consistent.
 * I would probably use a `case` statement for the second implementation
 * These should all start with the `all` method so that you could run `Model.where('something').update_all()` and it should only update the objects that you have scoped to.

Maybe something like:

```ruby
def update_all(updates, params = {})
  query = all.query_as(:n)

  case updates
  when Hash
    query.set(n: values).exec
  when String
    query.set(values).params(params).exec
  else
    raise ArgumentError, 'Invalid parameters.'
  end
end
```

There's also the argument to be made that Cypher makes this sort of thing really easy with `SET`, so that maybe users should just choose whichever of these paths that they want manually.  In the end, though, it's good to have compatibility with ActiveRecord so that plugins can work more easily between the two.

Let me know if you have any other questions!

PR #1113, opened 2016-01-20
Added #update_all method to ActiveNode and QueryProxy.

ProGM:
@cheerfulstoic 
I tried to implement #926 

Should this work also for `ActiveRel`?
I know you deprecated all query methods for `ActiveRel`, so I don't know if it could make sense.

### DIFF
--- a/lib/neo4j/active_node/labels.rb
+++ b/lib/neo4j/active_node/labels.rb
@@ -78,6 +78,8 @@ module Neo4j
       module ClassMethods
         include Neo4j::ActiveNode::QueryMethods
 
+        delegate :update_all, to: :all
+
         # Returns the object with the specified neo4j id.
         # @param [String,Integer] id of node to find
         def find(id)
--- a/lib/neo4j/active_node/query/query_proxy_methods.rb
+++ b/lib/neo4j/active_node/query/query_proxy_methods.rb
@@ -1,7 +1,9 @@
 module Neo4j
   module ActiveNode
     module Query
+      # rubocop:disable Metrics/ModuleLength
       module QueryProxyMethods
+        # rubocop:enable Metrics/ModuleLength
         class InvalidParameterError < StandardError; end
         FIRST = 'HEAD'
         LAST = 'LAST'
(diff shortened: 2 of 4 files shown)


cheerfulstoic:
Awesome, thanks!  I'll look more later, but some quick feedback:

 * I don't remember if we cache the result of `count` on a query proxy chain, but if we do then the `expect().not_to change` wouldn't work.  Did you check that?
 * The `QueryProxyMethods` module gets mixed into `QueryProxy` so I wonder why `Model.update_all` is working.  Normally we can define a class method like `update_all` on `ActiveNode` and `QueryProxy` will take care of doing `method_missing` magic to delegate the method call from `QueryProxy` chains.

Regarding updating relationships, that's a really good question.  Perhaps we should define something like `update_all_rels` (which actually would be defined on `QueryProxy` and not `ActiveRel`, I think) so that you could call it to update the `rel_var`.  Come to think of it there is a variable that you should probably be using so that `Model.association(:var).update_all()` works because you can't use `n`.  `QueryProxy` objects have an abliity to find the current variable for the node and the relationship.

I need to run now, but this is great!  Thanks so much!

ProGM:
Great!
* To be sure that the query is updating thing that shouldn't be updated... but maybe I can remove it.
* Check here, there's a `delegate :update_all, to: :all` https://github.com/neo4jrb/neo4j/pull/1113/files#diff-dfd59f2a8fc368af1d19efadf9411bb8R81

> Come to think of it there is a variable that you should probably be using so that Model.association(:var).update_all() works because you can't use n. QueryProxy objects have an abliity to find the current variable for the node and the relationship.

So I may replace `n` with "the current variable name" in the `update_all`, right? So this should be valid also for `ActiveNode`...

ProGM:
News in the last commit:
* I implemented `person = Person.friends(:f).update_all_rels(friendship: 100)` :+1: 
* I moved all tests from `active_node_spec.rb` to `query_proxy_methods_spec.rb`... Because it makes more sense.
* I monkey patched a rubocop offense adding a `# rubocop:disable Metrics/ModuleLength`

cheerfulstoic:
This is awesome and beautiful (with lots of tests), thanks so much! We might decide to move the `update_all` method into the `ActiveNode` space at some point to match with the existing `method_missing` behavior, but this works great for now.

Hopefully I can finish my other refactoring in the near future and we can get a release out with this in it!

ProGM:
Great! :+1: Do you think you can close #926 ?

cheerfulstoic:
We generally leave issues open until they're available in a rubygems gem, so I'll close it when we release.

Thanks!

Issue #1104, opened 2016-01-16
Why is Neo4j::ActiveNode::Labels::RecordNotFound not inheriting from Neo4j::RecordNotFound?


ProGM:
By default, ActiveRecord fails to find object raising an `ActiveRecord::RecordNotFound`. In rails, it's rescued by default with a 404 page.

Instead, `Neo4j::RecordNotFound` generates a more generic "internal server error".

So I tried to catch it with `rescue_with`:

```ruby
rescue_from Neo4j::RecordNotFound do
  render text: 'Not Found', status: 404
end
```

But, it does not catches all `Neo4j::ActiveNode::Labels::RecordNotFound` errors.

I would expect that `Neo4j::RecordNotFound` was the parent of `Neo4j::ActiveNode::Labels::RecordNotFound`, or to be both child of `Neo4j:: Neo4jrbError`.

Do you plan to add a default behavior for all neo4j errors or add some infos about it in the docs?

subvertallchris:
The reorganization of error classes is something that I've been meaning to
do for a little while but it's an issue of time and priorities. If you'd
like to spearhead that or get it started, we'd love your input!

subvertallchris:
Do we want to close this now?

ProGM:
Uhm... Do you think that errors need more refactor?

For me it's ok...

subvertallchris:
I think you made great progress, I'm ok with closing. We can always revisit if it's necessary later.

ProGM:
Great :+1:

cheerfulstoic:
:+1:

PR #1114, opened 2016-01-20
Neo4jrb errors refactors

ProGM:
@subvertallchris 
I tried to improve a bit errors, referring to #1104 

I cherry-picked some ideas from `ActiveRecord` and `MongoMapper`. Any suggestion is welcome!


subvertallchris:
I like this a lot. Do you think any of it should be moved into https://github.com/neo4jrb/neo4j/blob/master/lib/neo4j/errors.rb? ActiveRecord uses one `errors.rb`, https://github.com/rails/rails/blob/master/activerecord/lib/active_record/errors.rb, though I'm not sure if that's for _all_ of their errors or just some.

ProGM:
I found a few errors of ActiveRecord that are not in `errors.rb`, but they are just a couple:
https://github.com/rails/rails/blob/a461dbd637791d3bd399007f234a1ca5c919e199/activerecord/lib/active_record/associations.rb#L17
https://github.com/rails/rails/blob/428d47adfed8d6aa7b21aec2bf5ad890961c9de3/activerecord/lib/active_record/validations.rb#L13
So we can assume that all "main" errors in active record are in the `errors.rb` file.

I'm not sure how to deal with this. I think I don't have a complete knowledge about all neo4j errors, so please help me with this.

For example:

*  I can't find any reference about this error:
https://github.com/neo4jrb/neo4j/pull/1114/files#diff-dfd59f2a8fc368af1d19efadf9411bb8R23
Is it an old entry?

* All children of `Neo4j::InvalidParameterError` may be replaced with the more generic `Neo4j::InvalidParameterError`. Example:
https://github.com/neo4jrb/neo4j/pull/1114/files#diff-c2bb410db73f7a12d959186c21eb7bd3R4

* Instead `UnsetRelatedNodeError` can stay there, since it's quite specific: https://github.com/neo4jrb/neo4j/pull/1114/files#diff-fcd30cb21891b2b381a6385f88287956R7

* Finally, I'm not sure how to deal with this: https://github.com/neo4jrb/neo4j/pull/1114/files#diff-d883afb716d52a87e075fde8d7dbe1f7R12 . Should it be a child of Neo4j errors or not?

subvertallchris:
That first one, `InvalidQueryError`, came from https://github.com/neo4jrb/neo4j/commit/c7b964c8e4da4722c025540282def1f49272869a. It looks like it was for a very specific condition which no longer exists, so I think we can axe it.

The second one can probably be moved into `Errors` and they can all use that one definition.

:+1: where `UnsetRelatedNodeError` is concerned.

`UndefinedPropertyError` should be a child of Neo4j errors, if to identify it as coming from us if nothing else.

How's this sound? Thanks again for looking into all this, it's a huge help.

ProGM:
Ok, I fixed everything you said.

I also merged with master, since there were a few conflicts!

Please take a look :mag:

cheerfulstoic:
Thanks from me as well!  This looks great!

subvertallchris:
This is coooool! Would you just throw details about the changes into the CHANGELOG and we can roll with it?

ProGM:
Done. Maybe it's a bit verbose...

subvertallchris:
I don't think you will ever hear me or Brian complain that something is verbose. ;-)
I'll do the merge once Travis signs off.

subvertallchris:
I can't believe how long Travis takes now. WTF.

cheerfulstoic:
Hrmm, I imagine we don't need to check every version of Ruby against every version of Neo4j.  Or maybe there would be a way to have build priorities so that certain jobs are required to pass and the other ones can run whenever.  It would be particularly nice if those required jobs took priority over jobs from previous builds which haven't been pulled up yet.

cheerfulstoic:
Regarding the errors, I hate to bring things up semi-last minute, but this is something that people might start depending on and so I wanted to talk about what I was thinking.

The definition of "Neo4j.rb" has wavered somewhat.  In the past I think in referred to the `neo4j` gem, though it always made sense to be to call the various gems that we maintain (like `neo4j`, `neo4j-core`, and `neo4j-rake_tasks` most importantly) the Neo4j.rb project.  So in this case I'm not sure if `Neo4jrbError` makes sense.

I could see perhaps `Neo4j::StandardError`.  Alternatively as part of my refactoring I created a `Neo4j::ActiveBase` module which deals with logic that is global to `ActiveNode` and `ActiveRel`, so maybe it could be in there?

Thoughts?

cheerfulstoic:
Regarding travis, this might be something to consider:

https://docs.travis-ci.com/user/customizing-the-build/#Fast-Finishing

Also, of course, my idea was not original:

https://github.com/travis-ci/travis-ci/issues/2566

ProGM:
@cheerfulstoic 
Some random thoughts, maybe they could be useful:

I understand what you mean... but I really don't like `Neo4j::StandardError`. I think it's a confusing to have an error name for a gem that have the name of the most general error in ruby.

In MongoMapper, for example, they simply use `MongoMapper::Error`. Maybe we can follow the same convention.
https://github.com/mongomapper/mongomapper/blob/master/lib/mongo_mapper/exceptions.rb

`Neo4j::ActiveBaseError` instead, could be ok, but the term `ActiveBase` is not (yet) in the "vocabulary" of a neo4jrb user, so it's kind a "unknown" word.

cheerfulstoic:
Yeah, that's a good point.  I'm not sure what I was thinking with `StandardError` ;)

I'm fine with `Neo4j::Error`, actually.  Does that work for you?

ProGM:
For me it's ok. I'll rename them all and fix conflict, if you're all agree!

cheerfulstoic:
Yep, if you made that change I would be :smiley: !

ProGM:
Done! Take a look :wink:

cheerfulstoic:
LGTM, thanks!  @subvertallchris I'm :+1: to merge this.  Feel free to click the big green button if you're good too

subvertallchris:
Rad, I'll check it out when I get to the office.

subvertallchris:
Pulling the trigger. @ProGM, great work!

PR #1119, opened 2016-01-25
Use class_path when generating a model

michaeldelorenzo:
**Problem:** 
When creating a new model using the generator with Rails, namespaces are ignored. For example, you would expect `rails g model namespaces/my_model` to create a model file at the following location,  `app/models/namespaces/my_model.rb`. But _instead_, it is created at `app/models/my_model.rb`.

**Fix:**
Include the `class_path` variable when generating the file's path. Taken from [ActiveRecord's model generator](https://github.com/rails/rails/blob/571b090a85b09422b4ed2744383b90ada433a2f0/activerecord/lib/rails/generators/active_record/model/model_generator.rb#L23-L25).

Michael De Lorenzo
delorenzo.michael [at] gmail.com

### DIFF
--- a/lib/rails/generators/neo4j/model/model_generator.rb
+++ b/lib/rails/generators/neo4j/model/model_generator.rb
@@ -12,7 +12,7 @@ class Neo4j::Generators::ModelGenerator < Neo4j::Generators::Base #:nodoc:
   class_option :has_many,   type: :array,  desc: 'A list of has_many relationships'
 
   def create_model_file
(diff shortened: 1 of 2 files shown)


cheerfulstoic:
This is awesome, thanks so much!

We hope to be making a release from master in the near future, so keep an eye out for that

cheerfulstoic:
Actually, this is something that should be released sooner.  Just released it in `6.1.4`.  Thanks again!

Issue #1017, opened 2015-10-28
`find_or_create` (uses merge) should not change anything on match


subvertallchris:
This:

```ruby
      def find_or_create(find_attributes, set_attributes = {})
        on_create_attributes = set_attributes.merge(on_create_props(find_attributes))
        on_match_attributes =  set_attributes.merge(on_match_props)
        neo4j_session.query.merge(n: {self.mapped_label_names => find_attributes})
          .on_create_set(n: on_create_attributes).on_match_set(n: on_match_attributes)
          .pluck(:n).first
      end
```

Should be this:

```ruby
      def find_or_create(find_attributes, set_attributes = {})
        on_create_attributes = set_attributes.merge(on_create_props(find_attributes))
        neo4j_session.query.merge(n: {self.mapped_label_names => find_attributes})
          .on_create_set(n: on_create_attributes)
          .pluck(:n).first
      end
```

venkatd:
This sounds useful. I have some confusion about how merge works...

Suppose I have an Account node with
```
source_name: google
source_id: 1234
email: bob@hotmail.com
```

Then I later want to do a find_or_create like so:
```
Account.find_or_create({source_name: 'google', source_id: '1234'}, {email: 'coolbob@gmail.com'})
```

What would be the behavior of find-or_create?

subvertallchris:
In a single Cypher query, it will `find` using the first hash and if no match is found, it will create using that and the contents of the second hash.

leviwilson:
@venkatd _currently_ it will update all of your properties specified I think. So your `bob@hotmail.com` will get blown away with `coolbob@gmail.com`. Which is what I'm seeing anyway.

I _think_ the behavior should be this:

## `CREATE`
Takes both your "find" properties and your "set" properties and `Hash.merge` them to compose what your node should have.

## Already Exists
Nothing.

That seem legit to you in how that method should behave?

cheerfulstoic:
Oh, interesting... I came here to fix this issue but I think because #1030 says "fixes #1017" that when I closed the PR it closed the original issue.  Neat...

cheerfulstoic:
Released in 5.2.14, BTW

venkatd:
@leviwilson I thing that behavior makes sense with `find_or_create`. I do find myself frequently wanting a `find_or_update` so I think it would be helpful to have the a `find_or_update` method. I'll add it when I'm back from vacation.

cheerfulstoic:
[...]

PR #1123, opened 2016-01-28
Merge on match

leviwilson:
## Description
`find_or_create` is nice, and #1017 modifies it so it only does what it says; but sometimes I find myself needing [update-only attributes](https://github.com/neo4jrb/neo4j/issues/1017#issuecomment-154123488).

This pull modifies `merge` so that you can (optionally) separate your match-only / update-only attributes.

Feedback welcome!

### DIFF
--- a/lib/neo4j/active_node/persistence.rb
+++ b/lib/neo4j/active_node/persistence.rb
@@ -128,9 +128,12 @@ module Neo4j::ActiveNode
         end
       end
 
-      def merge(match_attributes, update_attributes = {})
-        find_or_create_query(match_attributes, update_attributes)
-          .on_match_set(n: on_match_props(update_attributes))
+      def merge(match_attributes, optional_attrs = {})
+        optional_attrs.default = {}
+        on_create_set, set_attrs = optional_attrs.values_at :on_create, :set
+
+        neo4j_session.query.merge(n: {self.mapped_label_names => match_attributes})
+          .on_create_set(n: on_create_props(on_create_set.merge(set_attrs)))
           .pluck(:n).first
       end
 
--- a/lib/neo4j/active_node/persistence.rb
+++ b/lib/neo4j/active_node/persistence.rb
@@ -167,7 +167,7 @@ module Neo4j::ActiveNode
       end
 
       def on_match_props(match_attributes)
-        match_attributes.tap { |props| props[:updated_at] = DateTime.now.to_i if attributes_nil_hash.key?('updated_at'.freeze) }
+        match_attributes.tap { |props| props[:updated_at] = DateTime.now.to_i if attributes_nil_hash.key?('updated_at') }
       end
     end
   end
--- a/lib/neo4j/active_node/persistence.rb
+++ b/lib/neo4j/active_node/persistence.rb
@@ -133,11 +133,11 @@ module Neo4j::ActiveNode
         optional_attrs.assert_valid_keys *options
 
         optional_attrs.default = {}
-        on_create_set, on_match_set, set_attrs = optional_attrs.values_at *options
+        on_create_attrs, on_match_attrs, set_attrs = optional_attrs.values_at *options
 
         neo4j_session.query.merge(n: {self.mapped_label_names => match_attributes})
-          .on_create_set(n: on_create_props(on_create_set.merge(set_attrs)))
-          .on_match_set(n: on_match_props(on_match_set.merge(set_attrs)))
+          .on_create_set(n: on_create_props(on_create_attrs.merge(set_attrs)))
+          .on_match_set(n: on_match_props(on_match_attrs.merge(set_attrs)))
           .pluck(:n).first
       end
 
--- a/lib/neo4j/active_node/persistence.rb
+++ b/lib/neo4j/active_node/persistence.rb
@@ -129,8 +129,11 @@ module Neo4j::ActiveNode
       end
 
       def merge(match_attributes, optional_attrs = {})
+        options = [:on_create, :on_match, :set]
+        optional_attrs.assert_valid_keys *options
+
         optional_attrs.default = {}
-        on_create_set, on_match_set, set_attrs = optional_attrs.values_at :on_create, :on_match, :set
+        on_create_set, on_match_set, set_attrs = optional_attrs.values_at *options
 
         neo4j_session.query.merge(n: {self.mapped_label_names => match_attributes})
           .on_create_set(n: on_create_props(on_create_set.merge(set_attrs)))
--- a/lib/neo4j/active_node/persistence.rb
+++ b/lib/neo4j/active_node/persistence.rb
@@ -130,10 +130,10 @@ module Neo4j::ActiveNode
 
       def merge(match_attributes, optional_attrs = {})
         options = [:on_create, :on_match, :set]
-        optional_attrs.assert_valid_keys *options
+        optional_attrs.assert_valid_keys(*options)
 
         optional_attrs.default = {}
-        on_create_attrs, on_match_attrs, set_attrs = optional_attrs.values_at *options
+        on_create_attrs, on_match_attrs, set_attrs = optional_attrs.values_at(*options)
 
         neo4j_session.query.merge(n: {self.mapped_label_names => match_attributes})
           .on_create_set(n: on_create_props(on_create_attrs))
--- a/lib/neo4j/active_node/persistence.rb
+++ b/lib/neo4j/active_node/persistence.rb
@@ -139,7 +139,10 @@ module Neo4j::ActiveNode
       end
 
       def find_or_create(find_attributes, set_attributes = {})
-        find_or_create_query(find_attributes, set_attributes)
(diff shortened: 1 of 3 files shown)


lib/neo4j/active_node/persistence.rb
```diff
@@ -128,17 +128,14 @@ def create!(*args)
         end
       end
 
-      def merge(attributes)
-        neo4j_session.query.merge(n: {self.mapped_label_names => attributes})
-          .on_create_set(n: on_create_props(attributes))
-          .on_match_set(n: on_match_props)
+      def merge(match_attributes, update_attributes = {})
+        find_or_create_query(match_attributes, update_attributes)
+          .on_match_set(n: on_match_props(update_attributes))
```
cheerfulstoic: I think people should be able to define attributes for `on_create_set`

leviwilson: This does that, right? the `match_attributes` are passed down in the same manner as `find_or_create` does, unless I'm missing something? Have an example use-case?

cheerfulstoic: If I'm reading it right, the `match_attributes` is used to determine which attributes are `MATCH`ed as part of the `MERGE`, but there's nothing that gives the user control over the `ON MATCH SET`

cheerfulstoic: Sorry, the `ON CREATE SET`, I mean

leviwilson: Ahh, that's true. It assumes that you want the `match_attributes.merge update_attributes` in the event that it doesn't exist. Do you have an example of how you'd like for that to work?

lib/neo4j/active_node/persistence.rb
```diff
@@ -158,12 +155,19 @@ def load_entity(id)
...
+      end
+
       def on_create_props(find_attributes)
         find_attributes.merge(self.new(find_attributes).props_for_create)
       end
 
-      def on_match_props
-        {}.tap { |props| props[:updated_at] = DateTime.now.to_i if attributes_nil_hash.key?('updated_at'.freeze) }
+      def on_match_props(match_attributes)
+        match_attributes.tap { |props| props[:updated_at] = DateTime.now.to_i if attributes_nil_hash.key?('updated_at'.freeze) }
```
subvertallchris: That `freeze` call can be dropped, I don't know why I let that sneak in originally.

subvertallchris: You should also track down the process by which `updated_at` gets sets normally and reuse that instead of doing it manually. The user might have set their timestamps to something other than `DateTime` and this will screw them up.

leviwilson: I can drop the `freeze`, no problem.

To be clear about the "`DateTime` and this will screw them up" part; this is the same method that was being called before in the `merge`. The only change is `{}` :arrow_right: `match_attributes` instead. Functionally it's the same with regard to `updated_at`, correct?

cheerfulstoic:
Finally sitting down to think about this ;)

First note: if I'm reading this right the `updated_at` is always being set to the current time whenever you do a `merge`.  Do we want that?

I feel like I'm thrown because the methods are shifted from what I would expect that they would be doing.  I feel like:

 * `find_or_create_by` shouldn't exist
 * `find_or_create` should be doing what `find_or_create_by` is doing (because that's what `ActiveRecord` does)
 * `merge` should be a generic convenience to doing a `MERGE` for your model.  That would mean giving people the option to define what they want for `ON CREATE SET`, `ON MATCH SET`, and `SET`

I think I'm also thrown by the changing of the variable name.  It comes into `merge` as `update_attributes` but then gets passed into `find_or_create_query` as `set_attributes`.

Looking at it further, I think I see what you're were saying to me earlier that in the `ActiveNode` case we just want to define a set of attributes that we can be sure are set the way we've specified, regardless of match or create.  In that case I don't think you'd want `ON CREATE SET` or `ON MATCH SET` but rather just `SET` which always runs.

What about having `match` have the first argument be the arguments to `MERGE` on, as currently, but then having the second argument be a `Hash` of options with the keys `on_create_set`, `on_match_set`, and `set` allowed (failing with `ArgumentError` if an undefined key is specified).  Then we could just pass those three hashes into calls to `on_create_set`, `on_match_set`, and `set` in the `Query` chain, merging with the default attributes from the model.  We'd probably need to add to the chain only if the keys were non-empty like;

```ruby
query = query.on_create_set(options[:on_create_set]) unless options[:on_create_set].empty?
```

Optionally you could do a `on_create_set, on_updated_set, set = options.values_at(:on_create_set, :on_update_set, :set)` to make that a bit lighter for each of those lines.

Additional note: we should probably be using the value of `node_var` rather than hard-coding `n` as the variable in our `MERGE`

leviwilson:
:+1: I'll get to work on it

leviwilson:
@cheerfulstoic I went ahead and made the changes to add `on_create`, `on_match` and `set` as an optional `{}` to pass into `.merge`. I've re-worked the specs for `.merge` as well to make sure we're covering all of our bases still.

I tried to do:

```ruby
session.query.merge(n: {...})
  .on_create_set(n: {...})
  .on_match_set(n: {...})
  .set(n: {...})
```

But the query that is generated seems to have the `SET` out of order from the `ON MATCH` and `ON CREATE` part of the cypher and it didn't like it; hence why I'm doing a `.merge(set_attrs)` in either case :/

Wasn't sure how `node_var` applied here since the `.merge` is a once-and-done, so I didn't do it. I also didn't know if I should do anything with `find_or_create` (like add a `deprecate`?).

Feedback welcome.

leviwilson:
I'm unsure why all of the jRuby builds are breaking :-/

leviwilson:
I pulled jRuby and it ran fine for me locally; only issue was with the `lazy` call for `Enumerable`; I switched that and it's still failing. Any assistance would be appreciated as I'm unsure what to look into.

cheerfulstoic:
Don't worry about `find_or_create`, I was saying that for the discussion.

Regarding the `node_var`, the reason I was thinking about it is because somebody might have a previous scope/chain which they then call `merge` on.  I haven't thought through the consequences of what that would mean, though, so we can leave it for the moment ;)

I'll look at the code now, thanks!

lib/neo4j/active_node/persistence.rb
```diff
@@ -128,15 +128,19 @@ def create!(*args)
         end
       end
 
-      def merge(attributes)
-        neo4j_session.query.merge(n: {self.mapped_label_names => attributes})
-          .on_create_set(n: on_create_props(attributes))
-          .on_match_set(n: on_match_props)
+      def merge(match_attributes, optional_attrs = {})
+        optional_attrs.default = {}
+        on_create_set, on_match_set, set_attrs = optional_attrs.values_at :on_create, :on_match, :set
```
cheerfulstoic: To keep things consistent I would call the attributes `on_create_attrs`, `on_match_attrs`, and `set_attrs`.

Also, we should have something like:

```ruby
invalid_options = optional_attrs.keys - [:on_create, :on_match, :set]
fail ArgumentError, "Invalid options: #{invalid_options.inspect}"
```

Honestly this happens so much we should probably have a helper method.  ... a couple of minutes pass ... looks like ActiveSupport already does that!

http://apidock.com/rails/Hash/assert_valid_keys

We're including `ActiveSupport` in the gem so you should just need to:

```ruby
require 'active_support/core_ext/hash/keys'
```

cheerfulstoic:
I made some comments for you to look at, but otherwise it looks great!  I suspect that you could fix the `SET` thing by doing `on_create_set().on_match_set().break.set()`.  If that works it would be a nice change because it would make things more as-expected for people looking at the resulting Cypher queries.

If @subvertallchris is good with it I'll just merge it into master with the jRuby bug and fix it afterward because I can't test it on my machine without checking out your fork.  ANGTFT...

Thanks!

cheerfulstoic:
Oh, and Rubocop is failing

cheerfulstoic:
Just looked closer at the error and I would say that since the order doesn't matter you should use `match_array` instead of `eq`

leviwilson:
I've updated the PR with the above comments; we'll see about jRuby

cheerfulstoic:
Awesome, I'll merge let @subvertallchris object if he cares to later

cheerfulstoic:
@leviwilson pointed out that this `PR` makes the code in `find_or_create` unneeded.  I know you don't like API changes, but what do you say to a deprecation cycle to do the following:

 * In 7.0:
   * advise people using `find_or_create` to use `merge` instead (with a link to a guide on what changes would need to be made)
   * advise people that `find_or_create_by` and `find_or_create_by!` are going away in 8.0
 * In 8.0 switch `find_or_create` to use the code from `find_or_create_by` to `find_or_create` and make a new `find_or_create!` which replaces `find_or_create_by!` (all of this to match with `ActiveRecord`'s API)

subvertallchris:
This all sounds good. I'll check it in the morning and let you guys know if
I have any feedback!

Issue #1128, opened 2016-02-11
Bug with property serializers


ProGM:
I was trying to implement enums #1082 , and I discovered a bug with `serializers`.

Here's a test that demonstrates it:

```ruby
describe Neo4j::ActiveNode do
  before(:each) do
    # This serializer adds a text when the data is saved on the db,
    # and removes it when deserializing
    class MySerializer
      def initialize(text)
        @text = text
      end

      def converted?(value)
        value.is_a?(db_type)
      end

      def db_type
        String
      end

      def convert_type
        Symbol
      end

      def to_ruby(value)
        value.gsub(@text, '').to_sym if value
      end

      alias_method :call, :to_ruby

      def to_db(value)
        "#{value}#{@text}" if value
      end
    end

    stub_active_node_class('MyData') do
      property :polite_string, serializer: MySerializer.new(', sir.')
      property :happy_string, serializer: MySerializer.new(' :)')
    end
  end

  it 'serializes correctly' do
    data = MyData.new
    data.polite_string = :hello
    data.happy_string = :hello
    data.save!
    expect(MyData.as(:d).pluck('d.polite_string, d.happy_string')).to eq([['hello, sir.', 'hello :)']])
  end
end
```

Here's the test output:

```ruby
1) Neo4j::ActiveNod serializes correctly
     Failure/Error: expect(MyData.as(:d).pluck('d.polite_string, d.happy_string')).to eq([['hello, sir.', 'hello :)']])

       expected: [["hello, sir.", "hello :)"]]
            got: [["hello :)", "hello :)"]]

       (compared using ==)
     # ./spec/e2e/enum_spec.rb:50:in `block (2 levels) in <top (required)>'
```

EDIT:
After exploring the code, I noticed that the `convert_type` output is used as a key to determine what is the correct TypeConverter for a certain type, and it's used for ALL properties.
https://github.com/neo4jrb/neo4j/blob/77190ab416d219f230dc86eb71e3b7b92df65b0b/lib/neo4j/shared/declared_property.rb#L115
It's a bit strange, since I'm passing the `serializer` setting only to a specific property, not as a global setting.

My question is: it's really a bug? or a design choice?

I'm a bit confused.

cheerfulstoic:
I'm not very familiar with serializers.  @subvertallchris is this an ActiveAttr thing, or something we put in?

subvertallchris:
It's ours, I will look over this today.

subvertallchris:
[...]

Issue #1082, opened 2015-12-23
Support `enum` like in ActiveRecord


cheerfulstoic:
See:

http://edgeapi.rubyonrails.org/classes/ActiveRecord/Enum.html
http://stackoverflow.com/questions/34420013/how-can-i-implement-enum-in-a-graph-database-like-neo4j

Items:

 - [ ] Validation that the properties is one of the values
 - [ ] Validation that enum values don't conflict with each other
 - [ ] `?` and `!` methods
 - [ ] Automatic scopes
 - [ ] Making sure that enums are supported in `where` and `rel_where` (including `Array` values)
 - [ ] Documentation (with information about prefix, suffix, and index options)

ProGM:
@cheerfulstoic  @subvertallchris 
Hello, I made an enum implementation using `TypeConverters`.

If you like it, I can make a pull request ;)

Usage example:
```ruby
class File
  include   include Neo4j::Timestamps
  include Neo4jEnum
    enum type: [:image, :video, :unknown]
end

f = File.new
f.type # => :image
f.type = :video
f.video? # => true
f.unknown!
f.type # => :unknown
f.save
#  CYPHER 5ms CREATE (n:`File` {props}) RETURN ID(n) | {:props=>{:uuid=>"fbab3d84-70da-417d-9b26-d4daecd1d628", :type=>2}}

# Finders:
File.unknown # => finds all Files where type is :unknown
File.where(type: :unknown) # => the same
File.where(type: [:video, :unknown]) # => NOT WORKING
```

Code:
```ruby
module Neo4jEnum
  extend ActiveSupport::Concern

  class EnumConverter
    def initialize(enum_list)
      @enum_list = enum_list
    end

    def converted?(value)
      value.is_a?(db_type)
    end

    def db_type
      Integer
    end

    def convert_type
      Symbol
    end

    def to_ruby(value)
      @enum_list[value] unless value.nil?
    end

    alias_method :call, :to_ruby

    def to_db(value)
      @enum_list.index(value.to_s.to_sym) || 0
    end
  end

  module ClassMethods
    attr_reader :neo4j_enum_data

    def enum(options = {})
      options.each do |property_name, enum_list|
        @neo4j_enum_data ||= {}
        @neo4j_enum_data[property_name] = enum_list
        property property_name, serializer: EnumConverter.new(enum_list),
                                default: enum_list.first
        define_enum_methods(property_name, enum_list)
      end
    end

    private

[...]

PR #1129, opened 2016-02-11
`enum` for Neo4j

ProGM:
First implementation for #1082 

I'm copying here the checklist:

* ~~Validation that the properties is one of the values~~ <-- Can be done by the user easily if needed
* [x] Validation that enum values don't conflict with each other
* [x] ? and ! methods
* [x] Automatic scopes
* [x] Making sure that enums are supported in where and rel_where (including Array values)
* [x] Documentation (with information about prefix, suffix, and index options)
* [x] if you define enum `type`, a class method `types` that return all the types should be defined

I had an issue with #1128 , I solved by defining a type `Enum` on the fly.
Please take a look to the `EnumConverter` class.


lib/neo4j/shared/enum.rb
```diff
@@ -0,0 +1,120 @@
...
+  module Enum
+    extend ActiveSupport::Concern
+
+    class ConflictingEnumMethodError < Neo4j::Error; end
+
+    module ClassMethods
+      attr_reader :neo4j_enum_data
+
+      def enum(parameters = {})
+        options, parameters = *split_options_and_parameters(parameters)
```
cheerfulstoic: Could we just have all of the parameter be all of the arguments except for an optional last options argument?  ActiveSupport even has a (not very complicated) method for this:

http://api.rubyonrails.org/v2.3/classes/ActiveSupport/CoreExtensions/Array/ExtractOptions.html

For the options, it would be nice if an `ArgumentError` was raised if an option was given that wasn't in `VALID_OPTIONS_FOR_ENUMS`

cheerfulstoic:
This looks great!  Just had one comment in the code to look at.  Are you still working, or would you like us to review for merge?

You didn't check the "Making sure that enums are supported in where and rel_where".  Is it working aside from the `Array` caveat that you mentioned?

I think you could throw in the validation pretty easily by putting this is in `define_property`:

```ruby
validates property_name, inclusion: { in: enum_keys }
```

lib/neo4j/shared/enum.rb
ProGM: Uhm, in ActiveRecord enum is using this convention:

Parameters are all key-values.

Keys that starts with `_` (`_prefix`, `_suffix`, ecc...) are options, the other are the name of the enum properties.

Check here:
http://edgeapi.rubyonrails.org/classes/ActiveRecord/Enum.html

I just tried to replicate it...

If you prefer, I can change this:

```ruby
enum key: [:list, :of, :keys], _option: true
```
to this:

```ruby
enum :key, [:list, :of, :keys], option: true
```

ProGM:
I'm still working of it, but I need a review for the `EnumConverter`, and your opinion on the code in general. ;)

Moreover, I don't know how to implement `where` for `Array` parameters.

> You didn't check the "Making sure that enums are supported in where and rel_where". Is it working aside from the Array caveat that you mentioned?

For now I tested only for `where` and it works. I'll add a test for `rel_where`, too.


> I think you could throw in the validation pretty easily by putting this is in define_property:

Uhm, in ActiveRecord Enums, wrong values are just ignored. (I think it falls back to the default one, or also `nil`). If I add a `validates`, the behavior will be quite different. Maybe we should add an option? (or make the user decide it?)

lib/neo4j/shared/enum.rb
cheerfulstoic: Ooooh, I see what it's doing now.  This is so that you can specify multiple enums in the same definition, right?

I do like the second syntax a lot better, though...  @subvertallchris are you fine with breaking convention?  It would make the code a lot simpler, and I even think that the API would be simpler for people to use (even if it would be different than what people used to ActiveRecord would expect)

cheerfulstoic:
I don't have a min to look at why the array thing isn't working, but we should leave open an issue for it.

For the enums validation, I think that sounds familiar now, from AR.  I'm happy to leave it the same as ActiveRecord and people can easily add the validation if they want.

ProGM:
I just noticed that `ActiveRel` has:

* No `index: :exact` for properties
* No `scope`s

So I have to create a more specific implementation for `ActiveRel`s

ProGM:
Ok, @subvertallchris @cheerfulstoic 
I'm about to end with this. I removed the terrible `method_missing`, replacing with a nicer `define_singleton_method`.

Last things: documentation (Oh, I hate this) and the change of syntax to:

```ruby
enum :key, [:list, :of, :keys], option: true
```

Please confirm me the last one! ;)

lib/neo4j/shared/enum.rb
subvertallchris: Sorry for not commenting earlier, didn't see I was tagged here. I don't really have a strong opinion either way. Is there any chance that breaking with convention would harm compatibility with gems that rely on ActiveRecord's implementation?

cheerfulstoic:
I hate to say it, but let's stick with `ActiveRecord`s implementation of `enum`.  It would be good to be compatible with things like `rails_admin`

ProGM:
@cheerfulstoic @subvertallchris Great!

I think I completed all the tasks.

I added some documentation, please take a look how it looks.

cheerfulstoic:
LGTM, thanks!  @subvertallchris any objections to a merge?

subvertallchris:
Looks great to me!

ProGM:
Now I think about it, should I add some infos in the `CHANGELOG`?

cheerfulstoic:
A line in the `CHANGELOG` would be great (don't forget to thank yourself!).  If you don't I'll throw it in in the morning (I'm headed to bed now).

We've been short on childcare the past couple of weeks and so I've had hardly any time to work on my refactor for 7.0.  My son should be in school the next month or so, so I hope to have more time to get that finished up and then this will be released at the same time

Thanks again!

ProGM:
Done

I see, if you need help with it, please ask ;)

cheerfulstoic:
Excellent, merged!  Thanks again ProGM / #1129!

PR #1134, opened 2016-02-13
fix #serialize method

subvertallchris:
Improvements to `serialize` class method. Workaround and possibly the path towards a solution for #1128?

### DIFF
--- a/spec/unit/shared/property_spec.rb
+++ b/spec/unit/shared/property_spec.rb
@@ -181,7 +181,7 @@ describe Neo4j::Shared::Property do
     before(:each) do
       # This serializer adds a text when the data is saved on the db,
       # and removes it when deserializing
-      class MySerializer
+      stub_const('MySerializer', Class.new do
         def initialize(text)
           @text = text
         end
@@ -205,7 +205,7 @@ describe Neo4j::Shared::Property do
         def to_db(value)
           "#{value}#{@text}" if value
         end
-      end
+      end)
 
       stub_active_node_class('MyData') do
         property :polite_string
--- a/spec/unit/shared/property_spec.rb
+++ b/spec/unit/shared/property_spec.rb
@@ -176,4 +176,52 @@ describe Neo4j::Shared::Property do
       expect(instance.class.declared_properties.convert_properties_to(instance, :ruby, instance.props)[:range]).to eq(range)
     end
   end
+
+  describe Neo4j::ActiveNode do
+    before(:each) do
+      # This serializer adds a text when the data is saved on the db,
+      # and removes it when deserializing
+      class MySerializer
+        def initialize(text)
+          @text = text
+        end
+
+        def converted?(value)
+          value.is_a?(db_type)
+        end
+
+        def db_type
+          String
+        end
+
+        def convert_type
+          Symbol
+        end
+
+        def to_ruby(value)
+          value.gsub(@text, '').to_sym if value
+        end
+
+        def to_db(value)
+          "#{value}#{@text}" if value
+        end
+      end
+
+      stub_active_node_class('MyData') do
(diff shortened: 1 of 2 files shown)


spec/unit/shared/property_spec.rb
```diff
@@ -176,4 +176,52 @@ def to_ruby(value)
       expect(instance.class.declared_properties.convert_properties_to(instance, :ruby, instance.props)[:range]).to eq(range)
     end
   end
+
+  describe Neo4j::ActiveNode do
+    before(:each) do
+      # This serializer adds a text when the data is saved on the db,
+      # and removes it when deserializing
+      class MySerializer
```
cheerfulstoic: I would use `stub_named_class`.  When we define classes directly in the specs we end up with weird issues.

cheerfulstoic: That's a helper in `spec_helpers.rb`, BTW

ProGM:
I'm still confused on how a serializer/typecaster should be used.

The main problem is here:

https://github.com/neo4jrb/neo4j/blob/serializer-fix/lib/neo4j/shared/type_converters.rb#L311

Since `type` can be both a type or a typecaster (and it's not explicit at all).

The second problem is that, when you set `serializer: MySerializer`

https://github.com/neo4jrb/neo4j/blob/77190ab416d219f230dc86eb71e3b7b92df65b0b/lib/neo4j/shared/declared_property.rb#L110

This is registering THAT `MySerializer ` as a global typeconverter for the type defined in `convert_type`.

Example:
```ruby
class MySerializer1
  class <<self
    //... some code here ...
    def convert_type
      String
   end
  end
end

class MySerializer2
  class <<self
    //... some code here ...
    def convert_type
      String
   end
  end
end
```

```ruby
property :a, serializer: MySerializer1
property :b, serializer: MySerializer2
property :c, type: String
```

This situation causes a strange behavior, since in `CONVERTERS[String]` will contain `MySerializer2` as default type converter for `String` type:
https://github.com/neo4jrb/neo4j/blob/77190ab416d219f230dc86eb71e3b7b92df65b0b/lib/neo4j/shared/declared_property.rb#L110


(Sorry, I'm not completely sure if what I'm saying is correct/clear, since all the process is a bit unclear to me too.)

subvertallchris:
You're totally clear and you're right, the `serializer` option is weird and needs attention. This PR isn't concerned with that, we'll need to address it separately. Here, we modify the `serialize` class method to provide the same behavior that you're trying to get from `serializer`. You can see it [here](https://github.com/neo4jrb/neo4j/pull/1134/files#diff-52ba3a2fe20b88cfc7c119ebde870c36R215).

The property option came from someone who was using the gem and contributing occasionally. It solved a problem he was having and seemed reasonable, so we merged it... and, from the looks of it, it hasn't much use since then, so it slipped between the cracks. If this PR does the trick for you, we can use it as the basis of a fix for that option; if not, that's ok, we can work on a different solution.

ProGM:
Oooh ok, now I understand.
I did a mess, confusing `serialize` method and `serializer` option.

Perfect. I think that this does the trick and could be a good solution for my issue ;)

subvertallchris:
Fixing Brian's note and merging. @cheerfulstoic, any reason you can think of not to cherrypick this into 6.1.x, too? Feels like a bugfix to me.

cheerfulstoic:
Yeah, backport by all means!

subvertallchris:
Cool. I'll take care of that tomorrow and do the release. As for the `serializer` property option, that will be a 7.0 thing that doesn't get backported. I'll take care of it separately. It might coincide with some refactoring of the type converters and I'm going to need the freedom to rename methods.

PR #1135, opened 2016-02-16
Active rel subclass fix

### LINKED ISSUE #1133: ActiveRel creation fails if a related node is subclassed
I have a [failing test](https://github.com/brucek/neo4j/blob/a2b80841204667f80f206755ea021ed1d725a853/spec/e2e/active_rel/persistence/query_factory_spec.rb) I'm trying to puzzle out now...

Basically this fails:

    class ParentClass
      include Neo4j::ActiveNode
    end
    
    class FromClass < ParentClass
      has_many :out, :to_classes, rel_class: 'RelClass'
    end
    
    class ToClass
      include Neo4j::ActiveNode
    end
    
    class RelClass
      type 'is_a'
      from_class :FromClass
      to_class :ToClass
    end

with this error:

    Neo4j::Session::CypherError: ParentTestClass_uuid not defined
      "MATCH (to_node) WHERE (ID(to_node) = {to_node_id})
      CREATE (from_node:`FromClass:ParentClass` {uuid: {from_node_FromClass:ParentClass_uuid}})
      CREATE UNIQUE from_node-[rel:is_a {rel_create_props}]->to_node
      SET rel.`created_at` = {setter_rel_created_at}
      RETURN rel, from_node"

It looks like it's a problem with that parameter needing to be quoted maybe to hide the identifier colon(:) from the parser?  But I'm still digging out where exactly the issue originates.  I wonder if this is related to #1128 also?

brucek:
Fixes #1133

### DIFF
--- a/lib/neo4j/active_rel/persistence/query_factory.rb
+++ b/lib/neo4j/active_rel/persistence/query_factory.rb
@@ -51,7 +51,7 @@ module Neo4j::ActiveRel::Persistence
       end
     end
 
-    # Each node must be either created or matched before the relationshp can be created. This class does not know or care about
+    # Each node must be either created or matched before the relationship can be created. This class does not know or care about
     #   how that happens, it just knows that it needs a usable Neo4j::Core::Query object to do that.
     # This method is "iterative" because it creates one factory for each node but the second builds upon the first.
     def iterative_query
--- a/lib/neo4j/shared/query_factory.rb
+++ b/lib/neo4j/shared/query_factory.rb
@@ -69,7 +69,7 @@ module Neo4j::Shared
 
     def create_query
       return match_query if graph_object.persisted?
-      base_query.create(identifier => {graph_object.labels_for_create.join(':').to_sym => graph_object.props_for_create})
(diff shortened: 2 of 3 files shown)


subvertallchris:
I'm a little confused, were changes merged into Neo4j-core to allow this or does just take advantage of existing behavior?

brucek:
Yeah, sorry, I'm working my way up the learning curve here...  This fix should take advantage of existing behavior.  Basically passing the labels in as an array already worked in neo4j-core, so the referenced subclass bug is fixed with this.

There was also a bug in neo4j-core when passing in parameter names with colon characters in them - my neo4j-core pull request is fixing that problem (but which is no longer triggered once this fix is merged here).

I also left in some explicit tests in that neo4j-core pull request for the existing "array of labels" behavior this fix uses that you guys may or may not want to keep.  My thought was to be explicit as well as doc the behavior in tests for future contributors.

subvertallchris:
This is great. Gonna merge, add a test, cherry-pick it back to 6.1.x, update changelog, and release a patch.

cheerfulstoic:
Just merged the PR which should help to support this

PR #1146, opened 2016-03-02
Add specs for property indexing

sineed:


cheerfulstoic:
I missed this one.  I think @subvertallchris is closer to the issue.  Does it look good to you?

subvertallchris:
Yes, but I'm not sure it belongs in this file. Let's go with it, more important to have the tests in there than delay more while figuring out where to move it.

cheerfulstoic:
Awesome, thanks ;)

Issue #1143, opened 2016-02-29
Scopes context doesn't include associations


ProGM:
I was trying to define a scope that selects all `Article`s that has a relation with `Version`.

Article and Version are defined like this:

```ruby
class Article
   ...
  has_one :out, :published_version, type: :published_version, model_class: :Version, unique: true
end

class Version
   ...
  has_one :in, :article, type: :published_version, model_class: :Version, unique: true
end
```

I built my query like this:

```ruby
Article.as(:n).published_version(:v).query.proxy_as(Article, :n)
```

And it works quite good.

So I tried to put this query in a scope:

```ruby
scope :published, -> { published_version(:v).query.proxy_as(Article, identity) }
```

But when I call `Article.published`, I receive:

```
NoMethodError: undefined method `published_version' for #<Neo4j::ActiveNode::Scope::ScopeEvalContext:0x007fb66aac2a38>
```

It's a bit annoying that Scopes can't see relations, since in ActiveRecord I could to the same by doing:

```ruby
scope :published, -> { join(:published_version) }
```

Is it a bug? Or I'm doing it wrong?

cheerfulstoic:
Interesting.... does this work?

```ruby
scope :published, -> { all.published_version(:v).query.proxy_as(Article, identity) }
```

If not, this should definitely work:

```ruby
  def self.published
    all.published_version(:v).query.proxy_as(Article, identity)
  end
```

Probably scopes should allow for this, though.  And we should have a better way to reach out through an association to do a conditional (or even just a `MATCH` like you're doing) and then return the original scope context.  Maybe something like:

```ruby
Article.branch { |articles| articles.published_version(:v) } # Returns a scope re-wound to the "articles" point
```

ProGM:
Uhm, I see. This is working:

```ruby
  def self.published
    all.published_version(:v).query.proxy_as(Article, identity)
  end
```
But definitely it's not a scope (so I can't chain).

The other one isn't. :(

`branch` method would be very useful!

Do you mean something like:

```ruby
def branch
  if block_given?
    yield(self).query.proxy_as(self.class, identity)
  else
    fail ArgumentError, "Block required"
  end
end
```

cheerfulstoic:
It's not a scope, but it should work just like a scope.  Give it a try in a chain! ;)  Keep in mind, though, that the `all` is the secret to making it chain-able

[...]

PR #1147, opened 2016-03-02
`branch` method for QueryProxy

ProGM:
An implementation for what we discussed here #1143

This method allows to `match` a relation inside the current query proxy, without changing the current scope.

Example:

```ruby
Person.where('age > 3').car # Returns a list of cars
Person.where('age > 3').branch { car } # Returns a list of people that have a car
Person.where('age > 3').branch { car(:c).where("c.kilometers < 5000") } # Returns a list of people that have a new car
```

### DIFF
--- a/lib/neo4j/active_node/query/query_proxy.rb
+++ b/lib/neo4j/active_node/query/query_proxy.rb
@@ -174,6 +174,25 @@ module Neo4j
           self
         end
 
+        # Executes the relation chain specified in the block, while keeping the current scope
+        #
+        # @example Load all people that have friends
+        #   Person.all.branch { friends }.to_a # => Returns a list of `Person`
+        #
+        # @example Load all people that has old friends
+        #   Person.all.branch { friends.where('age > 70') }.to_a # => Returns a list of `Person`
+        #
+        # @yield the block that will be evaluated starting from the current scope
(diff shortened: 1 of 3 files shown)


cheerfulstoic:
Yay, thank you!

Would like opinions from @subvertallchris and @leviwilson

ProGM:
Ach, I'm an idiot! Tests are failing since I moved things around before committing. I'm fixing ;)

subvertallchris:
This is fantastic!

leviwilson:
I dig it! Like the syntax.

ProGM:
Finally travis completed :zzz:

Tell me what do you think ;)

cheerfulstoic:
I like the `instance_eval` addition so that you don't need to use a passed in argument.

I would just like to see one more spec before merge.  It would be good to see that if we do this branching that we get the right result.  Say that we have two students and one lesson and only one of the students has a lesson.  Then we could do something like `expect(Student.all.branch { lessons }.count.to eq(1)`

ProGM:
Done :+1:

cheerfulstoic:
Great work again, thanks so much!

cheerfulstoic:
Added `CHANGELOG` entry

PR #1150, opened 2016-03-09
Add the ability to specify an array for enum properties in a where clauses

cheerfulstoic:


spec/e2e/enum_spec.rb
```diff
@@ -70,6 +70,16 @@
       expect(ids).to include(file2.id)
     end
 
+    it 'supports arrays' do
```
ProGM: Why did you define a new test instead of using the existing (skipped) one? Is it intentional?

ProGM:
I like this!

I just left a comment on tests ;)

spec/e2e/enum_spec.rb
cheerfulstoic: As I understand the rspec way, things should be separated into examples as much as possible.  Obviously you need to balance that if there's a lot of setup.

cheerfulstoic:
Thanks! ;)  @subvertallchris ?

spec/e2e/enum_spec.rb
ProGM: Uhm... I'm not sure I was clear...
When I implemented enums, I defined a test for array parameters, that is veeeery similar to this one. Since `where` with array parameters was not yet implemented, I left it with a `skip` message, so when someone else would implement it, he could use it.
I mean this:
https://github.com/neo4jrb/neo4j/blob/enum_array_values/spec/e2e/enum_spec.rb#L83

cheerfulstoic: Ah, I see now ;)  That spec right there next to the spec that I wrote! ;)

I removed my spec and used yours, though I added the third file again so that it could test that querying with an array will return multiple objects of different `type`s

subvertallchris:
Looks great!

cheerfulstoic:
Awesome, thanks!  Released as 7.0.0.rc.3!

PR #1153, opened 2016-03-11
Rescuing `Neo4j::ActiveNode::Labels::RecordNotFound` with a 404 error page

ProGM:
I noticed that `ActionDispatch` `rescue_responses` doesn't work with inheritance.

So, even if `Neo4j::ActiveNode::Labels::RecordNotFound` is a child of `Neo4j::RecordNotFound`, and `Neo4j::RecordNotFound` is rescued with a 404 error page, Rails does not rescues it, too.

I had to add it manually to the list.

#1104 

I think you should add this to 7.0 release!

### DIFF
--- a/lib/neo4j/railtie.rb
+++ b/lib/neo4j/railtie.rb
@@ -17,11 +17,13 @@ module Neo4j
     # For rails 3.2 and 4.0
     if config.action_dispatch.respond_to?(:rescue_responses)
       config.action_dispatch.rescue_responses.merge!(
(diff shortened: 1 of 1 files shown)


cheerfulstoic:
Great, thanks!  Released as 7.0.0.rc.4!

Joshfindit:
Just a quick note for any other stumblers:

Alternately use `Object.exist?(id: <id>)` to get a true/false answer

Issue #1126, opened 2016-02-04
`with_associations` and Kaminari-neo4j not working together


ProGM:
I'm having issues when running `with_associations` and `kaminari` together.
I have tried this query: 
```
Article.all.with_associations(:creator).page(1).to_a
```
The `page` method comes from `kaminari-neo4j` gem.

I'm getting this error:

```bash
Neo4j::Session::CypherError: Invalid input 'K': expected 't/T' or 'e/E' (line 1, column 84 (offset: 83))
"MATCH (n:`Article`) OPTIONAL MATCH n-[:`creator`]->creator WHERE (creator:Person) SKIP {skip_0} LIMIT {limit_25} RETURN n, [collect(creator)]"
```

The generated query:

```cypher
MATCH (n:`Article`) OPTIONAL MATCH n-[:`creator`]->creator WHERE (creator:Person) SKIP {skip_0} LIMIT {limit_25} RETURN n, [collect(creator)] | {:skip_0=>0, :limit_25=>25}
```
I'm not sure if it's a `with_association` bug or something related with kaminari
It seems like `with_associations` is pulling the `return` statement in the bottom of the query.

Any ideas?

@cheerfulstoic

cheerfulstoic:
Looks like there's a `break` in the code for `with_associations`.  I know I put it in there for a reason so I need to track that down ;)

Heading out to lunch now, but I'll look again shortly.  Thanks for the report!

cheerfulstoic:
Ok, so this is actually kind of tough... With `with_associations` the `OPTIONAL MATCH`es aren't added to the chain until you do the enumeration (the `to_a` in your case).  I did this to copy ActiveRecord, but I'm thinking maybe it's a bad idea and that we should build the `OPTIONAL MATCH` parts at the time you call `with_associations`.

@subvertallchris do you see a problem with that?  Your feedback is welcome too @ProGM  ;)

ProGM:
I know I already said on the previous issue with `with_associations`, but I think that adding a `with` statement before any call of `optional match` could be a solution.

Example:
```
MATCH (n:`Article`) WITH n SKIP {skip_0} LIMIT {limit_25} OPTIONAL MATCH n-[:`creator`]->creator WHERE (creator:Person)  RETURN n, [collect(creator)] | {:skip_0=>0, :limit_25=>25}

```
But, as you said, I'm not sure if this has some implication on performances....

ProGM:
[...]

Issue #1048, opened 2015-11-18
Trouble finding objects with `with_associations`


cheerfulstoic:
Trying this:

```ruby
    @person1 = Person.all.with_associations(:authored_episodes, :mentioned_in).where(name: params[:name1]).to_a.first
```

But the `WHERE` is after the `OPTIONAL MATCH` and thus doesn't work

Tried switching to:

```ruby
    @person1 = Person.where(name: params[:name1]).with_associations(:authored_episodes, :mentioned_in).to_a.first
```

But that doesn't work either.

Tried this:

```ruby
    @person1 = Person.where(name: params[:name1]).query_as(:person).proxy_as(Person, :person).with_associations(:authored_episodes, :mentioned_in).to_a.first
```

Which kind of works, except that since there aren't any matches for `authored_episodes` then the `OPTIONAL MATCH`es don't match anything.  ARGH!

cheerfulstoic:
I think #1054 might alleviate this some

ProGM:
Uhm I'm having a similar issues with eager loading.

I'm trying to load a desk with all it's authors:

```ruby
Desk.all.where(title: 'Desk Title 2').with_associations(:creator).to_a
```

Generated query:
```
MATCH (n:`Desk`) OPTIONAL MATCH n-[:`creator`]->creator WHERE (n.title = {n_title}) AND (creator:Person) RETURN n, [collect(creator)] | {:n_title=>"Desk Title 2"}
```

As a result, the `where` statement is completely ignored.

I have a proposal: what about adding a `with` statement to separate `OPTIONAL MATCH` and the `WHERE`?

So the generated query should be:

```
MATCH (n:`Desk`) WHERE (n.title = {n_title}) WITH (n)  OPTIONAL MATCH n-[:`creator`]->creator WHERE (creator:Person) RETURN n, [collect(creator)] | {:n_title=>"Desk Title 2"}
```

cheerfulstoic:
I think that `7.0.0` fixes this. Let me know if not!

PR #1154, opened 2016-03-15
Fixing `with_associations`

cheerfulstoic:
This should hopefully fix the issues with with_associations and kaminari as well as getting duplicates when using multiple associations with multiple results with at least one association

The big problem is that if you do multiple `OPTIONAL MATCH` clauses in a row you end up getting a cartesian product and your result will have duplicates.

This should also fix issues where previous query proxies are getting mixed up with the clauses from `with_associations` (like with `kaminari-neo4j`).

See #1126 #1048 and #1126 996

### DIFF
--- a/lib/neo4j/active_node/query/query_proxy_eager_loading.rb
+++ b/lib/neo4j/active_node/query/query_proxy_eager_loading.rb
@@ -5,7 +5,7 @@ module Neo4j
         def each(node = true, rel = nil, &block)
           return super if with_associations_spec.size.zero?
 
-          query_from_association_spec.pluck(identity, with_associations_return_clause).map do |record, eager_data|
+          query_from_association_spec.pluck(identity, "[#{with_associations_return_clause}]").map do |record, eager_data|
             eager_data.each_with_index do |eager_records, index|
               record.association_proxy(with_associations_spec[index]).cache_result(eager_records)
             end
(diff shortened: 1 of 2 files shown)


andrewhavens:
@cheerfulstoic I just tried 7.0.0.rc.7 and it did fix my cypher query error that I was getting in version 6.x. However, when using `order` and `with_associations`, it did not order the results as expected:

```ruby
post.comments.with_associations(:author).order(created_at: :asc).to_a`
```
I also tried changing the order of the query to see if that made a difference, but it didn't.

```ruby
post.comments.order(created_at: :asc).with_associations(:author).to_a`
```

cheerfulstoic:
Can you share the Cypher that in runs?  When I try something similar I get:

```cypher
MATCH person584
WHERE (ID(person584) = {ID_person584})
MATCH person584-[rel1:`films`]->(result_films:`Film`)
WITH result_films
ORDER BY result_films.title ASC
OPTIONAL MATCH result_films<-[:`films`]-characters WHERE (characters:`Person`)
WITH result_films, collect(characters) AS characters_collection
RETURN result_films, [characters_collection]
```

So the first `WITH` gets an `ORDER BY`, but then the results don't seem to come out ordered, though that is what I would expect, I think.

cheerfulstoic:
Already merged in and released in `7.0.0`

PR #1161, opened 2016-03-21
Raise an error if two associations with the same name are defined

### LINKED ISSUE #1139: Raise an error if two associations are defined with the same name
The only reason that I can think that this might not make sense is if a subclass defines an association to overwrite it's superclass' association.  Maybe we can support that but raise an error if two associations are defined on the exact same model class

cheerfulstoic:
Fixes #1139

This pull introduces/changes:
 * Exception when two associations with the same name are defined
 * Refactored migration specs to use `stub_active_node_spec` (Could you double-check the changed spec @subvertallchris to make sure the spec is still valid?)

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -38,7 +38,7 @@ Metrics/MethodLength:
 # Offense count: 3
 # Configuration parameters: CountComments.
 Metrics/ModuleLength:
-  Max: 162
+  Max: 165
 
 # Offense count: 35
 # Configuration parameters: EnforcedStyle, SupportedStyles.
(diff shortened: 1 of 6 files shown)


subvertallchris:
Looks good. Those migration specs were always a bit of a disaster, so every little improvement there is nice.

What happens if someone subclasses and wants to redefine an association? Is that something that could actually happen?

cheerfulstoic:
Great point!  I just pushed a spec to check that.  I think it will be possible, but I'll need to look at it later

cheerfulstoic:
Have a look.  I changed it to calculate the associations when requested rather than when inherited.  This way the `@associations` variable on the class only holds those associations which have been defined for the class.  It comes at the cost of doing some extra merging, but I don't think that it will happen much.

PR #1164, opened 2016-03-21
Implemented find_or_initialize_by and first_or_initialize

### LINKED ISSUE #1157: Was `find_or_initialize_by` method removed?
It was added here:
https://github.com/neo4jrb/neo4j/pull/33

But now `Model.find_or_initialize_by` returns a `NoMethodError`. Is it intentional?

ProGM:
Fixes #1157 

This pull introduces/changes:
 * `find_or_initialize_by` as class method for every model, like `Student. find_or_initialize_by`
 * `find_or_initialize_by` method for query proxy, like `Lesson.first.students.find_or_initialize_by(name: "jimmy")`
 * `first_or_initialize` method  for query proxy, like `Lesson.first.students.where(name: "jimmy").first_or_initialize`

It took a bit

This is not complete yet, because I'd like a revision from you. Here's a small briefing:

* The first method is pretty straight forward. simply does a `find_by`, then a `new`

* The second one has the problem to identify the inverse association given one. For example:
When you do `Lesson.first.students.find_or_initialize_by(name: "jimmy")`, you expect to obtain an student  "jimmy", with a relation with that lesson. So, in order to initialize it, I had to find the reverse relation that, from a Student links to a Lesson, then assign it.
You can find the code here: 
https://github.com/ProGM/neo4j/commit/8672051fb64cf3d67689d9db885c63dd28853aca#diff-7e32b136ebe6054450b7f3d67ceccbe1R179
(It's a very test-oriented implementation: It works mainly for the test I wrote)

* The third one had one more problem. Given a QueryProxy with some `where` chain, I have to extract all parameters from it. For example, if I do `Students.where(name: 'jon').where(surname: 'snow')`, I have to initialize a new student with `{ name: 'jon', surname: 'snow' }`
https://github.com/ProGM/neo4j/commit/8672051fb64cf3d67689d9db885c63dd28853aca#diff-7e32b136ebe6054450b7f3d67ceccbe1R194

Since I'm not an expert, I wrote some methods to do this, but I'm sure we can do it better.

Any suggestion would be very helpful!

Pings:
@cheerfulstoic
@subvertallchris


cheerfulstoic:
Does it seems like it would be valuable do to this?

```ruby
philosophy.students.where(name: 'Rebecca').first_or_initialize
```

If not, it might make sense for the logic to exist on `AssociationProxy` instead of `QueryProxy`.

I see `first_or_initialize` in this post: http://www.mitchcrowe.com/10-most-underused-activerecord-relation-methods/

So `find_or_initialize` probably works in ActiveRecord, though I'm not able to find an example with a cursory Googling.

Also, I wonder does our existing `find_or_create_by` work in the same way?  I suspect it probably should, but there don't seem to be any specs for `find_or_create_by`.....

ProGM:
`find_or_create_by` works, but the implementation is very simple, since it has to create the relation, not to initialize it. Check here for more details:

https://github.com/ProGM/neo4j/blob/find_or_initialize_by/lib/neo4j/active_node/query/query_proxy_methods.rb#L136

I added this two features to QueryProxy, since also `find_or_create_by` was there. I didn't think about it too much, to be honest.

ProGM:
@cheerfulstoic @subvertallchris 
I was a bit busy lately. I worked on it a bit, today. For now it works, but it can be largely improved, I think.

Any suggestions?

spec/e2e/query_proxy_methods_spec.rb
```diff
@@ -72,6 +72,110 @@ def destroy_called
     end
   end
 
+  describe 'find_or_initialize_by' do
+    context 'with relations chain' do
+      let(:emily)       { Student.create(name: 'Emily') }
+      let(:philosophy)  { Lesson.create(name: 'philosophy') }
+      before do
+        philosophy.students << jimmy
```
cheerfulstoic: Optionally you could also do

```ruby
Lesson.create(name: 'philosophy', students: [jimmy])
```

In the `philosophy` block.  Don't know if that would break any tests because this `before` is currently making sure that these `let` variables are referenced before any of the `it` blocks are run

spec/e2e/query_proxy_methods_spec.rb
```diff
@@ -72,6 +72,110 @@ def destroy_called
     end
   end
 
+  describe 'find_or_initialize_by' do
+    context 'with relations chain' do
```
cheerfulstoic: I would say that this should be a `describe`, not `context`.  `context` generally talks about the `let`/`before` things that you're setting up, where as here you`re further specifying that you're describing using `find_or_initialize_by` with relations chains

spec/e2e/query_proxy_methods_spec.rb
```diff
@@ -72,6 +72,110 @@ def destroy_called
...
+        expect(philosophy.students.where(name: 'Rebecca').first).to be_nil
+        student = philosophy.students.find_or_initialize_by(name: 'Rebecca')
+        expect(student).not_to be_persisted
+        expect(student.name).to eq('Rebecca')
+        expect(student.lessons.to_a).to include(philosophy)
+        expect { student.save }.to change { philosophy.reload.students.count }.by(1)
+      end
+
+      it 'returns the node after initializing' do
+        expect(philosophy.students.find_or_initialize_by(name: 'Jacob')).to be_a(Neo4j::ActiveNode)
```
cheerfulstoic: Since all of these are doing a `find_or_initialize` by on `philosophy.students` perhaps you could DRY it up a bit like:

```ruby
describe 'find_or_initialize_by on an association' do
  subject { philosophy.students.find_or_initialize_by(name: 'Jacob') }

  it { should be_a(Neo4j::ActiveNode) }
  its(:name) { should eq('Jacob') }
  it { should_not be_persisted }
  its(:lessons) { should include(philosophy) }
  it 'can be saved' do
    expect { subject.save }.to change { philosophy.reload.students.count }.by(1)
  end
end
```

lib/neo4j/active_node/persistence.rb
```diff
@@ -159,6 +159,14 @@ def find_or_create_by!(attributes, &block)
         find_by(attributes) || create!(attributes, &block)
       end
 
+      def find_or_initialize_by(attributes, &block)
+        if block
+          find_by(attributes) || new(attributes).tap(&block)
+        else
+          find_by(attributes) || new(attributes)
+        end
```
cheerfulstoic: Perhaps

```ruby
def find_or_initialize_by(attributes, &block)
  find_by(attributes) || (block ? new(attributes).tap(&:block) : new(attributes))
end
```

Or:

```ruby
def find_or_initialize_by(attributes, &block)
  find_by(attributes) || new(:attributes).tap {|o| block.call(o) if block }
end
```

Or simply:

```ruby
def find_or_initialize_by(attributes, &block)
  find_by(attributes) || new(:attributes).tap(&block || Proc.new)
end
```

cheerfulstoic:
Seems good overall, thanks!  I agree that the code looks like it could be smoothed some, but I don't see any immediate big wins.

I left some comments for you to consider. Some of them might require a decent amount of refactoring which I'm not too concerned with if you aren't.  If you're happy with everything I'll merge

spec/e2e/query_proxy_methods_spec.rb
ProGM: Definitely like this :+1:

spec/e2e/query_proxy_methods_spec.rb
ProGM: Uhm, I see, I'll fix

spec/e2e/query_proxy_methods_spec.rb
ProGM: Yep, I need it for that reason...

lib/neo4j/active_node/persistence.rb
ProGM: Uhm, good point

ProGM:
Ok, I'll take a look to it tomorrow :)

My biggest concern about my code was about this part:
https://github.com/neo4jrb/neo4j/pull/1164/files#diff-7e32b136ebe6054450b7f3d67ceccbe1R204

It's a tricky way to extract current `where()` hash parameters, and I'm not sure it would work in every case.

cheerfulstoic:
Great, thanks!  I see what you mean by the tricky code.  I'm happy to deal with it as problems come up ;)

ProGM:
@cheerfulstoic @subvertallchris I think I completed it. Please take a look and merge, if you like it 👍

cheerfulstoic:
Seems great, thanks so much!

Issue #1100, opened 2016-01-10
ActiveNode/ActiveRel helpers for moving relationships


cheerfulstoic:
Neo4j doesn't let you simply change the start/end nodes of a relationship.  To do this you need to delete the old one and create a new one (propagating any data stored on the relationship).  We should be able to make this easy from the Active(Node|Rel) APIs

pdipietro:
Well, this is exactly what I need today ;-)

When do you think to implement and release it?

TIA

cheerfulstoic:
Good question ;)  Right now @subvertallchris and I have been focusing on a couple of big refactors of the gems for the next major release.  I would like to see this in the near future, though.

If you have time it might not even be that hard since it would be all new functionality.  Perhaps you could propose an API and we could discuss it

pdipietro:
My dime:

IMHO the API must adhere to the KISS principle, then, starting with
```
rel = EnrolledIn.new(from_node: student, to_node: lesson)
#or
rel = EnrolledIn.create(from_node: student, to_node: lesson)
```

a solution could be 
```
rel.redirect!(:start, new_start_node)
rel.redirect!(:end, new_end_node)
```

an alternate better name could be `reorient`

and a less misleading solution coul then be

```
rel.reorient_start_to!(new_start_node)
rel.reorient_end_to!(new_end_node)
```

Maybe end node could be also a collection, so letting the user to easily duplicate the relationships.

A good solution should be able to solve the most general problem of reorienting all the 'same kind' relationship belonging to the Current node to a new node

Imagine a property transfer: I own several different things but I want transfer all my ownerships to another person or persons or (persons and pets).

CurrentNode.[rel:RelationKind].reorient(:to|:from).aNewNode

cheerfulstoic:
I like that, though I might propose `rewire` as an alternative.  Though I might still like `redirect` better.

Keeping it super simple, we might just add a `start_node=` method to compliment the `start_node` method.  What do you think?

As far as making duplicates, I thing the [SRP](https://en.wikipedia.org/wiki/Single_responsibility_principle) would say that it should be a separate method.  Perhaps you can call `clone` / `dup` on an `ActiveRel` object in a similar manner to `ActiveRecord` (see [#clone](http://apidock.com/rails/ActiveRecord/Core/clone) / [#dup](http://apidock.com/rails/ActiveRecord/Core/dup)).  That makes me realize that we should probably copy `ActiveRecord` in this respect for `ActiveNode` as well.

PR #1166, opened 2016-03-22
Add migration for mass relabeling relationship nodes

JustinAiken:
> So we're in the midst of updating the neo4j gem from 3.x to 6.x... One of the pain points was the way the relationship labels changed - `#old_way` to `NEW_WAY` (which is a change that makes sense and I'm glad you did!) Anyways, I'd written a helper that went through all the existing relationships and renamed them... [was wondering the best way to contribute it upstream]

- - me

> Hey @JustinAiken! That sounds like it might be a good fit for https://github.com/neo4jrb/neo4j/blob/master/lib/neo4j/migration.rb ?
You just need a class inheriting from `Neo4j::Migration` and the three methods in the public interface

- - @subvertallchris

Here's my first rough whack at fitting into a subclass of that... if this doesn't fit the project well I'll close and maybe create an outside gem, but if you like where it's heading I can clean it up, add more test cases, and do docs/changelog/etc

### DIFF
--- a/lib/neo4j/migration.rb
+++ b/lib/neo4j/migration.rb
@@ -126,5 +126,68 @@ MESSAGE
         end
       end
     end
+
+    class RelabelRelationships < Neo4j::Migration
+      attr_accessor :relationships_filename, :old_format, :new_format
+
+      def initialize(path = default_path)
+        @relationships_filename = File.join(joined_path(path), 'relabel_relationships.yml')
+      end
+
+      MESSAGE = <<MESSAGE
+# Provide relationships which should be relabled.
+# relationships: [students,lessons,teachers,exams]\nrelationships: []
+# Provide old and new label formats:
+# Allowed options are lower_hashtag, lower, or upper
+formats:\n  old: lower_hashtag\n  new: lower
+MESSAGE
+
+      def setup
+        super
+        return if File.file?(relationships_filename)
+        File.open(relationships_filename, 'w') { |f| f.write(MESSAGE) }
+      end
+
+      def migrate
+        config        = YAML.load_file(relationships_filename).to_hash
+        relationships = config['relationships']
+        @old_format   = config['formats']['old']
+        @new_format   = config['formats']['new']
+
+        output 'This task will relabel every given relationship.'
+        output 'It may take a significant amount of time, please be patient.'
+        relationships.each { |relationship| reindex relationship }
+      end
+
+      private
+
+      def count(relationship)
+        Neo4j::Session.query(
+          "MATCH (a)-[r:#{style(relationship, :old)}]->(b) RETURN COUNT(r)"
+        ).to_a[0]['COUNT(r)']
+      end
+
+      def reindex(relationship)
+        count = count(relationship)
+        output "Indexing #{count} #{style(relationship, :old)}s into #{style(relationship, :new)}..."
+        while count > 0
+          Neo4j::Session.query(
+            "MATCH (a)-[r:#{style(relationship, :old)}]->(b) CREATE (a)-[r2:#{style(relationship, :new)}]->(b) SET r2 = r WITH r LIMIT 1000 DELETE r"
+          )
+          count = count(relationship)
+          if count > 0
+            output "... #{count} #{style(relationship, :old)}'s left to go.."
+          end
+        end
+      end
+
+      def style(relationship, old_or_new)
+        case (old_or_new == :old ? old_format : new_format)
+        when 'lower_hashtag' then "`##{relationship.downcase}`"
+        when 'lower'         then "`#{relationship.downcase}`"
+        when 'upper'         then "`#{relationship.upcase}`"
+        end
+      end
+    end
   end
 end
--- a/lib/neo4j/migration.rb
+++ b/lib/neo4j/migration.rb
@@ -22,6 +22,10 @@ module Neo4j
       File.join(path.to_s, 'db', 'neo4j-migrate')
     end
 
+    def setup
+      FileUtils.mkdir_p('db/neo4j-migrate')
+    end
+
     class AddIdProperty < Neo4j::Migration
       attr_reader :models_filename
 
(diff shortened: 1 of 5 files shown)


spec/e2e/migration_spec.rb
```diff
@@ -13,7 +13,7 @@ class ThirdRelClass; end
       class User
         include Neo4j::ActiveNode
         property :name
-        has_many :out, :songs, model_class: :Song, type: 'songs'
+        has_many :out, :songs, model_class: :'MigrationSpecs::Song', type: 'songs'
```
JustinAiken: Before this change, doing `User.first.songs` gave me `ArgumentError: Could not find class: ::Song`

subvertallchris:
This looks cool. Will it also move properties and clean up old data? We've had a few discussions about adding a method or methods to do this, too. See https://github.com/neo4jrb/neo4j/issues/1100 for one discussion. Maybe there are some reusable pieces that'll help us move towards that?

@cheerfulstoic, thoughts?

JustinAiken:
> Will it also move properties

No... I'm not sure how to make it do that.  I added a test case to demonstrate that properties do get stripped.  

> and clean up old data?

Yes! Added a test case for that

> We've had a few discussions about adding a method or methods to do this, too. See #1100 for one discussion. Maybe there are some reusable pieces that'll help us move towards that?

Looks like a related issue, but I don't see any way to combine the per-relationship-node-instance API with the mass update... I'm pretty unfamiliar with the code though, so if you do see anything reusable, let me know and I'll move it over!

lib/neo4j/migration.rb
```diff
@@ -123,5 +126,68 @@ def new_id_for(model)
...
+        @relationships_filename = File.join(joined_path(path), 'relabel_relationships.yml')
+      end
+
+      MESSAGE = <<MESSAGE
+# Provide relationships which should be relabled.
+# relationships: [students,lessons,teachers,exams]\nrelationships: []
+# Provide old and new label formats:
+# Allowed options are lower_hashtag, lower, or upper
+formats:\n  old: lower_hashtag\n  new: lower
+MESSAGE
```
cheerfulstoic: Would be nice to see a full example of this `yml`.  Perhaps in the docs.  The docs are under the `docs` directory and they generate this site: http://neo4jrb.readthedocs.org/en/7.0.x/

cheerfulstoic: Oh, and you can use `rake docs` and `rake docs:open` to test the docs locally

lib/neo4j/migration.rb
```diff
@@ -123,5 +126,68 @@ def new_id_for(model)
...
+
+        output 'This task will relabel every given relationship.'
+        output 'It may take a significant amount of time, please be patient.'
+        relationships.each { |relationship| reindex relationship }
+      end
+
+      private
+
+      def count(relationship)
+        Neo4j::Session.query(
```
cheerfulstoic: Does this work?  I thought that you needed to do `Neo4j::Session.current.query`

spec/e2e/migration_spec.rb
cheerfulstoic: I actually did some refactoring of this just a couple of days ago so that the `MigrationSpecs` module isn't needed anymore.  I just merged [the PR](https://github.com/neo4jrb/neo4j/pull/1161) into master so you can try merging it into you branch if you like.  Otherwise we'll take care of it when this PR is done

cheerfulstoic:
It looks like your code does move properties (the `SET r2 = r`).  But maybe that's not what it's doing because you seem to have a spec testing that it does strip the properties.

cheerfulstoic:
It seems fine with me.  I'm fine with it just being a straight Cypher query instead of re-using other code since I see this as mainly a way to help people upgrade versions

spec/e2e/migration_spec.rb
JustinAiken: Ah, nice refactor... rebased this branch off it, will handle the docs tomorrow :smile:

cheerfulstoic: :metal:

lib/neo4j/migration.rb
JustinAiken: Just added the docs below... let me know if that's about what you had in mind!

lib/neo4j/migration.rb
JustinAiken: The other migration task (`AddIdProperty`) does just `Neo4j::Session.query`, so I kept that...
In my real-world use on 6.1.10 there was no `.current` either..

JustinAiken:
Screenshot of added doc page:

screen

lib/neo4j/migration.rb
cheerfulstoic: Fair enough ;)

cheerfulstoic:
LGTM!  @subvertallchris merge if it looks good to you?

subvertallchris:
I'll give it a close look again tomorrow!

subvertallchris:
It looks good but I have concerns about it not migrating properties. It's something I'd expect in a feature like this, especially since it's going to delete the old relationship afterwards. What about including properties to be migrated within the YML file?

subvertallchris:
Hey @JustinAiken, I'd love to get back into this if you're still interested. I'm happy to take a stab at adding the property migration if you want. What do you think?

JustinAiken:
Ah yes... I'm not sure what to do about migrating properties... All of our relationship nodes are just empty labels, so it never came up for us.  If you have an idea on how to rewrite the new relationships including the old properties, I'd love to hear it! 😄

subvertallchris:
Cool! I'll probably work off your branch and open a PR that covers it. Is
that ok?

JustinAiken:
> Cool! I'll probably work off your branch and open a PR that covers it. Is
that ok?

Of course! :)

ProGM:
Closing here :)

PR #1174, opened 2016-03-24
support for using neo_id as id_property

klobuczek:
Fixes #

This pull introduces/changes:
 * support for using neo_id as id_property (id_property :neo_id)
 * allowed usage of :id and primary_key in order clause and some case of pluck which do not call directly neo4jrb-core
 * generalized over 50 specs to be agnostic to id_property
* added 2nd rspec pass to travis with neo_id confuguration (to avoid duplication of specs, open for suggestions)

requires https://github.com/neo4jrb/neo4j-core/pull/254

Pings:
@cheerfulstoic
@subvertallchris


cheerfulstoic:
This looks good, thanks!  A few things:

 * I merged the `neo4j-core` PR and so I'm re-running Travis CI
 * There are a few rubocop failures.  I'd run `rubocop -aD` which should fix most/all of them
 * I think that `id_property` would be a better method than `primary_key`, though I think you're just using an existing method that we had.  There's `id_property_name` or `id_property_info`
 * I think that `object.id` would be better than using the `primary_key` helper method.  We might want to have some way to run the tests so that the `neo_id` is the default to make sure things continue to work
 * I'm thinking that things still might fail in a couple of places if somebody is using the `neo_id` as a primary key, but as long as `uuid` is working we can work those out over time

Thoughts @subvertallchris ?

klobuczek:
most of the test where I had to use the `primary_key` helper were testing both `object.id` and `object.uuid`. Should I simply remove the latter?

spec/e2e/query_spec.rb
```diff
@@ -86,9 +86,9 @@ def self.ordered_by_subject
       student.lessons << lesson
 
       expect(Student.as(:s).pluck(:s)).to eq([student])
-      expect(Student.all.pluck(:uuid)).to eq([student.uuid])
+      expect(Student.all.pluck(:id)).to eq([student.id])
```
klobuczek: Here `pluck(:id)` is working as `all` returns object of class implemented in neo4j.

spec/e2e/unpersisted_association_spec.rb
```diff
@@ -237,7 +237,7 @@ def save_and_expect_rel!
           chris = Student.create(name: 'Chris', favorite_class: math)
           expect(chris.errors).to be_empty
 
-          lessons = chris.query_as(:c).match('c-[:FAVORITE_CLASS]->(l:Lesson)').pluck('l.uuid')
+          lessons = chris.query_as(:c).match('c-[:FAVORITE_CLASS]->(l:Lesson)').pluck(l: Lesson.id_property_name)
```
klobuczek: Here `pluck(l: :id)` is not working as match returns directly object of neo4j-core so no way to intercept and not enough information in neo4j-core to implement. I think that all those chaining methods should return some proxy in neo4j rather neo4j-core. The same problem occurs with where clauses. This is not specific to neo_id.

klobuczek:
* rubocops issues fixed
* primary_key renamed
* added a test with neo_id as default config to build matrix
* there are still queries which are not compatible with addressing parameters by `:id` but it is a bigger problem and not neo_id specific, please see my inline comments

spec/e2e/unpersisted_association_spec.rb
cheerfulstoic: Yeah, agreed that using `Lesson.id_property_name` is the right approach here.  I'm not 100% sure why this is using the `Query` API from `neo4j-core`, but I'm guessing it's to verify in the database without depending on other functionality of `neo4j`

cheerfulstoic:
Awesome, thanks!

@subvertallchris Are you good with these changes?

I'd like to have another once over.  Recent days have been pretty busy.  I should be able to review deeper in a couple of days.

cheerfulstoic:
BTW @klobuczek I'm curious if this is working for you in your project as a replacement for the `active_node` gem

klobuczek:
@cheerfulstoic It will be still at least a week's effort to migrate our project to neo4j. The API is not identical. But I'm pretty sure that the id was the only prohibitive feature.

spec/e2e/id_property_spec.rb
```diff
@@ -64,7 +64,7 @@
           Neo4j::Config[:id_property] = :the_id
           Neo4j::Config[:id_property_type] = :auto
           Neo4j::Config[:id_property_type_value] = :uuid
-          stub_const('Clazz', Class.new do
+          stub_const('Clazz', UniqueClass.create do
```
subvertallchris: Can use `stub_active_node_class` here

spec/e2e/id_property_spec.rb
```diff
@@ -73,7 +73,25 @@
...
+      end
+
+      it 'it will find node by neo_id' do
+        node = NeoIdTest.new
+        expect(node).to respond_to(:id)
+        expect(node).to respond_to(:neo_id)
+        node.save
+        expect(node.id).to eq(node.neo_id)
+        expect(NeoIdTest.where(id: node).first).to eq(node)
       end
```
subvertallchris: Can we remove some of these `expect` calls or split them into different examples?

spec/e2e/id_property_spec.rb
```diff
@@ -340,38 +366,153 @@ def foobar
...
+    describe 'property id' do
+      it 'is not defined when before save ' do
+        node = NeoIdTest.new
+        expect(node.id).to be_nil
+      end
+
+      it 'is is set when saving ' do
+        node = NeoIdTest.new
+        node.save
+        expect(node.id).to be_present
```
subvertallchris: This is clear if you do something like

```ruby
expect { node.save }.to change { node.id.present? }.from(true).to(false)
```

It'd also let you remove the example above that checks for `nil` prior to save.

spec/e2e/id_property_spec.rb
```diff
@@ -340,38 +366,153 @@ def foobar
...
+
+    describe 'find_by_id' do
+      it 'finds it if it exists' do
+        NeoIdTest.create
+        node = NeoIdTest.create
+        NeoIdTest.create
+
+        found = NeoIdTest.find_by_id(node.id)
+        expect(found).to eq(node)
+      end
```
subvertallchris: What's the purpose of the extra nodes being created?

subvertallchris:
I really like that this helps us get away from explicit calls to `uuid` and lets the user trust that `#id` will be translated as needed. I'm not crazy about extra rounds of testing in Travis. Can we define shared examples or test a smaller section of the code with both property-based ID property and neo_id ID property within one round of tests?

spec/e2e/id_property_spec.rb
klobuczek: Actually `stub_active_node_class` is just a shortcut for what was there before. Since constraints are not cleaned up before each test, some other tests are bleeding into this specs and make it fail if the class is not unique.

spec/e2e/id_property_spec.rb
klobuczek: I just copied 1 of the existing tests for uuid. I guess to ensure it does not find just a random node.

klobuczek:
The extra round of testing was very helpful initially to find places where uuid (and consequently neo_id) were not translated into id (order, pluck, find_in_batches). All those are aspects not inherently related to neo_id and thus tested with the default config as well. So I think we could simply remove the extra entry from the build matrix.

cheerfulstoic:
Yeah, I'd be down with one round of tests (we can expand if we find later that this is insufficient).

Honestly I think that we could stand to lose some of the items in the build matrix

spec/e2e/id_property_spec.rb
klobuczek: Removing all constraints before each spec makes several spec fail as they meanwhile expect the bleeding over. I think this has to be fixed, but it is somehow our of the context of this pull request,

cheerfulstoic: Yeah, agreed.  Though I imagine we could remove all indexes/constraints before each full suite run and still get benefits of not having old ones lying around

cheerfulstoic: If you don't want to do this in this PR I would be just as happy with a quick issue covering it

klobuczek:
@cheerfulstoic @subvertallchris we have partially migrated our models to neo4j. And it's looking good so far. There is one aspect which active_node can do in one query and which I am still missing here: loading a tree of arbitrary depth with 1 query. But this is an unrelated topic for another pull request. Could we go ahead with this pull request?

cheerfulstoic:
Commenting here because Gitter flowed on.  @subvertallchris Do you see any reason not to merge this PR?

subvertallchris:
Nope, let's go with it.

@klobuczek, apologies for sitting on this for so long.

subvertallchris:
...once those conflicts are fixed. ;-)

klobuczek:
@subvertallchris all should be good now

subvertallchris:
Let's do it. Can you PR or just post text for docs about using this feature?

subvertallchris:
Also, this is great work!

PR #1178, opened 2016-04-10
Fix multiparam time

subvertallchris:
Fixes #

Bug re: multiparameter attributes for properties of type `Time`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@ All notable changes to this project will be documented in this file.
 This file should follow the standards specified on [http://keepachangelog.com/]
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [UNRELEASED]
+
+### Fixed
+
+- Multiparameter Attributes for properties of type `Time` were failing due to a hack that should have been removed with `ActiveAttr`'s removal
+
 ## [7.0.1] - 03-22-2016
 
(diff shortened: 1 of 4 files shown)


subvertallchris:
I really feel like we need to do something about JRuby. The amount of time spent messing with tests to satisfy it relative to number of users relying on embedded is so unbalanced.

subvertallchris:
Travis doesn't know what it's talking about.

cheerfulstoic:
Agreed!  Currently we've got a grid based on:

 * jruby 1.7.x / 9.0.4.0
 * neo4j 2.1.8 / 2.2.6 / 2.3.0

Perhaps we could drop 2.2.6 and assume that between testing it in MRI and it being between 2.1.8 and 2.3.0 that we'll be OK?

We'll also want to include Neo4j 3.0 soon.  Perhaps we should test:

 * jRuby 9.x with 2.1.x, 2.3.x, 3.0.x (testing most versions of the DB with the latest jRuby)
 * jRuby 1.7.x with 2.3.x (and eventually replace that with 3.0.x once 3.0.x has been out for a while).

That brings us from six jRuby builds down to four and would add testing for 3.0

Maybe we could do the same with MRI, too.  Test the latest version of MRI with all of the DBs and a couple of old versions with the latest (2.3.x or 3.0.x)

PR #1185, opened 2016-04-27
Rails 5 Beta 3 support

agustinvinao:
Fixes #

This pull introduces/changes:
 * Rails 5 beta 3 support
 * New generator to install an initializer for rails5 or update application.rb in rails < 5

Pings:
@cheerfulstoic
@subvertallchris


lib/neo4j/railtie.rb
```diff
@@ -4,7 +4,7 @@
 require 'action_dispatch/railtie'
 
 module Neo4j
-  class Railtie < ::Rails::Railtie
+  class Railtie < Rails::Railtie
```
cheerfulstoic: Interesting...  What made it so that not specifying the root `Rails` module fixed a problem?  I would expect this to break if we later, for whatever reason, make a `Neo4j::Rails` module

lib/rails/generators/neo4j/install/install_generator.rb
```diff
@@ -0,0 +1,27 @@
...
+      inject_into_file 'config/application.rb', File.read("../../templates/neo4j_rails4.rb"), :after => '[:password]'
+    end
+
+    # we are using JRuby - use Threads !
+    gsub_file "config/environments/production.rb", "# config.threadsafe!", "config.threadsafe!"
+  end
+
+  protected
+
+  def rails_5?
```
cheerfulstoic: Would like to see this method be something more descriptive.  I'm not sure what it might be, but something like `use_template` to reflect whatever changed in Rails` features.

Also, maybe this should be a `>=` instead of `==`?

lib/neo4j/railtie.rb
agustinvinao: I was checking edge api http://edgeapi.rubyonrails.org/classes/Rails/Railtie.html and they dont use ::Rails::Railtie, I'll check it with that.
I think what it works here is the require line about neo4j/railtie. I'll do more tests.

lib/rails/generators/neo4j/install/templates/neo4j_rails.rb
```diff
@@ -0,0 +1,8 @@
+# Enable Neo4j generators, e.g:  rails generate model Admin --parent User
+config.generators do |g|
+  g.orm             :neo4j
+  g.test_framework  :rspec, :fixture => false
```
cheerfulstoic: We use `rubocop` in the project, so I would run:

    rubocop -aD

and fix anything that it doesn't automatically fix before committing the changes.

I think that Travis is still failing for other reasons (maybe because of the things that @subvertallchris was still going to fix?)

lib/rails/generators/neo4j/install/install_generator.rb
agustinvinao: You're completely right, sorry for the crappy code, it was part of many hours checking code trying to make it work. I'll improve the code and make a new PR I think.

lib/rails/generators/neo4j/install/templates/neo4j_rails5.rb
```diff
@@ -0,0 +1,9 @@
+# Configure NEO4J as the ORM
+Rails.application.config.generators do |g|
+  g.orm             :neo4j
+end
+
+# Configure where the neo4j database should exist
+Neo4j::Config.use do |config|
+  config[:storage_path] = "#{Rails.application.config.root}/db/neo4j-#{Rails.env}"
```
cheerfulstoic: Is `Rails.application.config.root` a `Pathname` object?  If so this could be:

    Rails.application.config.root.join("db/neo4j-#{Rails.env}")

But a couple of things about this: People could be using Neo4j in either server or embedded modes and if it's in server mode they could be setting it up in many ways (via the desktop app, downloading the `tar.gz`, docker, using the `rake` tasks, etc....).  So I'm not sure that it makes sense to set up a default `storage_path` or `url`

neo4j.gemspec
```diff
@@ -27,12 +27,12 @@ A Neo4j OGM (Object-Graph-Mapper) for use in Ruby on Rails and Rack frameworks h
   s.rdoc_options = ['--quiet', '--title', 'Neo4j.rb', '--line-numbers', '--main', 'README.rdoc', '--inline-source']
 
   s.add_dependency('orm_adapter', '~> 0.5.0')
-  s.add_dependency('activemodel', '>= 4')
-  s.add_dependency('activesupport', '>= 4')
+  s.add_dependency('activemodel', '~> 5.0.0.beta3')
+  s.add_dependency('activesupport', '~> 5.0.0.beta3')
```
cheerfulstoic: We definitely want to continue to support `>= 4` for `activemodel`, `activesupport`, and `railties`.  Maybe when you're testing you can use `5.0.0.beta3` in your `Gemfile`?

cheerfulstoic:
Thanks so much for looking into this!  I made a few comments in the code.  It would also be useful to me to know what sort of issues you ran into.

Definitely interested in @subvertallchris 's comments

lib/rails/generators/neo4j/install/install_generator.rb
cheerfulstoic: I understand the feeling, thanks ;)

lib/rails/generators/neo4j/install/templates/neo4j_rails5.rb
agustinvinao: I think we need to create some strategy to detect or offer some alternative to maintain rails philosophy about convention over configuration.
Offer a default setup and allow the user to change it, maybe using the neo4j.yml file.

cheerfulstoic: Yeah, agreed about convention.  `http://localhost:7474` is actually the default in further down in `neo4j-core`, but I'm not sure what it should be for embedded.  I actually just realized I was going to mention before that we already have a convention for how the rake tasks create databases (for server mode, though maybe it could be shared behavior) in:

    db/neo4j/environment

lib/rails/generators/neo4j/install/install_generator.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    # ORM Config
+    if rails_5?
+      template 'neo4j_rails5.rb', 'config/initializers/neo4j.rb'
+    else
+      inject_into_file 'config/application.rb', File.read("../../templates/neo4j_rails4.rb"), :after => '[:password]'
+    end
+
+    # we are using JRuby - use Threads !
+    gsub_file "config/environments/production.rb", "# config.threadsafe!", "config.threadsafe!"
```
subvertallchris: This isn't true. Not everyone using this gem uses JRuby.

agustinvinao: I'll add a detection of the ruby version it uses in production to use threads.

subvertallchris: Can you link me to that? It's probably leftover from the legacy version of the gem. I'm not sure that this change is the gem's responsibility in the first place.

agustinvinao: From the first video of the seris about neo4j and rails, explain how to use the command: 
rails new myapp -m http://andreasronge.github.com/neo4j/rails.rb -O

That file http://neo4jrb.io/neo4j/neo4j.rb, you can see it here http://neo4jrb.io/neo4j/

And line 12 does that change in production.

cheerfulstoic: That page is out of date.  I just put in a redirect so that it goes to http://neo4jrb.readthedocs.io/

Where is the original video that you saw?  I have the original video on youtube but it should have a note over it redirecting people to the first video in the six-episode screencast series.

subvertallchris:
I'm about to add support for Rails 5 into the 7.1.x branch, we'll do the release shortly after that.

bloomdido:
I've been using Rails 5 (beta2, beta3, rc2, and now 5.0.0) for the last couple of months. Is there any known issues or areas I should be looking to test?

subvertallchris:
@bloomdido Nothing that I'm aware of.

We just released 7.1.0, which provides Rails 5 support. I'm going to merge those changes back into `master`, so please target your projects accordingly. We'll be deleting our other feature branches soon.

PR #1186, opened 2016-04-27
Added a config option for the validations failure behavior within transactions.

### LINKED ISSUE #1156: Strange behavior with transactions and `save`
I have a model like this:
```ruby
class Resource
  include Neo4j::ActiveNode
  ...
  property :url, type: String, constraint: :unique
  validates :url, presence: true, uniqueness: true
end
```
I tried this code:

```ruby
old = Resource.new(url: 'http://www.google.it')
old.save # => true

Neo4j::Transaction.run do |tx|
  new_res = Resource.new(url: 'http://www.google.it')
  puts tx.failed?  # => false
  new_res.save # => false
  puts tx.failed? # => true
  # many other things that doesn't fail
end
# HERE EVERYTHING IS SILENTLY ROLLBACKED
```

This behavior is veeeeery strange, since I expect that `save` is not like `save!`, so the transaction should not fail when if returns false.

ProGM:
Fixes #1156 

This pull introduces/changes:
 An option for `Neo4j::Config` to choose what to do when validations fail inside a transaction:

```ruby
Neo4j::Config[:fail_transaction_when_validations_fail] = false
```

It's set to `true` by default, to keep the backward compatibility.

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/spec/e2e/transaction_spec.rb
+++ b/spec/e2e/transaction_spec.rb
@@ -25,9 +25,8 @@ describe 'Neo4j::Transaction' do
   end
 
   describe 'transaction behaviour when a validation fails' do
-    let(:clazz) do
-      UniqueClass.create do
-        include Neo4j::ActiveNode
+    before(:each) do
+      stub_active_node_class('Clazz') do
         property :name
         validates :name, presence: true
       end
@@ -36,24 +35,26 @@ describe 'Neo4j::Transaction' do
     it 'rollbacks the current transaction by default' do
       expect do
         Neo4j::Transaction.run do |tx|
-          clazz.create
+          Clazz.create
           expect(tx).to be_failed
-          clazz.create(name: 'john')
+          Clazz.create(name: 'john')
         end
-      end.not_to change { clazz.count }
+      end.not_to change { Clazz.count }
     end
 
-    it 'keeps running the current transaction when `fail_transaction_when_validations_fail` is false' do
-      Neo4j::Config[:fail_transaction_when_validations_fail] = false
+    context 'when `fail_transaction_when_validations_fail` is disabled' do
+      let_config(:fail_transaction_when_validations_fail) { false }
 
-      expect do
-        Neo4j::Transaction.run do |tx|
-          clazz.create
-          expect(tx).not_to be_failed
-          clazz.create(name: 'john')
-        end
-      end.to change { clazz.count }.by(1)
-      Neo4j::Config.delete(:fail_transaction_when_validations_fail)
+      it 'keeps running the current transaction' do
+        expect do
+          Neo4j::Transaction.run do |tx|
+            Clazz.create
+            expect(tx).not_to be_failed
+            Clazz.create(name: 'john')
+          end
+        end.to change { Clazz.count }.by(1)
+        Neo4j::Config.delete(:fail_transaction_when_validations_fail)
+      end
     end
   end
 end
--- a/spec/e2e/transaction_spec.rb
+++ b/spec/e2e/transaction_spec.rb
@@ -33,7 +33,7 @@ describe 'Neo4j::Transaction' do
       end
     end
 
-    it 'rollbacks the current transaction' do
+    it 'rollbacks the current transaction by default' do
       expect do
         Neo4j::Transaction.run do |tx|
           clazz.create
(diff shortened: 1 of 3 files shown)


ProGM:
Uhm, seems like ruby 2.0 has some problems with the gem `listen (3.1.1)`. 😞

spec/e2e/transaction_spec.rb
```diff
@@ -23,4 +23,37 @@
...
+        Neo4j::Transaction.run do |tx|
+          clazz.create
+          expect(tx).to be_failed
+          clazz.create(name: 'john')
+        end
+      end.not_to change { clazz.count }
+    end
+
+    it 'keeps running the current transaction when `fail_transaction_when_validations_fail` is false' do
+      Neo4j::Config[:fail_transaction_when_validations_fail] = false
```
cheerfulstoic: There's a `let_config` helper in the `spec_helper` to temporarily define configuration values.  Definition:

https://github.com/neo4jrb/neo4j/blob/master/spec/spec_helper.rb#L116

Example:

https://github.com/neo4jrb/neo4j/blob/master/spec/e2e/active_model_spec.rb#L207

spec/e2e/transaction_spec.rb
```diff
@@ -23,4 +23,37 @@
...
     end
   end
+
+  describe 'transaction behaviour when a validation fails' do
+    let(:clazz) do
+      UniqueClass.create do
+        include Neo4j::ActiveNode
+        property :name
+        validates :name, presence: true
+      end
```
cheerfulstoic: using `stub_active_node_class` is the new, better way, to do this:

```ruby
stub_active_node_class('Clazz') do
  property :name
  validates :name, presence: true
end
```

Then you would refer to `Clazz` instead of `clazz`

cheerfulstoic:
I think that this is good, but I'm having trouble understanding why we fail transactions when validations fail.  Do you know if `ActiveRecord` does this?  I wonder about changing the default behavior (maybe just going forward with >= 8.0).  Otherwise the configuration seems oddly specific.

ProGM:
Uhm... I tried the exactly same thing with ActiveRecord and... yes, validations don't make transactions to fail.

Here's the example I tried:

Model:
```ruby
class User < ActiveRecord::Base
  validates :first_name, :last_name, presence: true
end
```

In console (assuming that there's already at least one user):
```ruby
irb(main):008:0> ActiveRecord::Base.transaction do
irb(main):009:1* u = User.new.save
irb(main):010:1> User.first.update(first_name: 'johnny')
irb(main):011:1> end
   (37.2ms)  BEGIN
  User Load (0.4ms)  SELECT  "users".* FROM "users"  ORDER BY "users"."id" ASC LIMIT 1
  SQL (0.3ms)  UPDATE "users" SET "first_name" = $1, "updated_at" = $2 WHERE "users"."id" = $3  [["first_name", "johnny"], ["updated_at", "2016-05-01 17:25:29.468313"], ["id", 1]]
   (5.6ms)  COMMIT
=> true
```

cheerfulstoic:
Oh, and I did the `< 3.1` in the `Gemfile`, so hopefully that will help.  Try merging from upstream

spec/e2e/transaction_spec.rb
ProGM: Ok, fixing :D

spec/e2e/transaction_spec.rb
ProGM: Oh, very useful!

cheerfulstoic:
Yeah, then I say we drop that bit of logic going forward toward v8.0.  It seems to have been added here:

https://github.com/neo4jrb/neo4j/commit/010eb20876ed81386ac4ee63adee5847c7c1b8f3

And it was changed when @subvertallchris made `ActiveRel`:

https://github.com/neo4jrb/neo4j/commit/53688e805eabccb51065e64b508f6ace246ac6bb

What do you think @subvertallchris ?  I'm not sure that this will affect too many people, but we can have it as part of the `CHANGELOG` to warn people.

ProGM:
Uhm, ok! In the meanwhile I fixed the two issues you reported and I merged with upstream 👍

cheerfulstoic:
Awesome, thanks ;) :metal:

subvertallchris:
Yeah, I'd say that we should make this the default in 8.0. It's odd, for sure.

cheerfulstoic:
:+1:

@ProGM could you make another PR to simply remove this code from master and any associated specs?

ProGM:
Done! #1192 

Should we close this?

cheerfulstoic:
Yup!  Merged #1192 and closing this PR.  Thanks!

PR #1191, opened 2016-05-02
fix ActiveNode mass-assignment association handling

### LINKED ISSUE #1182: Problems with `create!` with validations
Hi there,
I think I found a bug with `create!` method. It seems not to accept relations as parameters. Instead, if I use `new(params).save!` It works well

Here's an example I tried:

```ruby
class Article
  include Neo4j::ActiveModel
  property :title
  has_one :out, :user, type: :author
  validates :user, presence: true
end

Article.create!(title: 'test', user: User.first)
# => Neo4j::ActiveNode::Persistence::RecordInvalidError: User can't be blank

Article.new(title: 'test', user: User.first).save!
# => true
```

### Runtime information:

Neo4j database version: 2.2.7
`neo4j` gem version: 7.0.0
`neo4j-core` gem version: 6.1.2

subvertallchris:
Fixes #1182

Fixes the issue described. We're already able to handle associations with save, so create should just let it do its job.

@cheerfulstoic, I couldn't find a better place to put these specs. This seemed like the obvious choice since the methods are defined in persistence.rb, but let me know if there's a better place for them.

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/lib/neo4j/active_node/persistence.rb
+++ b/lib/neo4j/active_node/persistence.rb
@@ -103,27 +103,17 @@ module Neo4j::ActiveNode
       # Creates and saves a new node
       # @param [Hash] props the properties the new node should have
       def create(props = {})
-        association_props = extract_association_attributes!(props) || {}
         new(props).tap do |obj|
           yield obj if block_given?
           obj.save
-          association_props.each do |prop, value|
-            obj.send("#{prop}=", value)
-          end
         end
       end
 
       # Same as #create, but raises an error if there is a problem during save.
-      def create!(*args)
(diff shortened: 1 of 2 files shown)


cheerfulstoic:
All seems good to me!  Probably those are a good enough place for the specs.  Perhaps at some point we should re-think our spec organization.  For one thing, as I've been doing the new API refactor I've been moving some unit specs to e2e because a lot of our unit specs aren't really unit specs anymore because they need to `include` lots of other modules for things to work (maybe that's a problem with the way we structure our modules?)

PR #1192, opened 2016-05-03
Removed transaction failure when validations fail

### LINKED ISSUE #1156: Strange behavior with transactions and `save`

ProGM:
Fixes #1156 

This pull introduces/changes:
 * The default behavior with transactions, when a validation fails.

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- Changed the behavior with transactions when a validation fails. This is a potentially breaking change, since now calling `save` would not fail the current transaction, as expected. (see #1156)
 - Invalid options to the `property` method now raise an exception (see #1169)
 
 ### Added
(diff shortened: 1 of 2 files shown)


cheerfulstoic:
Wunderbar, thanks for helping us figure this out!

PR #1197, opened 2016-05-08
Neo4j migrations

ProGM:
This pull introduces an `ActiveRecord` like migration method.

It's W.I.P., but I have a small thing I want to discuss before proceeding:

I'm using this conventions: migration files are searched in `db/neo4j/migrate` folder using the same strategy as ActiveRecord.
Migrations state is saved in a `SchemaMigration` node.

Since i wanted to copy the ActiveRecord implementation, I created this set of tasks:

```bash
rake neo4j::migrate # Alias of ::all
rake neo4j::migrate::all
rake neo4j::migrate::up VERSION=some_version
rake neo4j::migrate::down VERSION=some_version
rake neo4j::rollback
```

However, the first one is conflicting the current neo4j migration task. For now, I renamed it `rake neo4j::legacy_migrate`. What do you think about it?

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/docs/Migrations.rst
+++ b/docs/Migrations.rst
@@ -1,0 +1,260 @@
+Migrations
+==================
+
+Neo4j.rb provides an ``ActiveRecord``-like migration framework and a set of helper methods to manipulate both database schema and data.
+
+
+Generators
+------------------
+
+Migrations can be created by using the built-in Rails generator:
+
+.. code-block:: bash
+
+  rails generate neo4j:migration RenameUserNameToFirstName
+
+This will generate a new file located in ``db/neo4j/migrate/xxxxxxxxxx_rename_user_name_to_first_name.rb``
+
+.. code-block:: ruby
+  class RenameUserNameToFirstName < Neo4j::Migrations::Base
+    def up
+    end
+
+    def down
+    end
+  end
+
+In the same way as ``ActiveRecord`` does, you should fill up the ``up`` and ``down`` methods to define the migration and (eventually) the rollback steps.
+
+
+Transactions
+------------------
+Every migrations runs inside a transaction by default. So, if some statement fails inside a migration fails, the database rollbacks to the previous state.
+
+However this behaviour is not always good. For instance, neo4j doesn't allow schema and data changes in the same transaction.
+
+To disable this, you can use the ``disable_transactions!`` helper in your migration definition:
+
+.. code-block:: ruby
+
+  class SomeMigration < Neo4j::Migrations::Base
+    disable_transactions!
+
+    ...
+  end
+
+Neo4j::Migrations::Helpers
+------------------
+
+#execute
+~~~~~~
+
+Executes a pure neo4j cypher query, interpolating parameters.
+
+.. code-block:: ruby
+
+  execute('MATCH (n) WHERE n.name = {node_name} RETURN n', node_name: 'John')
+
+.. code-block:: ruby
+
+  execute('MATCH (n)-[r:`friend`]->() WHERE n.age = 7 DELETE r')
+
+
+#query
+---------------
+
+An alias for ``Neo4j::Session.query``. You can use it as root for the query builder:
+
+.. code-block:: ruby
+
+  query.match(:n).where(name: 'John').delete(:n)
+
+
+#remove_property
+---------------
+
+Removes a property given a label.
+
+:Ruby:
+  .. code-block:: ruby
+
+    remove_property(:User, :money)
+
+:Cypher:
+  .. code-block:: cypher
+
+    MATCH (n:`User`) REMOVE n.money
+
+#rename_property
+------
+
+Renames a property given a label.
+
+:Ruby:
+  .. code-block:: ruby
+
+    rename_property(:User, :name, :first_name)
+
+:Cypher:
+  .. code-block:: cypher
+
+    MATCH (n:`User`) SET n.first_name = n.name REMOVE n.name
+
+#drop_nodes
+------
+
+Removes all nodes with a certain label
+
+:Ruby:
+  .. code-block:: ruby
+
+    drop_nodes(:User)
+
+:Cypher:
+  .. code-block:: cypher
+
+    MATCH (n:`User`) OPTIONAL MATCH (n)-[r]-() DELETE r,n
+
+#add_label
+------
+
+Adds a label to nodes, given their current label
+
+:Ruby:
+  .. code-block:: ruby
+
+    add_label(:User, :Person)
+
+:Cypher:
+  .. code-block:: cypher
+
+    MATCH (n:`User`) SET n:`Person`
+
+#add_labels
+------
+
+Adds labels to nodes, given their current label
+
+:Ruby:
+  .. code-block:: ruby
+
+    add_label(:User, [:Person, :Boy])
+
+:Cypher:
+  .. code-block:: cypher
+
+    MATCH (n:`User`) SET n:`Person`:`Boy`
+
+#remove_label
+------
+
+Removes a label from nodes, given a label
+
+:Ruby:
+  .. code-block:: ruby
+
+    remove_label(:User, :Person)
+
+:Cypher:
+  .. code-block:: cypher
+
+    MATCH (n:`User`) REMOVE n:`Person`
+
+#remove_labels
+------
+
+Removes labels from nodes, given a label
+
+:Ruby:
+  .. code-block:: ruby
+
+    remove_label(:User, [:Person, :Boy])
+
+:Cypher:
+  .. code-block:: cypher
+
+    MATCH (n:`User`) REMOVE n:`Person`:`Boy`
+
+#rename_label
+------
+
+Renames a label
+
+:Ruby:
+  .. code-block:: ruby
+
+    rename_label(:User, :Person)
+
+:Cypher:
+  .. code-block:: cypher
+
+    MATCH (n:`User`) SET n:`Person` REMOVE n:`Person`
+
+#drop_constraint
+------
+
+Drops an unique constraint on a given label attribute.
+
+**Warning** it would fail unless you define ``disable_transactions!`` in your migration file.
+
+.. code-block:: ruby
+
+  drop_constraint(:User, :name)
+
+#drop_index
+------
+
+Drops an exact index on a given label attribute.
+
+**Warning** it would fail unless you define ``disable_transactions!`` in your migration file.
+
+.. code-block:: ruby
+
+  drop_index(:User, :name)
+
+
+#say
+------
+
+Writes some text while running the migration.
+
+:Ruby:
+  .. code-block:: ruby
+
+    say 'Hello'
+
+:Output:
+  .. code-block:: ruby
+
+    -- Hello
+
+When passing ``true`` as second parameter, it writes it more indented.
+
+:Ruby:
+  .. code-block:: ruby
+
+    say 'Hello', true
+
+:Output:
+  .. code-block:: ruby
+
+      -> Hello
+
+#say_with_time
+------
+
+Wraps a set of statements inside a block, printing the given and the execution time. When an ``Integer`` is returned, it assumes it's the number of affected rows.
+
+:Ruby:
+  .. code-block:: ruby
+
+    say_with_time 'Trims all names' do
+      query.match(n: :User).set('n.name = TRIM(n.name)').pluck('count(*)')
+    end
+
+:Output:
+  .. code-block:: bash
+
+    -- Trims all names.
+       -> 0.3451s
+       -> 2233 rows
--- a/docs/Migrations.rst
+++ b/docs/Migrations.rst
@@ -1,7 +1,7 @@
 Migrations
 ==================
 
-Neo4j.rb provides an ``ActiveRecord``-like migration framework and a set of helper methods to manipulate both database schema and data.
+Neo4j does not have a set schema like relational databases, but sometimes changes to the schema and the data are required. To help with this, Neo4j.rb provides an ``ActiveRecord``-like migration framework and a set of helper methods to manipulate both database schema and data.
 
 
 Generators
@@ -18,9 +18,11 @@ This will generate a new file located in ``db/neo4j/migrate/xxxxxxxxxx_rename_us
 .. code-block:: ruby
   class RenameUserNameToFirstName < Neo4j::Migrations::Base
     def up
+      rename_property :User, :name, :first_name
     end
 
     def down
+      rename_property :User, :first_name, :name
     end
   end
 
@@ -43,7 +45,7 @@ To disable this, you can use the ``disable_transactions!`` helper in your migrat
     ...
   end
 
-Neo4j::Migrations::Helpers
+Migration Helpers
 ------------------
 
 #execute
@@ -67,7 +69,7 @@ An alias for ``Neo4j::Session.query``. You can use it as root for the query buil
 
 .. code-block:: ruby
 
-  query.match(:n).where(name: 'John').delete(:n)
+  query.match(:n).where(name: 'John').delete(:n).exec
 
 
 #remove_property
@@ -195,7 +197,7 @@ Renames a label
 
 Drops an unique constraint on a given label attribute.
 
-**Warning** it would fail unless you define ``disable_transactions!`` in your migration file.
+**Warning** it would fail if you make data changes in the same migration. To fix, define ``disable_transactions!`` in your migration file.
 
 .. code-block:: ruby
 
@@ -206,7 +208,7 @@ Drops an unique constraint on a given label attribute.
 
 Drops an exact index on a given label attribute.
 
-**Warning** it would fail unless you define ``disable_transactions!`` in your migration file.
+**Warning** it would fail if you make data changes in the same migration. To fix, define ``disable_transactions!`` in your migration file.
 
 .. code-block:: ruby
 
--- a/docs/Migrations.rst
+++ b/docs/Migrations.rst
@@ -77,120 +77,72 @@ An alias for ``Neo4j::Session.query``. You can use it as root for the query buil
 
 Removes a property given a label.
 
-:Ruby:
-  .. code-block:: ruby
+.. code-block:: ruby
 
-    remove_property(:User, :money)
-
-:Cypher:
-  .. code-block:: cypher
-
-    MATCH (n:`User`) REMOVE n.money
+  remove_property(:User, :money)
 
 #rename_property
 ------
 
 Renames a property given a label.
 
-:Ruby:
-  .. code-block:: ruby
+.. code-block:: ruby
 
-    rename_property(:User, :name, :first_name)
-
-:Cypher:
-  .. code-block:: cypher
-
-    MATCH (n:`User`) SET n.first_name = n.name REMOVE n.name
+  rename_property(:User, :name, :first_name)
 
 #drop_nodes
 ------
 
 Removes all nodes with a certain label
 
-:Ruby:
-  .. code-block:: ruby
+.. code-block:: ruby
 
-    drop_nodes(:User)
-
-:Cypher:
-  .. code-block:: cypher
-
-    MATCH (n:`User`) OPTIONAL MATCH (n)-[r]-() DELETE r,n
+  drop_nodes(:User)
 
 #add_label
 ------
 
 Adds a label to nodes, given their current label
 
-:Ruby:
-  .. code-block:: ruby
+.. code-block:: ruby
 
-    add_label(:User, :Person)
-
-:Cypher:
-  .. code-block:: cypher
-
-    MATCH (n:`User`) SET n:`Person`
+  add_label(:User, :Person)
 
 #add_labels
 ------
 
 Adds labels to nodes, given their current label
 
-:Ruby:
-  .. code-block:: ruby
+.. code-block:: ruby
 
-    add_label(:User, [:Person, :Boy])
-
-:Cypher:
-  .. code-block:: cypher
-
-    MATCH (n:`User`) SET n:`Person`:`Boy`
+  add_label(:User, [:Person, :Boy])
 
 #remove_label
 ------
 
 Removes a label from nodes, given a label
 
-:Ruby:
-  .. code-block:: ruby
+.. code-block:: ruby
 
-    remove_label(:User, :Person)
-
-:Cypher:
-  .. code-block:: cypher
-
-    MATCH (n:`User`) REMOVE n:`Person`
+  remove_label(:User, :Person)
 
 #remove_labels
 ------
 
 Removes labels from nodes, given a label
 
-:Ruby:
-  .. code-block:: ruby
+.. code-block:: ruby
 
-    remove_label(:User, [:Person, :Boy])
-
-:Cypher:
-  .. code-block:: cypher
-
-    MATCH (n:`User`) REMOVE n:`Person`:`Boy`
+  remove_label(:User, [:Person, :Boy])
 
 #rename_label
 ------
 
 Renames a label
 
-:Ruby:
-  .. code-block:: ruby
+.. code-block:: ruby
 
-    rename_label(:User, :Person)
-
-:Cypher:
-  .. code-block:: cypher
-
-    MATCH (n:`User`) SET n:`Person` REMOVE n:`Person`
+  rename_label(:User, :Person)
 
 #drop_constraint
 ------
@@ -251,7 +203,7 @@ Wraps a set of statements inside a block, printing the given and the execution t
   .. code-block:: ruby
 
     say_with_time 'Trims all names' do
-      query.match(n: :User).set('n.name = TRIM(n.name)').pluck('count(*)')
+      query.match(n: :User).set('n.name = TRIM(n.name)').pluck('count(*)').first
     end
 
 :Output:
--- a/lib/neo4j.rb
+++ b/lib/neo4j.rb
@@ -86,12 +86,11 @@ require 'neo4j/active_node/query'
 require 'neo4j/active_node/scope'
 require 'neo4j/active_node'
 
-require 'neo4j/migration'
-require 'neo4j/migrations/helpers'
-require 'neo4j/migrations/migration_file'
-require 'neo4j/migrations/base'
-require 'neo4j/migrations/runner'
-require 'neo4j/migrations/schema_migration'
+module Neo4j
+  extend ActiveSupport::Autoload
+  autoload :Migration
+  autoload :Migrations
+end
 
 require 'neo4j/active_node/orm_adapter'
 if defined?(Rails)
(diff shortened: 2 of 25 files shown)


cheerfulstoic:
I would probably rename it `migrate_task`, though @subvertallchris  wrote the migration stuff initially.  What do you think Chris?

subvertallchris:
We can change the old task but I think that `migrate` and `migrate_task` is confusing. Can we think of a better name?

I like the way this is implemented but I'm not sure how it'll work in practice since all of our schema settings are mixed throughout models. What would your ideal workflow be to take advantage of this?

ProGM:
An alternative naming convention could be keeping the old `neo4j::migrate`, and renaming the new migrations like this:

```bash
rake db:neo4j:migrate # Alias of ::all
rake db:neo4j:migrate::all
rake db:neo4j:migrate::up VERSION=some_version
rake db:neo4j:migrate::down VERSION=some_version
rake db:neo4j:rollback
```

>  What would your ideal workflow be to take advantage of this?

I think that neo4j migrations will be used mostly with cypher queries.

Some examples:

"Drop a table"-like migration
```ruby
class DropNodeMyLabel < Neo4j::Migrations::Base
  def up
    execute 'MATCH (n:`MyLabel`)-[r]-() DELETE n, r'
  end
  def down
    raise Neo4j::IrreversibleMigration
  end
end
```

Renaming a property.
```ruby
class RenameTitleToName < Neo4j::Migrations::Base
  def up
    execute 'MATCH (c:Category) where c.name is null SET c.name = c.title REMOVE c.title'
  end
  def down
    execute 'MATCH (c:Category) where c.title is null SET c.title = c.name REMOVE c.name'
  end
end
```

Maybe we could define some helpers for this actions, like `rename_property` or `drop_nodes`, or something similar :)

What do you think?

@subvertallchris @cheerfulstoic

subvertallchris:
Ah, I see. For tasks like that, yes, it'd be really nice to have and makes sense as written. I was fixated on the index and constraint definitions.

As for naming, we can deprecate them entirely if you want. We could move the UUID task somewhere else entirely. I don't want to confuse people with similar names and this makes more sense as _the_ `migrate`.

cheerfulstoic:
Yeah, I don't think there would be many schema migrations, though we'd probably want to have `drop_index` and `drop_constraint` helpers since removing the model definitions don't remove the them.

I think that `rename_property` would be good.  Probably it would take a label?

What would `drop_nodes` do?  Would that be like `Model.delete_all`?

Thinking through other cases, perhaps `add_label`, though I could see that as being something that we add to scoping in general like `Model.where(foo: 1).add_label(:Bar)`.  Though I guess that could just be `Model.where(foo: 1).query_as(:m).set('m:Bar').exec`.

ProGM:
@cheerfulstoic @subvertallchris I did it, please check 👍

(Any better idea to test migration is welcome 😆)

lib/neo4j.rb
```diff
@@ -86,6 +86,13 @@
 require 'neo4j/active_node/scope'
 require 'neo4j/active_node'
 
+require 'neo4j/migration'
+require 'neo4j/migrations/helpers'
+require 'neo4j/migrations/migration_file'
+require 'neo4j/migrations/base'
+require 'neo4j/migrations/runner'
+require 'neo4j/migrations/schema_migration'
+
```
cheerfulstoic: I feel like this `require` practice is growing a bit too much ;)  I wonder about using Ruby's `autoload`:

http://www.rubyinside.com/ruby-techniques-revealed-autoload-1652.html

Migrations will generally run in the rake task process and then the process dies.  So if we use `autoload` we (I think) won't get the `Migration` code loaded into memory in the application, just in the rake task.

If it works well here, I think it's something that we see if it makes sense elsewhere.

lib/neo4j/migrations/helpers.rb
```diff
@@ -0,0 +1,50 @@
+module Neo4j
+  module Migrations
+    module Helpers
+      def rename_property(label, old_property, new_property)
+        by_label(label).where(n: {new_property => nil})
```
cheerfulstoic: I don't know that I would expect that this would only work on nodes where the `new_property` is `nil`.

My first instinct was to just blow away any values which are on the `new_property`, but perhaps we should run a check to see if the `new_property` is non-null anywhere and give an error if that's the case.

lib/neo4j/migrations/helpers.rb
```diff
@@ -0,0 +1,50 @@
...
+        execute("DROP INDEX ON :#{label}(#{property})")
+      end
+
+      def execute(string, params = {})
+        query(string, params).to_a
+      end
+
+      private
+
+      delegate :query, to: Neo4j::Session
```
cheerfulstoic: This `query` would be nice as a public helper.

lib/neo4j/migrations/helpers.rb
```diff
@@ -0,0 +1,50 @@
...
+    module Helpers
+      def rename_property(label, old_property, new_property)
+        by_label(label).where(n: {new_property => nil})
+                       .set("n.#{new_property} = n.#{old_property}")
+                       .remove("n.#{old_property}")
+                       .exec
+      end
+
+      def drop_nodes(label)
+        query.match("(n:`#{label}`)")
```
cheerfulstoic: You can also do: `query.match(n: label)`, I think.

Kudos on not going with the `DETACH DELETE` (though of course there's no method for it.  Just put a note on my list to add one ;)

lib/neo4j/migrations/helpers.rb
```diff
@@ -0,0 +1,50 @@
...
+             .optional_match('(n)-[r]-()')
+             .delete(:r, :n).exec
+      end
+
+      def add_labels(label, *labels)
+        by_label(label).set("n:#{labels.join(':')}").exec
```
cheerfulstoic: You might be able to do `.set(n: labels)` here.  If not it's something we might add

lib/neo4j/migrations/helpers.rb
```diff
@@ -0,0 +1,50 @@
...
+
+      def remove_labels(label, *labels)
+        by_label(label).remove("n:#{labels.join(':')}").exec
+      end
+
+      alias add_label add_labels
+      alias remove_label remove_labels
+
+      def remove_constraint(label, property)
+        execute("DROP CONSTRAINT ON (n:`#{label}`) ASSERT n.#{property} IS UNIQUE")
```
cheerfulstoic: I believe that you can do something like `Neo4j::Label.new(label). create_constraint(property, type: :unique)`.  @subvertallchris ?

subvertallchris: Indeed you can. See https://github.com/neo4jrb/neo4j-core/blob/master/lib/neo4j-server/cypher_label.rb and https://github.com/neo4jrb/neo4j-core/blob/master/lib/neo4j/label.rb. The interface might still have some rough edges. I gave it an overhaul a few months ago but it was borne from a very Java perspective that wasn't exactly friendly...

lib/neo4j/migrations/migration_file.rb
```diff
@@ -0,0 +1,24 @@
...
+    class MigrationFile
+      attr_reader :file_name, :symbol_name, :class_name, :version
+
+      def initialize(file_name)
+        @file_name = file_name
+        extract_data!
+      end
+
+      def create
+        require file_name
```
cheerfulstoic: Not a huge deal, but I generally prefer to refer to instance variables internally (like `@file_name`), not the accessors

lib/neo4j.rb
ProGM: I'll take a look 👍

lib/neo4j/migrations/helpers.rb
ProGM: Adding a test for this ;)

lib/neo4j/migrations/helpers.rb
ProGM: 👍

lib/neo4j/migrations/helpers.rb
ProGM: @cheerfulstoic Maybe I could fail by default if there's some new_property `nil`, and add a parameter `force`?

lib/neo4j/migrations/migration_file.rb
```diff
@@ -0,0 +1,24 @@
...
+        class_name.constantize.new(version)
+      end
+
+      private
+
+      def extract_data!
+        @version, @symbol_name = File.basename(file_name, '.rb').split('_', 2)
+        @class_name = @symbol_name.split('_').map(&:capitalize).join('')
```
cheerfulstoic: Would `@symbol_name.classify` work?

lib/neo4j/migrations/runner.rb
```diff
@@ -0,0 +1,106 @@
...
+    class Runner
+      STATUS_TABLE_FORMAT = '%-10s %-20s %s'.freeze
+      SEPARATOR = '--------------------------------------------------'.freeze
+      FILE_MISSING = '**** file missing ****'.freeze
+      STATUS_TABLE_HEADER = ['Status'.freeze, 'Migration ID'.freeze, 'Migration Name'.freeze].freeze
+      UP_MESSAGE = 'up'.freeze
+      DOWN_MESSAGE = 'down'.freeze
+
+      def initialize
+        @up_versions = SchemaMigration.all.pluck(:migration_id)
```
cheerfulstoic: Probably a minor performance improvement, but if this were a `Set` then the `include?` calls would be faster.

lib/neo4j/migrations/runner.rb
```diff
@@ -0,0 +1,106 @@
...
+      end
+
+      def all
+        migration_files.each do |migration_file|
+          next if @up_versions.include?(migration_file.version)
```
cheerfulstoic: Since this is done so much, maybe a private method called something like `up?`?

lib/neo4j/migrations/runner.rb
```diff
@@ -0,0 +1,106 @@
...
+      def status
+        output STATUS_TABLE_FORMAT, *STATUS_TABLE_HEADER
+        output SEPARATOR
+        all_migrations.each do |version|
+          status = @up_versions.include?(version) ? UP_MESSAGE : DOWN_MESSAGE
+          migration_file = find_by_version(version)
+          migration_name = migration_file ? migration_file.class_name : FILE_MISSING
+          output STATUS_TABLE_FORMAT, status, version, migration_name
+        end
+      end
```
cheerfulstoic: :+1: for this

cheerfulstoic: :+1: for this method

lib/neo4j/migrations/runner.rb
```diff
@@ -0,0 +1,106 @@
...
+          migration.migrate(:up)
+        end
+      end
+
+      def migrate_down(migration_file)
+        migration_message(migration_file, 'reverting', 'reverted') do
+          migration = migration_file.create
+          migration.migrate(:down)
+        end
+      end
```
cheerfulstoic: Maybe these methods could be one `migrate` method that takes a second `direction` argument of `:up` or `:down` and you could switch passing in `migrated` / `reverted` based on that argument.

lib/neo4j/migrations/runner.rb
```diff
@@ -0,0 +1,106 @@
...
+
+      def migration_message(migration, running_message, complete_message)
+        output "== #{migration.version} #{migration.class_name}: #{running_message}... ========="
+        yield
+        output "== #{migration.version} #{migration.class_name}: #{complete_message} ========="
+      end
+
+      def output(*string_format)
+        puts format(*string_format) unless ENV['silenced']
```
cheerfulstoic: I normally expect ENV vars to be all caps.  Also maybe it could be more specific like `MIGRATIONS_SILENCED`?

ProGM: I copy-pasted it from the old `Migration` module:
https://github.com/neo4jrb/neo4j/blob/master/lib/neo4j/migration.rb#L10
However, it obviously makes sense

lib/neo4j/migrations/runner.rb
ProGM: 👍

lib/neo4j/migrations/migration_file.rb
ProGM: Is ActiveSupport included? If so, ok 👍

lib/neo4j/migrations/schema_migration.rb
```diff
@@ -0,0 +1,11 @@
+module Neo4j
+  module Migrations
+    class SchemaMigration
+      include Neo4j::ActiveNode
+
+      property :migration_id, type: String, constraint: :unique
```
cheerfulstoic: Perhaps this could be the `id_property` because it's a natural identifier and then we don't need to worry about `uuid` (which would just mean more data returned from the DB anyway)

lib/neo4j/tasks/migration.rake
```diff
@@ -21,4 +21,44 @@ namespace :neo4j do
       migration.migrate
     end
   end
+
+  desc 'A shortcut for neo4j::migrate::all'
+  task :migrate do
+    Rake::Task['neo4j::migrate::all'].invoke
```
cheerfulstoic: I thought you only did a single `:` between namespaces in rake.  Can you do both?

spec/e2e/migrations_spec.rb
```diff
@@ -0,0 +1,113 @@
+describe Neo4j::Migrations::Runner do
```
cheerfulstoic: One trick that I've seen is that you can surround your `RSpec` with the module(s) in question to reduce the amount of repetition of those modules.  For example here you could do:

```ruby
module Neo4j
  module Migrations
    describe Runner do
      # etc...
      before do
        # etc...
        SchemaMigration.delete_all
      end
      # etc...
    end
  end
end
```

spec/e2e/migrations_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+  before(:all) { ENV['silenced'] = 'true' }
+  after(:all)  { ENV['silenced'] = nil }
```
cheerfulstoic: It's not a big deal for this I think, but we might think about having a `let_env_variable` helper which would work something like `let_env_variable('silenced', true)` and it would store and restore the variable back to it's original state after the spec is done (similar to `let_config`)

spec/unit/migrations/helpers_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+    Book.create!(name: 'Book1')
+    Book.create!(name: 'Book2')
+    Book.create!(name: 'Book3')
+  end
+
+  describe '#rename_property' do
+    it 'renames a property' do
+      rename_property :Book, :name, :title
+      expect(Book.all(:n).pluck('n.title')).to include('Book1', 'Book2', 'Book3')
```
cheerfulstoic: We should test whatever edge-case behavior we decide on regarding my comment above about what happens to nodes which already have the new property set.

spec/unit/migrations/helpers_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+  describe '#drop_nodes' do
+    it 'drops all nodes given a label' do
+      drop_nodes :Book
+      expect(Book.all.count).to eq(0)
+    end
+  end
+
+  describe '#add_labels' do
+    it 'adds labels to a node' do
+      add_labels :Book, :Item, :Readable
```
cheerfulstoic: I think that there should actually be a second argument which is an `Array`.  Looking at this at first I thought you were adding three labels, and then I remembered the method definition.  Like so:

```ruby
add_labels :Book, [:Item, :Readable]
add_label :Book, :Item
```

spec/unit/migrations/helpers_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    it 'executes plan cypher query with parameters' do
+      expect do
+        execute 'MATCH (b:`Book`) WHERE b.name = {book_name} DELETE b', book_name: Book.first.name
+      end.to change { Book.count }.by(-1)
+    end
+  end
+
+  describe '#remove_index' do
+    it 'removes an index from a property' do
+      remove_index :Book, :author_name
```
cheerfulstoic: This could probably use the `Label` class from code to get the list of indexes and you could see if the `author_name` index exists

lib/neo4j/migrations/helpers.rb
cheerfulstoic: I'm down with that.  @subvertallchris ?

lib/neo4j/migrations/runner.rb
cheerfulstoic: Ah, I didn't know that.  Meh, new migration, new major version, new behavior ;)  I would just make sure that it goes into the `CHANGELOG` under `Changes`

lib/neo4j/migrations/migration_file.rb
cheerfulstoic: Yup!  We use the modules `PerThreadRegistry`, `Notifications`, `Concern`, and a few core extensions, so it's pretty ingrained into the gem by now ;)  But we try to require only the sections that we need

cheerfulstoic: Yup!  We use the modules `PerThreadRegistry`, `Notifications`, `Inflector`, `Concern`, and a few core extensions, so it's pretty ingrained into the gem by now ;)  But we try to require only the sections that we need

cheerfulstoic: Yup!  We use the modules `PerThreadRegistry`, `Notifications`, `Inflector`, `Concern`, and a few core extensions, so it's pretty ingrained into the gem by now ;)  But we try to require only the sections that we need

lib/neo4j/migrations/helpers.rb
ProGM: first done.

About DETACH DELETE, what should I do?

lib/neo4j/migrations/helpers.rb
ProGM: done 👍

lib/neo4j/migrations/migration_file.rb
ProGM: done 👍

cheerfulstoic:
I had a number of comments, but it looks really great overall!  Really excited to have this coming up, thanks!

A couple of things to make sure we address:

 * `CHANGELOG` should lay out anything that changed / was added
 * We should have docs.  I don't think that migrations are in readthedocs yet, but this seems like a good opportunity.  Probably this should be it's own section.

lib/neo4j/migrations/helpers.rb
cheerfulstoic: We can't use `DETACH DELETE` as we still support Neo4j 2.1 / 2.2.

spec/unit/migrations/helpers_spec.rb
ProGM: Done!

lib/neo4j/tasks/migration.rake
ProGM: No, it's a typo 👎 fixed!

lib/neo4j/migrations/runner.rb
```diff
@@ -0,0 +1,106 @@
+module Neo4j
+  module Migrations
+    class Runner
+      STATUS_TABLE_FORMAT = '%-10s %-20s %s'.freeze
+      SEPARATOR = '--------------------------------------------------'.freeze
+      FILE_MISSING = '**** file missing ****'.freeze
+      STATUS_TABLE_HEADER = ['Status'.freeze, 'Migration ID'.freeze, 'Migration Name'.freeze].freeze
+      UP_MESSAGE = 'up'.freeze
+      DOWN_MESSAGE = 'down'.freeze
```
ProGM: haha 😆 😆

lib/neo4j/migrations/migration_file.rb
ProGM: Triple comment, good work github :metal:

lib/neo4j/migrations/helpers.rb
ProGM: Uh, I see 👍

lib/neo4j/migrations/helpers.rb
ProGM: Done 👍

lib/neo4j/migrations/runner.rb
ProGM: (done)

lib/neo4j/migrations/runner.rb
ProGM: done (refactored)

lib/neo4j/migrations/runner.rb
ProGM: done!

spec/unit/migrations/helpers_spec.rb
ProGM: done!

spec/e2e/migrations_spec.rb
ProGM: done!

spec/e2e/migrations_spec.rb
ProGM: Done!

ProGM:
@cheerfulstoic @subvertallchris Ok, some updates: I fixed all issues you wrote, doing some refactoring.

I also added the `remove_property` helper and updated the CHANGELOG.

Please take a look, later I'll take care of docs 😭

subvertallchris:
Very cool

ProGM:
Ach, jruby broken 😞

cheerfulstoic:
This all looks amazing!  Thanks again!

Once docs are there and jruby is working, feel free to merge away!

subvertallchris:
...and when there's a test for the transactional stuff we talked about.

ProGM:
@cheerfulstoic @subvertallchris I did a bit of refactor to improve error messages.

Moreover I (hopefully) fixed jruby and added some tests for transactions.

Only docs are missing now. However, if you have time, take a look to this last commit, I did a few changes that may require a revision. 😜

cheerfulstoic:
I believe that `Transaction.run` takes a boolean argument to say if you want to have a transaction (strange, I know, but it's pretty much for this sort of thing).  

Also, completely optional, but I just wanted to make sure that this thought wasn't just it my head: we might also implement [`say`](http://apidock.com/rails/v3.0.9/ActiveRecord/Migration/say/class) and [`say_with_time`](http://apidock.com/rails/ActiveRecord/Migration/say_with_time)

Thanks!

ProGM:
Ok, I also added `say` and `say_with_time` (they are more or less copy-pasted from ActiveRecord, since it works as-is)

ProGM:
@cheerfulstoic @subvertallchris I added also `rename_label` helper and I created a `Migrations` section in the docs.

What do you think? :) (please check the grammar 😆 )

docs/Migrations.rst
```diff
@@ -0,0 +1,260 @@
+Migrations
+==================
+
+Neo4j.rb provides an ``ActiveRecord``-like migration framework and a set of helper methods to manipulate both database schema and data.
```
cheerfulstoic: For people that are new to Neo4j they may need some context since the use of our migrations isn't quite like `ActiveRecord`'s.  Maybe saying something like "Neo4j does not have a set schema like relational databases, but sometimes changes to the schema and the data are required.  To help with this, Neo4j.rb provides an ``ActiveRecord``-like" etc...

docs/Migrations.rst
```diff
@@ -0,0 +1,260 @@
...
+
+.. code-block:: ruby
+
+  class SomeMigration < Neo4j::Migrations::Base
+    disable_transactions!
+
+    ...
+  end
+
+Neo4j::Migrations::Helpers
```
cheerfulstoic: Perhaps a title like "Migration Helpers" so that people don't think they need to know about the `Neo4j::Migrations::Helpers" class.

Also, perhaps the `RenameUserNameToFirstName` migration example should use at least one helper so that people can make the association about where they would call the methods below.

docs/Migrations.rst
```diff
@@ -0,0 +1,260 @@
...
+
+
+#query
+---------------
+
+An alias for ``Neo4j::Session.query``. You can use it as root for the query builder:
+
+.. code-block:: ruby
+
+  query.match(:n).where(name: 'John').delete(:n)
```
cheerfulstoic: To actually get this to execute they would need to end this with `exec` or some `Enumerable` method (`to_a`/`each`/`select`/etc...)

docs/Migrations.rst
```diff
@@ -0,0 +1,260 @@
...
+
+:Ruby:
+  .. code-block:: ruby
+
+    remove_property(:User, :money)
+
+:Cypher:
+  .. code-block:: cypher
+
+    MATCH (n:`User`) REMOVE n.money
```
cheerfulstoic: In thinking about this I went back and forth, but I think we don't need the `Cypher`.  If we ever change how the helper executes then we'd need to remember to update the docs.

This makes me realize, though, that we **should** make sure we're displaying the `Cypher` which is being executed during a migration run.  Do you know if that's currently happening?  If not, it should be possible to just set the log level (see the `railtie.rb`)

docs/Migrations.rst
```diff
@@ -0,0 +1,260 @@
...
+#drop_constraint
+------
+
+Drops an unique constraint on a given label attribute.
+
+**Warning** it would fail unless you define ``disable_transactions!`` in your migration file.
+
+.. code-block:: ruby
+
+  drop_constraint(:User, :name)
```
cheerfulstoic: Will this fail if you don't define `disable_transactions!` if you're only making schema changes?

cheerfulstoic:
Feedbacked!

docs/Migrations.rst
ProGM: Currently... yes. And maybe it's not a great idea, though.
Check:
https://github.com/neo4jrb/neo4j/pull/1197/files#diff-ab636d4950487b612dfbe9fde69e387bR48

It's better to leave the neo4j error pass through

docs/Migrations.rst
ProGM: 👍 fixed

docs/Migrations.rst
ProGM: Sounds good, fixed

docs/Migrations.rst
ProGM: 👍

docs/Migrations.rst
ProGM: Uhm, right. I have to check for output.

ProGM:
@cheerfulstoic @subvertallchris travis is 👍 , what do you think?

To merge or not to merge? 💀

subvertallchris:
This is really fantastic, go for it!

cheerfulstoic:
❗

ProGM:
👍

PR #1202, opened 2016-05-18
Docs improvements

ProGM:
Checklist:
* [x] Improve and fix docs for Migrations
* [x] Add `docs/api` to `.gitignore`
* [x] Talk about `id_property`
* [ ] Move everything is missing from the wiki to the docs

Pings:
@cheerfulstoic
@subvertallchris


cheerfulstoic:
:+1: so far, thanks!

ProGM:
@cheerfulstoic @subvertallchris 
Two things:

* what about `.gitignore`-ing the `docs/api` folder? I discussed about it with @cheerfulstoic . Every time you run rake docs, it generates a lot of changes in that folder, and it's pretty strange to commit them in the repo, since they are auto-generated.

* I copied (with a few changes) the wiki about UniqueIDs on the docs: https://github.com/neo4jrb/neo4j/pull/1202/commits/83e8356935cf649ab1b904866570c9d7ed713cbf .
I think that the last two sections ("Adding IDs to Existing Data" and "Working with Legacy Schemas") should be rewritten, or maybe removed. What do you think?
In depth:
**"Working with Legacy Schemas"** maybe should be rewritten so that it uses migrations instead of pure cypher queries
**"Adding IDs to Existing Data"** uses legacy migrations. What we want to do with them? Maybe drop and create a helper in new migrations? Leave them as-is?

docs/UniqueIDs.rst
```diff
@@ -0,0 +1,92 @@
+Unique IDs
+==========
+
+The database generates unique IDs and they are accessible from all nodes and relationships using the ``neo_id`` method. These keys are somewhat volatile and may be reused or change throughout a database's lifetime, so they are unsafe to use within an application.
+
+To work around this, you can define which key should act as primary key on ``Neo4j::ActiveNode`` classes instead of using the internal Neo4j ids. By default, ActiveNode will generate a unique ID using ``SecureRandom::uuid``. The instance methods ``id`` and ``uuid`` will both point to this.
```
cheerfulstoic: "To work around this, you can define" makes it sound like it's optional.  You mention "by default" below, but I think this should be clear outside of the gate that some `id_property` is required for every model, and if you don't define one, one will be defined for you.

Also, I don't think that `uuid` will point to the `id_property` when you define an `id_property` other than `uuid`, right?  I think it's just `id` / `id=`

docs/UniqueIDs.rst
```diff
@@ -0,0 +1,92 @@
...
+
+    class Person
+      include Neo4j::ActiveNode
+      id_property :neo_id
+    end
+
+A note regarding constraints
+----------------------------
+
+By default, a uniqueness constraint will be set for all ID properties. To disable this, you can call ``id_property`` with ``constraint: false`` in your third param.
```
cheerfulstoic: It's not a third param but another key in the second param's `Hash`

ProGM: Seems like the wiki was pretty incorrect 👍 I didn't noticed it

cheerfulstoic:
I'm totally about `.gitignoring` the `docs/api`.  I think I didn't think about it very closely before, but those `rst` files also often show up when I'm searching the code and it's annoying.

One suggestion that I've heard, by the way is to make once sentence per-line, but not have spaces between lines except for when you're making a new paragraph.  That way diffs become a lot cleaner.  If you and @subvertallchris also think this would be a good idea we might start with this PR.

Agreed about re-writing "Working with Legacy Schemas" to use new migrations functionality.

I would love to see a `add_id_property` helper for the existing migrations to keep things consistent.  @subvertallchris ?

Thanks!

docs/UniqueIDs.rst
ProGM: Uhm, let me try, but I think you're right. However, I just copied from the wiki.

cheerfulstoic: Thanks! ;)  It doesn't surprised me too much if the wiki is out-of-date because I don't think it's been updated since v2-v3 of the `neo4j` gem.  It will be good to have it all in one place so that it will be more in our face

ProGM:
@cheerfulstoic @subvertallchris 
I completed the `UniqueIDs` section and improved a few other things about docs.

Please check and tell me if something else needs to be documented/improved!

docs/UniqueIDs.rst
```diff
@@ -3,9 +3,9 @@ Unique IDs
 
 The database generates unique IDs and they are accessible from all nodes and relationships using the ``neo_id`` method. These keys are somewhat volatile and may be reused or change throughout a database's lifetime, so they are unsafe to use within an application.
 
-Neo4j.rb requires yout to define which key should act as primary key on ``Neo4j::ActiveNode`` classes instead of using the internal Neo4j ids. By default, ActiveNode will generate a unique ID using ``SecureRandom::uuid`` saving it in a ``uuid`` property. The instance method ``id`` will also point to this.
```
cheerfulstoic: yout

cheerfulstoic:
This all seems great to me!  Should we perhaps merge just to get some of this good stuff into place?  I'm particularly excited about the `.gitignore` (It's been bothering me to see `rst` files in my search results ever since we talked about it ;)

ProGM:
For me it's ok to merge! @subvertallchris ?

subvertallchris:
Very cool!

ProGM:
🚀

PR #1204, opened 2016-05-23
Some improvements on Migrations

ProGM:
This pull introduces/changes:
 * `add_index` and `add_constraint` helpers
 * A `populate_id_property` helper, as a replacement for the legacy `AddIdProperty` task
 * Splits migration helpers in submodules

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/lib/neo4j/migration.rb
+++ b/lib/neo4j/migration.rb
@@ -23,6 +23,8 @@ module Neo4j
     end
 
     class AddIdProperty < Neo4j::Migration
+      include Neo4j::Migrations::Helpers::IdProperty
+
       attr_reader :models_filename
 
       def initialize(path = default_path)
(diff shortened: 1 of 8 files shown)


lib/neo4j/migration.rb
```diff
@@ -23,24 +23,30 @@ def joined_path(path)
...
+      include Neo4j::Migrations::Helpers::IdProperty
+
       attr_reader :models_filename
 
       def initialize(path = default_path)
         @models_filename = File.join(joined_path(path), 'add_id_property.yml')
       end
 
       def migrate
+        ActiveSupport::Deprecation.warn '`AddIdProperty` task is deprecated and may be removed from future releases. '\
```
cheerfulstoic: I was actually thinking about putting the deprecation into 7.0.x and then removing it in 8.0.  What do you think?

cheerfulstoic:
LGTM!  @subvertallchris ?

lib/neo4j/migration.rb
ProGM: Yep, I though the same. I added the deprecation warning here, then we can remove the entire (old) `Neo4j::Migration` namespace from 8.0+ , if you agree :)

Issue #1205, opened 2016-05-23
scopes are not inherited


veetow:
Given two `ActiveNode` classes, `Parent` and `Child`, if `Parent` includes a `scope`, that `scope` is not available in `Child`.

PR coming with failing spec to illustrate the issue.

### Runtime information:

Neo4j database version: 2.3.2
`neo4j` gem version: 6.1.5
`neo4j-core` gem version: 6.1.1

veetow:
I've verified that this is also broken in the `7.1.x` and `master` branches.

cheerfulstoic:
Great catch, thanks (and thanks for the failing spec)!

I'm headed to bed, but I'll put this on my list to look at tomorrow

cheerfulstoic:
I've got a fix!  Give 6.1.11 a try and let me know if it works for you.  If it does I'll move the fix forward to the 7.0.x series.

Thanks again for reporting this!

veetow:
Thanks @cheerfulstoic.  6.1.11 does work for me.

cheerfulstoic:
Awesome!  Also released as 7.0.7

PR #1206, opened 2016-05-23
add failing spec for inherited scope issue

veetow:
Illustrates #1205

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/spec/e2e/scope_spec.rb
+++ b/spec/e2e/scope_spec.rb
@@ -8,7 +8,10 @@ describe 'Neo4j::NodeMixin::Scope' do
       property :name
       property :score
       property :level_num
+      property :date_of_death
       has_many :out, :friends, type: nil, model_class: 'Person'
+
+      scope :only_living, ->(identifier) { where("#{identifier}.date_of_death IS NULL") }
     end
   end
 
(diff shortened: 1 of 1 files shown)


spec/e2e/scope_spec.rb
```diff
@@ -8,7 +8,10 @@
       property :name
       property :score
       property :level_num
+      property :date_of_death
       has_many :out, :friends, type: nil, model_class: 'Person'
+
+      scope :only_living, ->(identifier) { where("#{identifier}.date_of_death IS NULL") }
```
cheerfulstoic: You could skip the identifier and do `-> { where(date_of_death: nil) }`

spec/e2e/scope_spec.rb
```diff
@@ -26,6 +29,17 @@
...
 
+  describe 'Inherited scope' do
+    before { stub_named_class('Mutant', Person) }
+
+    let(:alive) { Mutant.create name: 'aa' }
+    let(:dead)  { Mutant.create name: 'bb', date_of_death: 'yesterday' }
+
+    it 'has the scope of the parent class' do
+      expect(Mutant.scope?(:only_living)).to be true
+      expect(Mutant.as(:m).only_living(:m).to_a).to eq([alive])
```
cheerfulstoic: There's no need for the node variable, I think.  You could just do `Mutant.all.only_living.to_a`.  You might even be able to drop the `to_a` because I think when it compares `eq` it will enumerate it.

cheerfulstoic:
Just a couple of notes, but this looks great, thanks!

cheerfulstoic:
Merging so that I can look at this

PR #1207, opened 2016-05-25
Implementing `having_rel` and `not_having_rel`

ProGM:
This pull introduces/changes:
 * `having_rel`, that returns nodes having a certain relation
 * `not_having_rel`, that returns nodes not having a certain relation

Both accept an `association_name` and an optional hash of association properties.

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/lib/neo4j/active_node/query/query_proxy.rb
+++ b/lib/neo4j/active_node/query/query_proxy.rb
@@ -191,6 +191,36 @@ module Neo4j
           instance_eval(&block).query.proxy_as(self.model, identity)
         end
 
+        # Matches all nodes having at least a relation
+        #
+        # @example Load all people having a friend
+        #   Person.all.having_rel(:friends).to_a # => Returns a list of `Person`
+        #
+        # @example Load all people having a best friend
+        #   Person.all.having_rel(:friends, best: true).to_a # => Returns a list of `Person`
+        #
+        # @return [QueryProxy] A new QueryProxy
+        def having_rel(association_name, rel_properties = {})
+          association = model.associations[association_name]
+          fail "No such association #{association_name}" unless association
+          where("(#{identity})#{association.arrow_cypher(nil, rel_properties)}()")
+        end
+
+        # Matches all nodes not having a certain relation
+        #
+        # @example Load all people not having friends
+        #   Person.all.not_having_rel(:friends).to_a # => Returns a list of `Person`
+        #
+        # @example Load all people not having best friends
+        #   Person.all.not_having_rel(:friends, best: true).to_a # => Returns a list of `Person`
+        #
+        # @return [QueryProxy] A new QueryProxy
+        def not_having_rel(association_name, rel_properties = {})
+          association = model.associations[association_name]
+          fail "No such association #{association_name}" unless association
+          where_not("(#{identity})#{association.arrow_cypher(nil, rel_properties)}()")
+        end
+
         def [](index)
           # TODO: Maybe for this and other methods, use array if already loaded, otherwise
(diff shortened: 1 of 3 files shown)


lib/neo4j/active_node/query/query_proxy.rb
```diff
@@ -191,6 +191,36 @@ def branch(&block)
...
+        # @example Load all people having a friend
+        #   Person.all.having_rel(:friends).to_a # => Returns a list of `Person`
+        #
+        # @example Load all people having a best friend
+        #   Person.all.having_rel(:friends, best: true).to_a # => Returns a list of `Person`
+        #
+        # @return [QueryProxy] A new QueryProxy
+        def having_rel(association_name, rel_properties = {})
+          association = model.associations[association_name]
+          fail "No such association #{association_name}" unless association
```
cheerfulstoic: This feels like we should have a `association` method which takes a `String` or `Symbol` as an argument and returns an association, but also would `fail` if the association doesn't exist.

Also, using `ArgumentError` would probably be appropriate

spec/e2e/query_proxy_methods_spec.rb
```diff
@@ -663,6 +663,39 @@ def destroy_called
...
+    describe 'not_having_rel' do
+      it 'returns students having at least a lesson' do
+        expect(Student.all.not_having_rel(:lessons)).to contain_exactly(@bill)
+      end
+
+      it 'returns students having at least a lesson with no absences' do
+        expect(Student.all.not_having_rel(:lessons, absence_count: 0)).to contain_exactly(@bill, @frank)
+      end
+    end
+  end
```
cheerfulstoic: For both methods it would be good to test cases that return empty sets.  Probably would be good to test both:
 * where the association has no nodes
 * where a property is specified which eliminates all nodes.

I also didn't know about `contain_exactly`.  I've used `match_array` before, but I like the semanticness of this.

cheerfulstoic:
I like this a lot!  Thanks so much for knocking it out!

One case that I think could be valuable, though I'm not sure quite how to do it: what if somebody wants to filter the relationships based on something other than exact matching (like `>`/`<` or partial string matching)?  I was thinking that users could provide a string, though we'd need to give them a way to specify the variable name.  What do you think?

ProGM:
@cheerfulstoic Uhm. I think that if the user wants to match by `>`/`<` and similar, we have to change a bit the query (and so, we can't use `WHERE (n)-[:relname]-()` and `WHERE NOT (n)-[:relname]-()`).

Instead, we have to switch to an optional match, maybe. Something like:

```cypher
MATCH (n:`Some`) OPTIONAL MATCH (n)-[r:some_other]->() WHERE r IS NOT NULL AND r.some_property > {some_value}
```

However, I'm not sure of it. Thoughts?

cheerfulstoic:
Ah, right, good point because you can't define a variable for the rel in a `WHERE`.  Then yeah, I say we can leave it as you have it for now because matching rels in `WHERE` can also only match exact, so I think it makes sense.

@subvertallchris Any thoughts / objections / praise?

cheerfulstoic:
We discussed this with @subvertallchris some and he was concerned about raw Cypher inside the method.  I would say, though, that the API is good and we can always change the implementation later if we want.  So I say merge.  In somewhat of the the spirit of the [C4 specification](http://rfc.zeromq.org/spec:42/) if anybody doesn't like it, they can make another PR ;)

ProGM:
For me it's ok 👍

cheerfulstoic:
Done! ;)

PR #1215, opened 2016-06-03
ActiveNode inclusion respects existing id_property_name

subvertallchris:
The problem:

Your id property name is not going to be `uuid`. You do not want to enable `eager_load`. Every time your app starts, it checks for that `uuid` constraint and sets it, even if you don't want it there, then it checks for your custom id property constraint. You find this annoying.

The really simple solution:

``` ruby
class MyModel
  def self.id_property_name
    :cool_id
  end
  include Neo4j::ActiveNode
```

This PR changes the `id_property_name` method to respect an existing value of that method. If you set it before including the module, it will use that value for that class. I'm not sure that this is the absolute best possible solution but it sure is simple, which is rather nice.

Pings:
@cheerfulstoic
@ProGM

### DIFF
--- a/lib/neo4j/active_node.rb
+++ b/lib/neo4j/active_node.rb
@@ -88,7 +88,7 @@ module Neo4j
 
       Neo4j::Session.on_next_session_available do |_|
         next if manual_id_property?
-        id_property :uuid, auto: :uuid unless self.id_property?
+        id_property id_property_name || :uuid, auto: :uuid unless self.id_property?
 
         name = Neo4j::Config[:id_property]
         type = Neo4j::Config[:id_property_type]
--- a/lib/neo4j/active_node/id_property.rb
+++ b/lib/neo4j/active_node/id_property.rb
@@ -162,7 +162,7 @@ module Neo4j::ActiveNode
       end
 
       def id_property_name
-        id_property_info[:name]
+        defined?(super) ? super : id_property_info[:name]
       end
 
       def manual_id_property?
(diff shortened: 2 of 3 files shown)


cheerfulstoic:
This seems fine to me, though you should maybe check out the differences to these parts of the code in the `new_cypher_api` branch.  I've done away with `on_next_session_available`, for example.

ProGM:
Seems fine!

For the future, maybe we should think about changing something in this "auto-creating index on include" mechanism, that could create this strange tricks.

cheerfulstoic:
I explained it in the other issue (going back-and-forth here ;) but I think `new_cypher_api` takes care of it by auto-creating the index when the `id_property` info is sought for whatever reason (that should be the last possible moment that it could have changed).  I also changed it to make schema changes in a thread with a different transaction so that we don't have schema/data conflict issues in general (unless somebody runs a manual query)

kbighorse:
FWIW, Ruby 2.0 has been deprecated as of February 2016, can any/all of the failing checks be removed?
https://www.ruby-lang.org/en/news/2016/02/24/support-plan-of-ruby-2-0-0-and-2-1/

cheerfulstoic:
Actually we have dropped checks for Ruby 2.0 in more recent releases:

https://travis-ci.org/neo4jrb/neo4j/builds/167095097

@subvertallchris What's the deal with this PR?

subvertallchris:
Now that adding constraints is a manual, deliberate process, I think this should be fine to close.

PR #1220, opened 2016-06-15
Fix label to model mapping

### LINKED ISSUE #1219: Class loading problems with subclasses & modules
I have the following ActiveNode definitions:

```
class URL
  include Neo4j::ActiveNode

  property :value
end

class Event
  include Neo4j::ActiveNode
end

module DataSource
  class URL < ::URL
    self.mapped_label_name = "DataSource"

    has_one :out, :recipient, type: "DATA_SOURCE_FOR", model_class: false
  end
end

module DataSource
  class Event < DataSource::URL
  end
end
```

When I make the following query, the class I get back is an `::Event`, not a `DataSource::Event`.

```
data_source = DataSource::URL.find_by(value: "/some/value")
event = data_source.recipient    # method_missing error because data_source is an ::Event, not DataSource::Event
```

This is happening because `Neo4j::ActiveNode::Labels::model_for_labels([:URL, :DataSource, :Event]) `is returning `::Event`.

Two me, there are two actual bugs here:

1. `Neo4j::ActiveNode::Labels::model_for_labels` isn't returning the correct class
2. `DataSource::URL.find_by` is returning something other than a `DataSource::URL`

I am going to create a test for this (once I can figure out how the spec tests work) and probably try to fix issue 1 today with a PR. Issue 2 seems like a more systemic problem with type checking, and possibly is related to #1213.

bloomdido:
Fixes #1219 

This pull introduces/changes:
 * Added spec tests to highlight mismapped node models from labels
 * Fix mapping by switching from `Enumerable::max` to `Enumerable::max_by`

Pings:
@cheerfulstoic
@subvertallchris


bloomdido:
After this change, I now receive the following rspec failure:

```
     migration tasks AddIdProperty class respects the id_property declared on the model
     Failure/Error: expect(user).not_to respond_to(:uuid)
       expected #<Song uuid: nil, name: nil> not to respond to :uuid
     # ./spec/e2e/migration_spec.rb:99:in `block (3 levels) in <top (required)>'
```

I am unsure yet as to why this is failing, but I did find a typo in neo4j/spec/e2e/migration_spec.rb:15. Pretty sure `rel_class: :oSecondRelClass` should be `rel_class: :SecondRelClass`.

bloomdido:
As expected, this commit fails the CI checks pretty spectacularly, but so does `master` right now. I am pretty certain the fix itself is the intended behavior and should probably be considered a patch for older branches/releases.

bloomdido:
I just found a problem with this patch. It's going to need some more work because if there are multiple classes that match the labels with the same match score, it still fails. Working on fix now...

bloomdido:
I was wrong. What I am seeing is the SECOND bug I identified in #1219. `Event.all` is matching any node with the label `:Event`, but not filtering out those that aren't actually `Event` classes. On it...

cheerfulstoic:
Thanks a lot for looking into this!

I'm going to be focusing in the few days on getting `master` to a stable place.  That will probably involve merging the `new_cypher_api` branch so that we're all on the same page.  I'll update when that's in order

bloomdido:
FTR, I tried merging the f98592cbac4d7bff6cc38fc858798f0df3cfabd6 commit into the 7.0.x branch and there were no rspec failures.

cheerfulstoic:
Sorry for the delay.  Still working on getting `master` in order, but this is a very straightforward fix so I'm going to merge it.  Will cherry-pick to 7.0.x shortly

cheerfulstoic:
Released as 7.0.12!

Issue #1208, opened 2016-05-25
Bug with scopes/class methods and `searchkick` gem


ProGM:
I was following the `searchkick` manual.
https://github.com/ankane/searchkick#indexing

As written, I defined a `search_import` scope to define eager loaded values for custom indexed data.

I tried both defining a scope (as stated in the guide) and a class method. Both raises an error when I run `reindex`:

```ruby
class Article
  scope :search_import, -> { all.with_associations(:published_version) }
end
```

```ruby
class Article
  def self.search_import
    all.with_associations(:published_version)
  end
end
```

They both raises `NoMethodError: undefined method 'identity' for #<#<Class:0x007f961a2a97b8>:0x007f961a2a8520>`

### Runtime information:

Neo4j database version:
`neo4j` gem version: 7.0.6+ (master HEAD)
`neo4j-core` gem version: 6

cheerfulstoic:
I think I've fixed it.  Just released patch versions: 6.0.9, 6.1.12, and 7.0.8.  Give that a try! ;)

ProGM:
Wow, it was really a code bug!

I tested the patch. Now I got no error, but the eager load is not working...

```ruby
class Article
  include Neo4j::ActiveNode
  include Neo4j::Timestamps
  include Searchkick::Model
  searchkick
  def self.search_import
    all.with_associations(:published_version)
  end

  def search_data
    attributes.merge(
      'published' => !published_version.nil?
    )
  end
end
```

```bash
irb(main):016:0> Article.reindex
 CYPHER 19ms MATCH (n:`Article`) RETURN n ORDER BY n.uuid LIMIT {limit_1000} | {:limit_1000=>1000}
 Article#published_version 4ms MATCH (article516) WHERE (ID(article516) = {ID_article516}) MATCH (article516)-[rel1:`published_version`]->(result_published_version:`Version`:`Quotable`) RETURN result_published_version | {:ID_article516=>516}
 Article#published_version 2ms MATCH (article101) WHERE (ID(article101) = {ID_article101}) MATCH (article101)-[rel1:`published_version`]->(result_published_version:`Version`:`Quotable`) RETURN result_published_version | {:ID_article101=>101}
 Article#published_version 6ms MATCH (article74) WHERE (ID(article74) = {ID_article74}) MATCH (article74)-[rel1:`published_version`]->(result_published_version:`Version`:`Quotable`) RETURN result_published_version | {:ID_article74=>74}
[...]

PR #1226, opened 2016-06-27
Refactoring and fixing scopes.

### LINKED ISSUE #1224: Cannot use ? params in scope where clause

    class Person
        include Neo4j::ActiveNode
        property :name
        has_many :out, :friends, type: :friend, model_class: 'Person'
        scope :has_friends ->(count) { where("SIZE((#{identity}-[:friend]-()) = ?", count)  }
    end

    Person.all.has_friends(3)

### Runtime information:

raises the error: ```gems/neo4j-c2680108df1e/lib/neo4j/active_node/scope.rb:114:in `where': wrong number of arguments (given 2, expected 0..1) (ArgumentError)```

Looks like the problem is  at this line: 

https://github.com/neo4jrb/neo4j/blob/39456bf782f5a4fdb58f2108110e64837750c1ac/lib/neo4j/active_node/scope.rb#L109

ProGM:
Fixes #1224 #1143 #1208 and hopefully all issues related to scopes 

This pull introduces/changes:
 * Support for associations methods in scopes
 * Support for _any_ `QueryProxy` / `AssociationProxy` method inside scopes
 * Support for ? clause in scopes
 * Fix reverse inheritance with parent scopes (Scope defined in parents were overwriting child ones!!!)

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/lib/neo4j/active_node/scope.rb
+++ b/lib/neo4j/active_node/scope.rb
@@ -105,12 +105,11 @@ module Neo4j::ActiveNode
       end
 
       Neo4j::ActiveNode::Query::QueryProxy::METHODS.each do |method|
-        module_eval(%{
-            def #{method}(*args)
-              @target.all.scoping do
-                query_proxy_or_target.#{method}(*args)
-              end
-            end}, __FILE__, __LINE__)
+        define_method(method) do |*args|
+          @target.all.scoping do
+            query_proxy_or_target.public_send(method, *args)
+          end
+        end
       end
 
       def method_missing(name, *params, &block)
--- a/lib/neo4j/active_node/scope.rb
+++ b/lib/neo4j/active_node/scope.rb
@@ -64,7 +64,7 @@ module Neo4j::ActiveNode
       end
 
       def full_scopes
-        scopes.merge(self.superclass.respond_to?(:scopes) ? self.superclass.scopes : {})
+        self.superclass.respond_to?(:scopes) ? self.superclass.scopes.merge(scopes) : scopes
       end
 
       def _call_scope_context(eval_context, query_params, proc)
@@ -101,16 +101,30 @@ module Neo4j::ActiveNode
       end
 
       def identity
-        @query_proxy ? @query_proxy.identity : :n
+        query_proxy_or_target.identity
       end
 
       Neo4j::ActiveNode::Query::QueryProxy::METHODS.each do |method|
         module_eval(%{
-            def #{method}(params={})
+            def #{method}(*args)
               @target.all.scoping do
-                (@query_proxy || @target).#{method}(params)
+                query_proxy_or_target.#{method}(*args)
               end
             end}, __FILE__, __LINE__)
+      end
+
+      def method_missing(name, *params, &block)
+        if query_proxy_or_target.respond_to?(name)
+          query_proxy_or_target.public_send(name, *params, &block)
+        else
+          super
+        end
+      end
+
+      private
+
+      def query_proxy_or_target
+        @query_proxy_or_target ||= @query_proxy || @target
       end
     end
 
--- a/spec/e2e/scope_spec.rb
+++ b/spec/e2e/scope_spec.rb
@@ -143,5 +143,32 @@ describe 'Neo4j::NodeMixin::Scope' do
     it_behaves_like 'scopable model'
     it_behaves_like 'chained scopable model'
   end
-  # end
+
+  describe 'Person.scope :top_friends, -> { friends.where(score: 42) }' do
+    before(:each) do
+      Person.scope :top_friends, -> { friends.where(score: 42) }
+    end
+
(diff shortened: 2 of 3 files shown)


lib/neo4j/active_node/scope.rb
```diff
@@ -101,17 +101,31 @@ def initialize(target, query_proxy)
...
       Neo4j::ActiveNode::Query::QueryProxy::METHODS.each do |method|
         module_eval(%{
-            def #{method}(params={})
+            def #{method}(*args)
               @target.all.scoping do
-                (@query_proxy || @target).#{method}(params)
+                query_proxy_or_target.#{method}(*args)
               end
             end}, __FILE__, __LINE__)
       end
```
cheerfulstoic: This might be a nice opportunity to transform this to `define_method` or `define_class_method` (whatever is appropriate).  Don't know how hard that would be

cheerfulstoic:
This seems really great, thanks!

I'm going to look over it more in a bit, but I definitely look forward to scopes working better.

cheerfulstoic:
Just finished looking it over.  Seems great!

You can decide if you want to do the refactor to a `define_method`.  Either way, I'm happy to have it merged when you think it's ready!

ProGM:
@cheerfulstoic For me no problem to refactor that part ;)

However, I admit my ignorance: what is the advantage by using `define_method` over a `module_eval` in this case (performances, style, whatever)?

cheerfulstoic:
It should be faster generally:

https://tenderlovemaking.com/2013/03/03/dynamic_method_definitions.html

I've also found that it's nicer to work with.  Previously the gem has a whole series of `eval`s / `module_eval`s and refactoring to `define_method` made things much better (this was in `has_n.rb`)

ProGM:
@cheerfulstoic Uhm, I see... I have a couple of doubts about performances. Surely `define_method` is faster to define methods, instead of `module_eval`. However, in runtime, I think that calling the defined method it's faster using module_eval.

An example:
```ruby
      Neo4j::ActiveNode::Query::QueryProxy::METHODS.each do |method|
        module_eval(%{
            def #{method}(*args)
              @target.all.scoping do
                query_proxy_or_target.#{method}(*args)
              end
            end}, __FILE__, __LINE__)
      end
```

Should be replaced with:

```ruby
      Neo4j::ActiveNode::Query::QueryProxy::METHODS.each do |method|
        define_method method do |*args|
          @target.all.scoping do
             query_proxy_or_target.public_send(method, *args)
           end
        end
      end
```

However, the first method calls `where`, `limit` and all other query methods directly. Instead define_method calls a `public_send` or `send`.

In general calling `public_send` is times slower than calling methods directly, isn't it?

However, I'm not sure of it, and probably it's pretty trivial. However, I can make a few benchmarks, if you want ;)

cheerfulstoic:
Ah, sure, good point.  Perhaps another alternative would be to use `method(method_name)` (I'm changing the block parameter here) to get a method object into scope and then call that method inside of the method definition.  That might be faster than either because there's no need to figure out which method is being referenced.

I'd definitely be interested in a benchmark if you throw one together ;)  [benchmark-ips](https://github.com/evanphx/benchmark-ips) is a great gem for that if you haven't used it.  Just remember that if you're testing something that can run 100,000 times per second and comparing it against something that runs 120,000 times per second, that's some pretty fast code so we should make sure that it's on critical paths before worrying too much about optimization.

ProGM:
Here's my test
```ruby
# frozen_string_literal: true
require 'benchmark/ips'

GC.disable

class Bar
  def foo
  end

  def bar
  end
end

class Foo
  def initialize
    @bar = Bar.new
    @method = @bar.method(:foo)
  end

  class_eval "def bar\n @bar.bar \nend"
  define_method('foo') { @bar.send(:foo) }
  define_method('foo2') { @method.call }
end

Benchmark.ips do |x|
  foo = Foo.new
  x.report('class_eval') { foo.bar }
  x.report('define_method') { foo.foo }
  x.report('define_method caching method') { foo.foo2 }
end
```

Here's my results:

```ruby
$ ruby benchmark.rb
Calculating -------------------------------------
          class_eval   146.839k i/100ms
       define_method   140.506k i/100ms
define_method caching method
                       135.354k i/100ms
-------------------------------------------------
          class_eval      7.268M (± 3.3%) i/s -     36.416M
       define_method      4.807M (± 2.7%) i/s -     24.027M
define_method caching method
                          4.322M (± 3.4%) i/s -     21.657M
```

My verdict:
Even if class_eval, in this case, seems slightly faster, it allocates more memory, and it seems not very critical.

Moreover, as you said, `define_method` is much more readable and maintainable. 👍 
What do you think?

cheerfulstoic:
That's interesting how the relative times per 100ms and per 1s are very different.  Either way, with millions of operations per second I'm guessing it's better to save memory than performance in this case, so I'm good with your change.

:shipit: at will!

Issue #1210, opened 2016-05-27
`neo4j:migrate` task is not creating a session (and not initializing the app)


ProGM:
If I run:

```bash
$ bin/rake neo4j:migrate
```

I get this error:
```bash
rake aborted!
No session, please create a session first with Neo4j::Session.open(:server_db) or :embedded_db

Tasks: TOP => neo4j:migrate:all
(See full trace by running task with --trace)
```

Instead if I run this:

```bash
$ bin/rake db:migrate neo4j:migrate
```

The migration runs correctly. I think that the rake task is not initializing the environment.

### Runtime information:

Neo4j database version:
`neo4j` gem version: 7.0.8
`neo4j-core` gem version: 6.0.4

cheerfulstoic:
My guess is that `db:migrate` is loading the Rails environment and thus the `railtie.rb`.  The `legacy_migrate` seems to load the environment like this:

https://github.com/neo4jrb/neo4j/blob/master/lib/neo4j/tasks/migration.rake#L5

But would that work outside of Rails?  Could we use a generic way to open the session?  Maybe the code from `railtie.rb` should be extracted and simply called from `railtie.rb` and then we could also call it from things like these `rake` tasks or give people the option of using it manually so that they could just get things like `config/neo4j.yml` etc... for free

ProGM:
Good idea! Maybe I can take a look ;)

cheerfulstoic:
By all means! ;)

ProGM:
Closing here ;)

PR #1231, opened 2016-07-06
Loading environment in migration tasks

ProGM:
Temporary fix for #1210 

Probably it would not work for non-rails, but for now it's a good solution, I think.

This pull introduces/changes:
 * Initialized environment in all migration rake tasks

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/lib/neo4j/migrations/runner.rb
+++ b/lib/neo4j/migrations/runner.rb
@@ -98,7 +98,11 @@ module Neo4j
       end
 
       def files_path
-        Rails.root.join('db', 'neo4j', 'migrate', '*.rb')
+        app_root.join('db', 'neo4j', 'migrate', '*.rb')
+      end
+
+      def app_root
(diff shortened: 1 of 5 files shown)


cheerfulstoic:
This seems fine.  I think maybe we depend on `environment` elsewhere, but for non-Rails apps we may need to define an `environment` task which can be overridden by Rails (or whatever app-specific stuff) so that it doesn't just fail.  If you've checked that this works in non-Rails I'm happy to have the change

ProGM:
@cheerfulstoic Uhm, you're right, maybe I should define a custom environment task to be overwritten by rails

ProGM:
@cheerfulstoic I rebased master to avoid conflicts with #1235 

I moved the part that creates the neo4j session in a new class `Neo4j::SessionManager` and I conditionally defined a `environment` task, that simply creates the session.

Now migrations should work both inside and outside rails ;)

PR #1233, opened 2016-07-07
Add support for Rails 5

subvertallchris:
This pull introduces/changes:
 * Rails 5!

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,12 @@ end
 
 gem 'listen', '< 3.1'
 
+if RUBY_VERSION.to_f < 2.2
+  gem 'activemodel', '~> 4'
+  gem 'activesupport', '~> 4'
+  gem 'railties', '~> 4'
+end
+
 group 'test' do
   gem 'coveralls', require: false
   gem 'codecov', require: false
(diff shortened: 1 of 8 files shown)


subvertallchris:
I've been running into different issues throughout this. There were changes needed to the Gemfile to keep old Rubies from installing incompatible versions of Rack, then a bug in JRuby 9.0.x.0's `Range` class, and finally a Bundler/JRuby error in the version of JRuby that fixes the aforementioned `Range` bug. I just pushed something that attempts to fix this. I'm confident that our side of the code is solid (at least where our tests are confirmed) but I still want all of this passing before I merge or release.

subvertallchris:
Oh yes! Let's get someone to cosign and we'll make this happen!

Gemfile
```diff
@@ -12,6 +12,12 @@ end
 
 gem 'listen', '< 3.1'
 
+if RUBY_VERSION.to_f < 2.2
```
ProGM: Not very critical, but this has strange behaviors when having versions like:
```ruby
"2.2.3".to_f < "2.10.1".to_f # => false ???
```
I generally prefer:

```
Gem::Version.new("2.2.3") < Gem::Version.new("2.10.1")
```

PR #1245, opened 2016-07-20
Support for Unpermitted Attributes API

### LINKED ISSUE #1244: Can't assign `enum` properties from `ActionController::Parameters` in Rails 5
Can't assign `enum` properties from `ActionController::Parameters` in Rails 5:

```ruby
irb(main):005:0> Desk.new(ActionController::Parameters.new({title: "some", type: 'timeline'}))
=> #<Desk uuid: nil, created_at: nil, title: "some", type: :standard, updated_at: nil>
```

(`type` is ignored, default value is used instead)

The issue seems to be related to rails 5:
In fact `ActionController::Parameters` is no longer a child of `Hash`.

### Runtime information:

Neo4j database version: 2.3.3
`neo4j` gem version: 7.1.0
`neo4j-core` gem version: 6.x

ProGM:
Fixes #1244 

This pull introduces/changes:
 * Fixes bug with enum and `ActionController::Parameters`
 * Implements `ForbiddenAttributesProtection` API from ActiveRecord

Pings:
@cheerfulstoic
@subvertallchris


lib/neo4j/shared/permitted_attributes.rb
```diff
@@ -0,0 +1,23 @@
...
+    def attributes=(attributes)
+      attributes = sanitize_for_mass_assignment(attributes)
+      super(attributes)
+    end
+
+    protected
+
+    def sanitize_for_mass_assignment(attributes)
+      attributes = super(attributes)
+      attributes.respond_to?(:symbolize_keys) ? attributes.symbolize_keys : attributes
```
cheerfulstoic: I think that `symbolize_keys` shouldn't be what `sanitize_for_mass_assignment` is responsible for.  Did you move it out so that it would apply for both `ActiveNode` and `ActiveRel`?

spec/e2e/enum_spec.rb
```diff
@@ -160,4 +160,23 @@
       end.to raise_error(Neo4j::Shared::Enum::ConflictingEnumMethodError)
     end
   end
+
+  context 'when using `ActionController::Parameters`' do
+    require 'rack'
+    require 'rack/test'
+    require 'action_controller/metal/strong_parameters'
+    let(:params) { ActionController::Parameters.new('type' => 'image') }
```
cheerfulstoic: Maybe we could create a helper function like `action_controller_params(hash)` which does the appropriate `require`s to clean things up?

cheerfulstoic:
Looks great!  Just a couple of comments and then merge-at-will!

cheerfulstoic:
I'll cut a 7.1.x when it's done

spec/e2e/enum_spec.rb
ProGM: 👍

lib/neo4j/shared/permitted_attributes.rb
ProGM: @cheerfulstoic  I moved it out to allow `ActiveRel` to be supported as well but... I haven't tested it. I should, indeed.

About symbolize_keys, maybe it's better to split it as you said. ;)

cheerfulstoic:
Was looking at https://github.com/rails/strong_parameters

And apparently you need to do an `include` into `ActiveRecord` to make it work:

https://github.com/rails/strong_parameters#installation

ProGM:
@cheerfulstoic Not really.
Strong Parameters are integrated in rails4+.
That tutorial is for upgrading from rails3 to rails4 😜

ProGM:
@cheerfulstoic A few words on my the last commit:
I tried to propagate the Permitted Attributes check to `ActiveRel` as well. I encountered a strange path when inspecting `create!` and `create` method.
As far as I understood, the main difference between `ActiveRel` and `ActiveNode` initialization is about the necessity of taking care of `from_node` and `to_node`. They can be passed both using an hash, mixed inside properties, or using three parameters.

```
MyRel.create(from_node: some, to_node: some, an_attribute: some)
MyRel.create(from_node, to_node, an_attribute: some)
```

This parameter split/check was replicated kind in a similar way both in (here)[https://github.com/neo4jrb/neo4j/pull/1245/files#diff-cf5b1070c03db808dbe7b8957f229d5aL67], [here](https://github.com/neo4jrb/neo4j/pull/1245/files#diff-4c6b105366c57fc300b90870848c8288L58) and also [here](https://github.com/neo4jrb/neo4j/pull/1245/files#diff-cf5b1070c03db808dbe7b8957f229d5aR25).

However, I noticed that removing any but the one in the `initialize` method works as well (since it is called by any of this).

Specs are working, so I think it was a residue, but... I need confirmation 😜

cheerfulstoic:
Hrmm, that is certainly weird and seems reside-ish to me.  @subvertallchris did a lot of the `ActiveRel` code initially, though, so he might have a better insight

ProGM:
@cheerfulstoic Merged master and improved some tests ;)

@subvertallchris if you can check the ActiveRel part, I think it's ready to be merged (and maybe backported to 7.1.0 as well?)

subvertallchris:
Was going to ask about the removed code but I see it's been addressed. It probably was just old code hanging around. As long as tests are passing, looks good to me.

PR #1250, opened 2016-07-29
Allow to pass a Proc for a default property value

knapo:
This pull introduces/changes:
 * Allow to pass a Proc for a default property value. Similar to mongoid, which allows to have dynamic default values.

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,6 +20,7 @@ end
 
 group 'test' do
   gem 'coveralls', require: false
+  gem 'tins', '< 1.7' if RUBY_VERSION.to_f < 2.0
   gem 'codecov', require: false
   gem 'simplecov', require: false
   gem 'simplecov-html', require: false
--- a/lib/neo4j/shared/declared_properties.rb
+++ b/lib/neo4j/shared/declared_properties.rb
@@ -156,7 +156,7 @@ module Neo4j::Shared
 
     def inject_defaults!(object, props)
       declared_property_defaults.each_pair do |k, v|
-        props[k.to_sym] = v if object.send(k).nil? && props[k.to_sym].nil?
+        props[k.to_sym] = v.respond_to?(:call) ? v.call : v if object.send(k).nil? && props[k.to_sym].nil?
       end
       props
     end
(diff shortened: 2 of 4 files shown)


cheerfulstoic:
This is great, thank you!  I think it would be good to have one more spec that verifies that the proc is called each time.  So maybe it would increment some value rather than returning a static string.

Also, we can't use the stabby proc syntax (`-> { }`) because we need to support jRuby 1.7.  So could you change that to an older style `proc` / `Proc.new` / `lambda`?

Aside from that it seems great to me!

knapo:
@cheerfulstoic Ok, good points. I just noticed that it doesn't work as supposed (same static value for each record). Going to fix them and add the spec you mentioned.

knapo:
@cheerfulstoic Done. The build fails for jruby 1.7 on gem bundling - don't think it's the PR-related.

cheerfulstoic:
I think that you're right.  @ProGM you had an issue with `tins` on the `new_cypher_api` branch, right?  Could it be that there was a version update and both of our branches started failing?  I've got to go to sleep, but feel free to merge this if you have time and think you can fix it.  Otherwise I'll do it Monday morning

cheerfulstoic:
(Oh, and a `CHANGELOG.md` entry, of course ;)

Thanks @knapo !

ProGM:
@knapo As @cheerfulstoic said, I encountered the same issue today on `new_cypher_api` branch.
I fixed it by adding this to my `Gemfile`:

```
gem 'tins', '< 1.7' if RUBY_VERSION.to_f < 2.0
```

cheerfulstoic:
Since `tins` is just a dependency on `coveralls` it shouldn't affect the gem that we release

Try that change and see if Travis passes.  If not, I'll merge it anyway and figure out a way to fix it on Monday.  Thanks!

knapo:
@cheerfulstoic @ProGM Forcing tins version helped - build's passing now. Thx!

ProGM:
LGTM 👍 

Merging, thank you @knapo

PR #1253, opened 2016-08-02
Register execution time and queries in migrations

### LINKED ISSUE #1252: Several issues with migrations
* [x] Migrations should print the performed queries, like active record does
* [x] Migrations should print the execution time
* [x] When making schema and data changes in the same migration, the error should suggest to set `disable_transactions!` in the migration file.
* [x] As an alternative: we might just run the migration itself inside the transaction and do a "rollback" outside of the transaction by checking to see if an error was raised and deleting the SchemaMigration node if so
* [x] When the previous error appears, `http` connection hangs instead of giving an error.
* [x] Schema-only migrations should not raise errors

About migrating from neo4j 7.x to 8.x
* [x] There should be a rails specific generator, to migrate everything at once
* [ ] The `neo4j:generate_schema_migration` should accept multiple parameters:
```
rake neo4j:generate_schema_migration[index,Desk,type,name,some,other]
```
Or also:
```
rake neo4j:generate_schema_migration[index:Desk:type:name,constraint:Desk:other]
```

### Runtime information:

Neo4j database version: 3.0.3
`neo4j` gem version: 8.0.0.alpha.1
`neo4j-core` gem version: 7.0.0.alpha.1

ProGM:
Fixes #1252 

This pull introduces/changes:
 * Prints queries and execution time in migrations
 * Caches model indexes and constraints, to fix migration hanging
 * Fix migrations schema update in transactions

Pings:
@cheerfulstoic
@subvertallchris


lib/neo4j/migrations/base.rb
```diff
@@ -35,6 +35,13 @@ def down
 
       private
 
+      def log_queries
+        subscriber = Neo4j::Core::CypherSession::Adaptors::Base.subscribe_to_query(&method(:output))
+        yield
+      ensure
+        ActiveSupport::Notifications.unsubscribe(subscriber) if subscriber
```
cheerfulstoic: Ah, nice.  This seems familiar now that I see the solution ;)

cheerfulstoic:
Awesomeness and ❤️  all over.  :shipit: if you think it's done

Do you want to build up a few changes before we make an alpha update?

ProGM:
👍 
Maybe I can use this one also for more improvements on migrations ;)

cheerfulstoic:
Cool ;)

Gemfile
```diff
@@ -8,7 +8,7 @@ if branch = ENV['TRAVIS_BRANCH']
   same_branch_exists = `curl --head https://github.com/neo4jrb/neo4j-core/tree/#{branch} | head -1`.match(/200 OK/)
   gem 'neo4j-core', github: 'neo4jrb/neo4j-core', branch: same_branch_exists ? branch : 'master'
 else
-  gem 'neo4j-core', github: 'neo4jrb/neo4j-core'
```
cheerfulstoic: Did you mean to leave this in?

This was intended to serve as the option when there isn't a branch with the same name in the `neo4j-core` repository.  I'm definitely often specifying the local `../neo4j-core` path so maybe this whole if/else statement can exist inside of a `if ENV['CI']` and the `else` for that would be to use the local repo

lib/neo4j/migrations/base.rb
```diff
@@ -34,6 +26,43 @@ def down
...
 
+      def migrate_up
+        schema = SchemaMigration.create!(migration_id: @migration_id, incomplete: true)
+        begin
+          run_migration(:up)
+        rescue StandardError => e
+          schema.destroy
+          handle_migration_error!(e)
+        end
+        schema.update!(incomplete: false)
```
cheerfulstoic: I would use `nil` instead of `false` so that the property is dropped

Gemfile
ProGM: @cheerfulstoic Ooops, I committed it by mistake ;(

Maybe we can require local path by using an ENV?

lib/neo4j/migrations/base.rb
ProGM: Oh, I didn't know about it! 👍

spec/e2e/migrations_spec.rb
```diff
@@ -1,5 +1,7 @@
 module Neo4j
+  # rubocop:disable Metrics/ModuleLength
   module Migrations
+    # rubocop:enable Metrics/ModuleLength
```
cheerfulstoic: We should perhaps disable this cop for everything under `spec`

Gemfile
cheerfulstoic: Yeah, that's probably a better idea ;p

Issue #1255, opened 2016-08-04
`rails g neo4j:model` should generate a migration as well


ProGM:
When creating a new model with neo4j, a new migration should be added as well, that defines the `uuid` constraint. ;) This is exactly how the AR generator works

### Runtime information:

Neo4j database version: 3.0.3
`neo4j` gem version: 8.0.0
`neo4j-core` gem version: 7.0.0

PR #1256, opened 2016-08-05
Improve rails generators, added a migrator for 8.x.

### LINKED ISSUE #1252: Several issues with migrations

ProGM:
Fixes #1252 #1255 

This pull introduces/changes:
 * Model generator creates a migration as well
 * A generator to migrate all legacy model schema definitions at once

W.I.P.

Pings:
@cheerfulstoic
@subvertallchris


ProGM:
Tested on a rich rails project:
The migration task works quite good, but can't identify index defined by `enum`s

I think we need an `ensure_enum_defined!` or something like this.

cheerfulstoic:
Don't we define enums at load time of the class?  Maybe we need reflection to see what enums are defined?

ProGM:
Indeed I'm confused. I though enums where created on class definition, but in some way, you can't get them in the indexes list till you make a query. :\
Investigating...

ProGM:
@cheerfulstoic I think I finished. ;)

subvertallchris:
This is really great. 👍

cheerfulstoic:
Seems great to me too!  Whats up with:

 * Making enum indexes warnings and not exceptions
 * #1261 (this might also be appropriate for back-porting to 7.1)

PR #1268, opened 2016-08-15
Refactor code in the Reloading module

cheerfulstoic:
This is an attempt to fix the issue where `CypherNode` objects cease to be wrapped anymore until the Ruby process (server, background workers, whatever) process is restarted.

@subvertallchris I ripped out some of this and so I'm somewhat worried that I've broken something, but it seems to work locally on my app.  Curious for your feedback

### DIFF
--- a/lib/neo4j/active_node/labels/reloading.rb
+++ b/lib/neo4j/active_node/labels/reloading.rb
@@ -2,20 +2,9 @@ module Neo4j::ActiveNode::Labels
   module Reloading
     extend ActiveSupport::Concern
 
-    MODELS_TO_RELOAD = []
-
     def self.reload_models!
-      MODELS_TO_RELOAD.each(&:constantize)
-      MODELS_TO_RELOAD.clear
-    end
-
-    module ClassMethods
-      def before_remove_const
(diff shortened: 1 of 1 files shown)


subvertallchris:
The lines you pulled out aim to do exactly what you're having trouble with. [This PR](https://github.com/neo4jrb/neo4j/pull/945) has some notes that are familiar but I can't remember who was seeing this or how to reproduce. I guess I don't have better tests in there because the reloader makes the difficult, wish I had included more notes. :-(

If I had to guess, maybe a reload of the app followed by a query that returns nodes without loaded models would return CypherNode? That doesn't really make sense to me since t should just follow the process that occurs when the app is first started. It should look at the labels and try to find the class. I'm not entirely sure... Maybe sneak it into the alpha and we'll test it along with everything else?

cheerfulstoic:
:+1:, thanks ;)

cheerfulstoic:
(We've been using this branch at work for a while now and I don't think we've seen the CypherNode error or any other loading issues.  We'll see how it goes in 8.0 alpha)

PR #1270, opened 2016-08-15
Accepting `nil` as enum value

### LINKED ISSUE #1261: Allow nil for enums
It seems the current enum implementation default the model's property to the first declared value if none is specified.  For example:

```
class Service
  include Neo4j::ActiveNode

  enum service_type: {
    foo: 0,
    bar: 1,
  }
```

Will mean that `Service.create` will return a Foo Service by default, even if no `service_type` is given.  

Ideally, I would like to allow a user to select from a list and fail a validation if no selection is given.  

Is there a way I can get the enum to default to `nil` instead of `0` in this case?

ProGM:
Fixes #1261 

This pull introduces/changes:
 * Adds a `:_default` option to define the default value for enums.

This is a braking change.
If we want to keep the previous behavior, we can implement an `_allow_nil` option, to enable `nil` as default value.

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/neo4j.gemspec
+++ b/neo4j.gemspec
@@ -29,7 +29,7 @@ A Neo4j OGM (Object-Graph-Mapper) for use in Ruby on Rails and Rack frameworks h
   s.add_dependency('orm_adapter', '~> 0.5.0')
   s.add_dependency('activemodel', '~> 4')
   s.add_dependency('activesupport', '~> 4')
-  s.add_dependency('neo4j-core', '>= 6.0.0')
+  s.add_dependency('neo4j-core', '< 7.0.0')
   s.add_dependency('neo4j-community', '~> 2.0') if RUBY_PLATFORM =~ /java/
   s.add_development_dependency('railties', '~> 4')
   s.add_development_dependency('pry')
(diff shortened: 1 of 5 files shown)


neo4j.gemspec
```diff
@@ -29,7 +29,7 @@ A Neo4j OGM (Object-Graph-Mapper) for use in Ruby on Rails and Rack frameworks h
   s.add_dependency('orm_adapter', '~> 0.5.0')
   s.add_dependency('activemodel', '~> 4')
   s.add_dependency('activesupport', '~> 4')
-  s.add_dependency('neo4j-core', '>= 6.0.0')
```
cheerfulstoic: Maybe both `>= 6.0.0` and `< 7.0.0`?  Is that possible in a `gemspec`?

ProGM: @cheerfulstoic  Yup, it's possible ;)

cheerfulstoic:
I'm good with this!  I hope that nobody is depending on the breaking change, but they may very well need the `_default` option for backwards compatibility until they can fix their app, so that's great to have there.  If you merge I can release a patch and put out a warning on Gitter / Slack

ProGM:
Great!
I fixed the gemspec thing ;)

Dunno if it's ok to merge now, or just wait for the test here: #1261

ProGM:
@cheerfulstoic ruby 2.0.0 fails to bundle ._.

ProGM:
Fixed!
Merging ;)

PR #1271, opened 2016-08-22
Backporting #1245 (Unpermitted parameters) to 7.2.x

ProGM:
Backporting #1245 to 7.x

I'm opening a PR since there where a few conflicts.

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,7 +25,6 @@ group 'test' do
     gem 'overcommit', '< 0.35.0'
   else
     gem 'overcommit'
-    gem 'activesupport', '>= 4.2'
   end
   gem 'codecov', require: false
   gem 'simplecov', require: false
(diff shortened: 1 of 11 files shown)


cheerfulstoic:
Seems good to me, but I think that this would be `when you add functionality in a backwards-compatible manner` which would be a minor update according to [semver.org](http://semver.org/)

Shall we release as 7.2?  I've been trying to release patch versions whenever there's something to be released which leads us up to `.15` sometimes, but I feel like we hold back on minor updates

ProGM:
Since this fixes this bug:
https://github.com/neo4jrb/neo4j/issues/1244

I considered it as a "bugfix", but instead it's a feature.

So, for me it's ok to release a 7.2.x version ;)

PR #1272, opened 2016-08-23
ActiveRel factories will build CREATE String

subvertallchris:
Fixes the bug documented in specs: when using ActiveRel to create nodes in addition to rels, node properties that look like Cypher params were being interpreted literally. This modifies ActiveRel's query factory to build a CREATE string. There will be a performance boost, too, since the entire hash of properties will be parameterized instead of just the values.

### DIFF
--- a/lib/neo4j/shared/query_factory.rb
+++ b/lib/neo4j/shared/query_factory.rb
@@ -69,7 +69,8 @@ module Neo4j::Shared
 
     def create_query
       return match_query if graph_object.persisted?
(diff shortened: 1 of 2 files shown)


ProGM:
This looks good to me!
I'm not 100% sure I understood the bug. What do you mean for "being interpreted literally". You mean that `{}` in string parameters are interpreted like parameters bindings?

subvertallchris:
@ProGM That is correct. If you have a value that is `{my actual value}`, the gem interprets it as a parameter and it can cause an error.

PR #1276, opened 2016-08-30
remove blank ids in _ids= methods (active_record compatibility)

klobuczek:
Fixes #

This pull introduces/changes:
 * as much as I dislike the change as it should not be a concern of an ORM, this is required for compatibility with active record. My colleagues use this feature quite extensively. Apparently in rails when implementing a collection as series of checkboxes if all are unchecked this is the value you get in the controller `['']`. Why this cleanup has been implemented in activerecord instead of actionpack? Go figure.
 * at the same time added a convenience to assign single id in _ids= methods, which is as well present in activerecord

Pings:
@cheerfulstoic
@subvertallchris

…methods and cleaning blank values out of the argument list

### DIFF
--- a/lib/neo4j/active_node/has_n.rb
+++ b/lib/neo4j/active_node/has_n.rb
@@ -386,7 +386,7 @@ module Neo4j::ActiveNode
 
         define_method_unless_defined("#{name.to_s.singularize}_ids=") do |ids|
           clear_deferred_nodes_for_association(name)
-          association_proxy(name).replace_with(ids)
+          association_proxy(name).replace_with(Array(ids).reject(&:blank?))
         end
 
         define_method_unless_defined("#{name.to_s.singularize}_neo_ids") do
--- a/spec/e2e/has_many_spec.rb
+++ b/spec/e2e/has_many_spec.rb
@@ -557,7 +557,7 @@ describe 'has_many' do
     end
 
     it 'ignores blank IDs for has_many' do
-      post = Post.create(comment: Comment.create)
+      post = Post.create(comments: Comment.create)
 
       post.comment_ids = ['', nil]
 
--- a/spec/e2e/has_many_spec.rb
+++ b/spec/e2e/has_many_spec.rb
@@ -556,6 +556,26 @@ describe 'has_many' do
       expect(post.comments.to_a).to match_array([comment])
     end
 
+    it 'ignores blank IDs for has_many' do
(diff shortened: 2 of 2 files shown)


cheerfulstoic:
I'm totally fine with the change, though one thing that we've had to be careful about is making sure that changes that we make which work in Rails will also work outside of Rails on older versions of Ruby.  With `master` / 8.0 we've bumped up our requirement from Ruby 1.9 to 2.1 (though jRuby is a different story....).

All of that said, do you know where `blank?` comes from?  I know it was originally an `ActiveSupport` thing.  We actually use `ActiveSupport`, though I don't see it loaded anywhere specifically (I think it would be `active_support/core_ext/object/blank`).  I do see `require 'active_support'` in `spec/unit/load_hooks_spec.rb` but I think you need to `require 'active_support/all'` to bring in everything, but if not we may be loading things into our test environment which make us think we can use things when we can't.

Could you check to see where blank is coming from by doing something like:

```ruby
ids.each do |id|
  puts id.method(:blank?).source_location.inspect
end
```

That should tell you the file and line number that the specific `blank?` methods are coming from.  If you get `nil` that should mean that it's from the Ruby source

klobuczek:
The `blank?` method does come from `active_support/core_ext/object/blank.rb`. What do you suggest to do? Reimplement the `blank?` method in neo4j or don't call the method if the object does not `respond_to?(:blank?)`. 
activerecord does not have the ambition to run without activesupport: https://github.com/rails/rails/blob/master/activerecord/activerecord.gemspec#L24

klobuczek:
@cheerfulstoic maybe I misunderstood the reason for your question, but I see neo4j relies on activesupport as well: https://github.com/neo4jrb/neo4j/blob/master/neo4j.gemspec#L31.
So we should be good.
In detail the `blank?` method comes from those 2 places respectivelly for String and Nil with a generic `blank?` on `Object` as well:
["/Users/heinrich/.rvm/gems/ruby-2.3.0/gems/activesupport-5.0.0/lib/active_support/core_ext/object/blank.rb", 114]
["/Users/heinrich/.rvm/gems/ruby-2.3.0/gems/activesupport-5.0.0/lib/active_support/core_ext/object/blank.rb", 52]

cheerfulstoic:
Yeah, probably this is fine.  We can either require the appropriate `ActiveSupport` files or refactor this later

Thanks for this!

PR #1279, opened 2016-08-31
Validates presence on has_many associations

klobuczek:
Fixes #
This is an attempt to start fixing numerous bugs around validation and saving of associations which is either wrong or inconsistent with active_record. It is by no means complete.
This pull introduces/changes:
 * Fixes validation for new object
 * Does not fix validation for update

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/lib/neo4j/active_node/has_n.rb
+++ b/lib/neo4j/active_node/has_n.rb
@@ -29,10 +29,14 @@ module Neo4j::ActiveNode
 
       extend Forwardable
       %w(include? empty? count find first last ==).each do |delegated_method|
-        def_delegator :@enumerable, delegated_method
+        def_delegator :deferred_objects_or_enumerable, delegated_method
       end
 
       include Enumerable
+
+      def deferred_objects_or_enumerable
+        @query_proxy.start_object.try(:new_record?) ? @deferred_objects : @enumerable
+      end
 
       def each(&block)
         result_nodes.each(&block)
--- a/lib/neo4j/active_node/has_n.rb
+++ b/lib/neo4j/active_node/has_n.rb
@@ -35,7 +35,7 @@ module Neo4j::ActiveNode
       include Enumerable
 
       def deferred_objects_or_enumerable
-        #This is not good enough. Updates have to be taken care of as well
+        # This is not good enough. Updates have to be taken care of as well
         @query_proxy.start_object.try(:new_record?) ? @deferred_objects : @enumerable
       end
 
--- a/lib/neo4j/active_node/has_n.rb
+++ b/lib/neo4j/active_node/has_n.rb
@@ -35,6 +35,7 @@ module Neo4j::ActiveNode
       include Enumerable
 
       def deferred_objects_or_enumerable
+        #This is not good enough. Updates have to be taken care of as well
         @query_proxy.start_object.try(:new_record?) ? @deferred_objects : @enumerable
       end
 
--- a/spec/e2e/unpersisted_association_spec.rb
+++ b/spec/e2e/unpersisted_association_spec.rb
@@ -221,7 +221,11 @@ describe 'association creation' do
       it 'does not raise error, creates rel on save' do
         expect_any_instance_of(Neo4j::Core::Query).not_to receive(:delete)
         expect { chris.lesson_ids = [math.id] }.not_to raise_error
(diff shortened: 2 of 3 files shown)


lib/neo4j/active_node/has_n.rb
```diff
@@ -29,11 +29,16 @@ def inspect
...
       %w(include? empty? count find first last ==).each do |delegated_method|
-        def_delegator :@enumerable, delegated_method
+        def_delegator :deferred_objects_or_enumerable, delegated_method
       end
 
       include Enumerable
 
+      def deferred_objects_or_enumerable
+        #This is not good enough. Updates have to be taken care of as well
+        @query_proxy.start_object.try(:new_record?) ? @deferred_objects : @enumerable
```
subvertallchris: We should avoid `try` if possible, too.

spec/e2e/unpersisted_association_spec.rb
```diff
@@ -221,7 +221,11 @@ def save_and_expect_rel!
       it 'does not raise error, creates rel on save' do
         expect_any_instance_of(Neo4j::Core::Query).not_to receive(:delete)
         expect { chris.lesson_ids = [math.id] }.not_to raise_error
-        expect { chris.save }.to change { chris.lessons.count }
+        expect { chris.save }.not_to change { chris.lessons.count }
+      end
```
subvertallchris: I think you should demonstrate here that the relationship is still present when a new instance of the node is loaded from the DB. You could say...

```ruby
expect { chris.save }.to change { Lesson.where(id: math.id).students.count }.by(1)
```

or...

```ruby
chris.save
expect(Student.find(chris).lessons.count).to eq 1
```

spec/e2e/unpersisted_association_spec.rb
```diff
@@ -221,7 +221,11 @@ def save_and_expect_rel!
...
+
+      it 'changes count' do
```
subvertallchris: You might save someone a moment of consideration in the future if you changed this to just `it 'changes count, even before persisting the node'` or something to be super clear about this example. (Ignore this if an earlier context makes it clear!)

spec/e2e/validation_association_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    it 'should be valid with comments' do
+      expect(Post.new(comments: [Comment.create])).to be_valid
+    end
+  end
+
+  #The below spec pass on active_record as is
+  context 'active_record behaviour' do
+    before :each do
+      @post = Post.create(comments: [Comment.create])
+    end
```
subvertallchris: We try to avoid the use of instance variables in our specs these days. Instead:

```ruby
let(:post) { Post.create(comments: [Comment.create]) }
```

And then you can just call `post` in your examples.

lib/neo4j/active_node/has_n.rb
klobuczek: Yes, the `try` is here because I don't understand under which conditions the start_object is actually present. This line is anyway very questionable as it does not take care of persisted dirty records which it should.

klobuczek:
Thanks for the feedback. I would love some comments on how to make sure all those methods delegate to the right object before save even on objects to be updated. The `start_object` for new records is debatable as well.

lib/neo4j/active_node/has_n.rb
```diff
@@ -29,11 +29,16 @@ def inspect
...
       %w(include? empty? count find first last ==).each do |delegated_method|
-        def_delegator :@enumerable, delegated_method
+        def_delegator :deferred_objects_or_enumerable, delegated_method
       end
 
       include Enumerable
 
+      def deferred_objects_or_enumerable
+        # This is not good enough. Updates have to be taken care of as well
+        @query_proxy.start_object.try(:new_record?) ? @deferred_objects : @enumerable
```
cheerfulstoic: Regarding Chris' comment on `try`, is the `start_object` `nil` sometimes?  I can't imagine it would be something other than an `ActiveNode` object...

ProGM:
@klobuczek Can you take a look to #1293 ?

klobuczek:
@ProGM @cheerfulstoic I believe the entire scope of this PR should be fixed by https://github.com/neo4jrb/neo4j/pull/1293

ProGM:
Closing 👍

PR #1281, opened 2016-09-12
Relax restriction on ActiveSupport & ActiveModel to allow version 3.2 of those libraries

dominicsayers:
This pull introduces/changes:
 * backports of ActiveSupport modules & ActiveModel modules missing from version 3.2 (copied unchanged from Rails 4.2.7.1)
 * Workaround for ActiveSupport 3.2 inflection that incorrectly classifies names such as `test_class` to `TestClas`
 * Amended expectations in specs where Rails 3.2 behaviour is different
 * Two specs marked as pending where the existing expectation seems counter-intuitive (expects the `_was` value to have changed)
 * One spec marked as pending where the time returned differs from the time set (as discussed with @cheerfulstoic)

I'm not expecting this PR to be merged any time soon - there's probably too much churn for that. I leave it here in case anybody needs to use Neo4J in an existing Rails 3.2 project as I did.

Pings:
@cheerfulstoic
@subvertallchris


dominicsayers:
The Rails code I imported doesn't use the same Rubocop standards as this project so they don't pass. I can exclude them from Rubocop if you like.

dominicsayers:
I forgot to not redefine the imported modules if they already existed. Closing this and reopening when I've fixed it :-1:

PR #1282, opened 2016-09-12
Relax restriction on ActiveSupport & ActiveModel to allow version 3.2 of those libraries

dominicsayers:
This pull introduces/changes:
 * backports of ActiveSupport modules & ActiveModel modules missing from version 3.2 (copied unchanged from Rails 4.2.7.1)
 * Workaround for ActiveSupport 3.2 inflection that incorrectly classifies names such as `test_class` to `TestClas`
 * Amended expectations in specs where Rails 3.2 behaviour is different
 * One spec marked as pending where the time returned differs from the time set (as discussed with @cheerfulstoic)

I'm not expecting this PR to be merged any time soon - there's probably too much churn for that. I leave it here in case anybody needs to use Neo4J in an existing Rails 3.2 project as I did.

Pings:
@cheerfulstoic
@subvertallchris


subvertallchris:
Hey Dominic! This is very cool, I'm glad you were able to get things working. I'm not sure that this is something that should be merged in, as it puts a big maintenance commitment on the project. I'm all for this being available to anyone who wants to us it, of course, so maybe it could be maintained and released to Rubygems as an independent gem? What do you think?

dominicsayers:
Hi Chris. I agree it should not be merged, there's way too much backported code. When I've used it in production for a while I'll release it as a separate gem as you suggest.

In the meantime I'll close this PR to cut down on the noise for you.

PR #1283, opened 2016-09-13
Railtie refactor

cheerfulstoic:
Refactored railtie to start to extract session connection logic out for use outside of Rails

Neo4j sessions are now established when needed, not at Rails startup.  This fixes sessions in multiple threads and avoids the need to have a Neo4j server up when doing things like compiling assets

Pings:
@subvertallchris
@ProGM


cheerfulstoic:
Should be based off `8.0.x`

Issue #1267, opened 2016-08-14
session_url cannot be configured in config/application.rb anymore


klobuczek:
In 8.0 session_url configured in config/application.rb (or the other env files) is ignored.

Additional information which could be helpful if relevent to your issue:

### Runtime information:

Neo4j database version:
`neo4j` gem version: 8.0
`neo4j-core` gem version:

cheerfulstoic:
Thanks for noting this!

ProGM:
Released as 8.0.0.alpha.11

PR #1286, opened 2016-09-14
fixed railtie bugs

klobuczek:
Fixes #

- https://github.com/neo4jrb/neo4j/issues/1267
- nil error on config.session.url
- fixes swapped url and path

This pull introduces/changes:
 * 
 * 

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/lib/neo4j/railtie.rb
+++ b/lib/neo4j/railtie.rb
@@ -9,8 +9,11 @@ require 'neo4j/core/cypher_session/adaptors/embedded'
 
 module Neo4j
   class Railtie < ::Rails::Railtie
-    config.neo4j = ActiveSupport::OrderedOptions.new
-    config.neo4j.session = ActiveSupport::OrderedOptions.new
+    def empty_config
+      ActiveSupport::InheritableOptions.new(sessions: ActiveSupport::OrderedOptions.new)
+    end
+
+    config.neo4j = empty_config
 
     if defined?(ActiveSupport::Reloader)
       ActiveSupport::Reloader.to_prepare do
@@ -58,7 +61,7 @@ module Neo4j
       end
     end
 
-    def setup!(neo4j_config = OpenStruct.new(session: {}))
+    def setup!(neo4j_config = empty_config)
       type, url, path, options, wait_for_connection = final_config!(neo4j_config).values_at(:type, :url, :path, :options, :wait_for_connection)
       register_neo4j_cypher_logging(type || default_session_type)
 
--- a/lib/neo4j/railtie.rb
+++ b/lib/neo4j/railtie.rb
@@ -10,7 +10,7 @@ require 'neo4j/core/cypher_session/adaptors/embedded'
 module Neo4j
   class Railtie < ::Rails::Railtie
     def empty_config
-      ActiveSupport::InheritableOptions.new(sessions: ActiveSupport::OrderedOptions.new)
+      ActiveSupport::InheritableOptions.new(session: ActiveSupport::OrderedOptions.new)
     end
 
     config.neo4j = empty_config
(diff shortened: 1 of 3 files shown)


klobuczek:
I think this is still not perfect. I am getting a failure on rake tasks. The Railtie code might initialized differently on rake.

klobuczek:
It's all good now, aside from the mysterious `setup!` invocation in migration.rake.

lib/neo4j/railtie.rb
```diff
@@ -57,8 +58,8 @@ class Railtie < ::Rails::Railtie
       end
     end
 
-    def setup!(neo4j_config = nil)
-      type, url, path, options, wait_for_connection = final_config!(neo4j_config).values_at(:type, :path, :url, :options, :wait_for_connection)
+    def setup!(neo4j_config = OpenStruct.new(session: {}))
```
cheerfulstoic: This is a good change, but I wonder if we shouldn't use `ActiveSupport::OrderedOptions` instead of `OpenStruct` since this is the `Railtie`.  But we would need to refactor that rake task that we were talking about on Gitter

lib/neo4j/railtie.rb
```diff
@@ -57,8 +58,8 @@ class Railtie < ::Rails::Railtie
...
+      type, url, path, options, wait_for_connection = final_config!(neo4j_config).values_at(:type, :url, :path, :options, :wait_for_connection)
```
cheerfulstoic: Oh, gosh, thanks for catching this

lib/neo4j/railtie.rb
```diff
@@ -74,8 +75,10 @@ def final_config!(neo4j_config)
     end
 
     def support_deprecated_session_configs!(neo4j_config)
-      ActiveSupport::Deprecation.warn('neo4j.config.sessions is deprecated, please use neo4j.config.session (not an array)') if neo4j_config.sessions.present?
-      neo4j_config.session ||= (neo4j_config.sessions && neo4j_config.sessions[0]) || {}
+      if neo4j_config.sessions.present?
```
cheerfulstoic: Maybe you meant `session` instead of `sessions`?

klobuczek: No, I meant `sessions`. The 2nd replaced line makes sense only under the above condition. Previously this line contained a lazy initialization as a side effect, but in a wrong place. We cannot depend on this deprecation method to initialize the session. That's the reason for default parameter in the `setup!` method.

lib/neo4j/railtie.rb
klobuczek: changed

ProGM:
It looks pretty good to me. What do you think @cheerfulstoic ?

klobuczek:
It works, however, I still don't understand why https://github.com/klobuczek/neo4j/blob/master/lib/neo4j/tasks/migration.rake#L9 is called with an empty config on rake tasks depending on `:environment`, but this is not subject to this pull request.

lib/neo4j/railtie.rb
cheerfulstoic: ;(  Huh, when I refactored things (I don't recall when) I thought I made it so that `neo4j_config` doesn't have a `sessions` property anymore but rather the `session` property is set to `sessions[0]`.  That refactor might have happened after this PR.  I believe we made some more changes recently so you might want to merge `master` in

klobuczek: But that's exactly what this code is doing. I am assuming you want to keep old configurations working. If someone has configured `sessions` we will take the first one. Newest master is now in.
My refactoring of your code is equivalent expect that I removed that hidden initialization side effect.

klobuczek: keep in mind as well that I just moved the existing `if` condition from being a modifier to a full statement.

cheerfulstoic: Oh, right, I'm sorry.  Carry on ;)

PR #1287, opened 2016-09-16
Allow HTTPS scheme in the NEO4J_URL

jacob-ewald:
[This](https://github.com/neo4jrb/neo4j/commit/fdf341d44be5914c2aced54751fb1a7c983e760a) commit introduced a change to enforce the URI scheme on the `NEO4J_URL`, which inadvertently broke the scenario of using an `HTTPS` scheme. This PR restores the ability to use an `HTTPS` scheme (by masquerading as a regular `HTTP` scheme).


cheerfulstoic:
Will merge once travis passes

PR #1288, opened 2016-09-16
Skip queries for unpersisted objects

### LINKED ISSUE #1273: Database query fired when interrogating unpersisted association on upersisted object.
When trying to get values of associations on unpersisted objects database queries with practically false conditions are fired. Additionally depending on the method chaining variations the original value of the association might not be returned. Please see below. 

```
class Structure
  include Neo4j::ActiveNode
  id_property :neo_id # I think it's not limited to :neo_id
  has_many :out, :bases, type: :basis, model_class: Structure
end
```
```
[11] pry(main)> >> Structure.new(bases: [Structure.new]).bases.first
 Structure#bases 
  MATCH (structure)
  WHERE (ID(structure) IS NULL)
  MATCH (structure)-[rel1:`basis`]->(result_bases:`Structure`)
  RETURN result_bases
  ORDER BY ID(result_bases)
  LIMIT {limit_1} | {:limit_1=>1}
 HTTP REQUEST: 12ms POST http://localhost:7476/db/data/transaction/commit (1 bytes)
=> nil
[12] pry(main)> >> Structure.new(bases: [Structure.new]).bases.to_a.first
 Structure#bases 
  MATCH (structure)
  WHERE (ID(structure) IS NULL)
  MATCH (structure)-[rel1:`basis`]->(result_bases:`Structure`)
  RETURN result_bases 
 HTTP REQUEST: 3ms POST http://localhost:7476/db/data/transaction/commit (1 bytes)
=> #<Structure neo_id: nil, archived: nil, comment: nil, created_at: nil, file_name: nil, name: nil, updated_at: nil>
```
### Runtime information:

Neo4j database version:
`neo4j` gem version: master
`neo4j-core` gem version: master

ProGM:
Fixes #1273

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/lib/neo4j/active_node/query/query_proxy_enumerable.rb
+++ b/lib/neo4j/active_node/query/query_proxy_enumerable.rb
@@ -14,7 +14,7 @@ module Neo4j
         end
 
         def result(node = true, rel = nil)
-          return [].freeze if @start_object && @start_object.new_record?
+          return [].freeze if unpersisted_start_object?
 
           @result_cache ||= {}
           return result_cache_for(node, rel) if result_cache?(node, rel)
--- a/lib/neo4j/active_node/query/query_proxy_enumerable.rb
+++ b/lib/neo4j/active_node/query/query_proxy_enumerable.rb
@@ -14,6 +14,8 @@ module Neo4j
         end
 
         def result(node = true, rel = nil)
+          return [].freeze if @start_object && @start_object.new_record?
+
           @result_cache ||= {}
           return result_cache_for(node, rel) if result_cache?(node, rel)
 
--- a/lib/neo4j/active_node/has_n.rb
+++ b/lib/neo4j/active_node/has_n.rb
@@ -28,7 +28,7 @@ module Neo4j::ActiveNode
       end
 
       extend Forwardable
-      %w(include? empty? count find first last ==).each do |delegated_method|
+      %w(include? find first last ==).each do |delegated_method|
         def_delegator :@enumerable, delegated_method
       end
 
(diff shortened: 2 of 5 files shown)


klobuczek:
This is part of a bigger problem some aspects of which https://github.com/neo4jrb/neo4j/pull/1279 tries to address.

ProGM:
@klobuczek I see! I haven't seen your changes on this topic in your PR. I can try to fix this minor problem for now, then merge together with your. What do you think?

On a side note, please try to link PR to issues, so it's easier to keep track of changes ;)

lib/neo4j/active_node/query/query_proxy_enumerable.rb
```diff
@@ -14,6 +14,8 @@ def each(node = true, rel = nil, &block)
         end
 
         def result(node = true, rel = nil)
+          return [].freeze if @start_object && @start_object.new_record?
```
klobuczek: as I remember this will not help to prevent querying the database on updates in active model hooks which should use the newly updated values instead of the old values from the database.

ProGM: Indeed, I think it prevents querying when accessing directly to associations.

I'll probably need to write another spec and a fix for your case. Good catch, thanks!

klobuczek: Have a look as well how I tried to solve the problem there (like you using the `@start_object). And like you I covered the creation case only, but not the update case.

PR #1290, opened 2016-09-19
Fix `where` only using the first array value in question mark queries

### LINKED ISSUE #1247: Arrays passed as question mark params only use the first value

```ruby
.where("something.something IN ?", ['foo', 'bar'])
```

### Runtime information:

Neo4j database version:
`neo4j` gem version: 7.0.14
`neo4j-core` gem version: 6.1.4

ProGM:
Fixes #1247

Pings:
@cheerfulstoic
@subvertallchris


ProGM:
:+1:

PR #1293, opened 2016-09-22
Fix validations with unpersisted nodes and various bugs with associations

### LINKED ISSUE #1292: Associations change in `update` is not rollbacked when validations fail
```ruby
p = Post.find(some_id)
p.update(comments: [])
# => false
# The relations are saved even if update fails!!!
p.comments # => []
```

ProGM:
Fixes #1292 , #1273 and is a replacement for both #1288 and #1279

This pull introduces/changes:
 * Does not fire database when accessing to unpersisted associations
 * Fixes `.count` method on associations, that should ALWAYS fire the database, also if the cache is available (like active record does)
* Fixed `size` and `length` on associations, that should take account of @deferred_objects
* Added (and passed) all AR specs about relationships: https://github.com/neo4jrb/neo4j/blob/16143eec2a0c1cae8434282f0d7bf8456c49497d/spec/e2e/validation_association_spec.rb Thanks @klobuczek 

Pings:
@cheerfulstoic
@subvertallchris
@klobuczek

### DIFF
--- a/lib/neo4j/active_node/has_n.rb
+++ b/lib/neo4j/active_node/has_n.rb
@@ -38,9 +38,8 @@ module Neo4j::ActiveNode
         result_nodes.each(&block)
       end
 
-      def count(*args)
-        @deferred_objects.count + @enumerable.count(*args)
-      end
+      # .count always hits the database
+      def_delegator :@query_proxy, :count
 
       def length
         @deferred_objects.length + @enumerable.length
@@ -75,9 +74,7 @@ module Neo4j::ActiveNode
       def result_nodes
         return result_objects if !@query_proxy.model
 
-        result_objects.map do |object|
-          object.is_a?(Neo4j::ActiveNode) ? object : @query_proxy.model.find(object)
-        end
+        map_results_as_nodes(result_objects)
       end
 
       def result_objects
@@ -117,9 +114,12 @@ module Neo4j::ActiveNode
       end
 
       def replace_with(*args)
-        @cached_result = nil
-
-        @query_proxy.public_send(:replace_with, *args)
+        nodes = @query_proxy.public_send(:replace_with, *args).to_a
+        if @query_proxy.start_object.try(:new_record?)
+          @cached_result = nil
+        else
+          cache_result(nodes)
+        end
       end
 
       QUERY_PROXY_METHODS = [:<<, :delete, :create, :pluck, :where, :where_not, :rel_where, :rel_order, :order, :skip, :limit]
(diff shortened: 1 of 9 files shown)


lib/neo4j/active_node/has_n.rb
```diff
@@ -101,9 +114,12 @@ def cached?
       end
 
       def replace_with(*args)
-        @cached_result = nil
-
-        @query_proxy.public_send(:replace_with, *args)
+        nodes = @query_proxy.public_send(:replace_with, *args).to_a
```
cheerfulstoic: Why not `@query_proxy.replace_with(*args).to_a`?

lib/neo4j/active_node/query/query_proxy_methods.rb
```diff
@@ -39,6 +39,7 @@ def order_property
 
         # @return [Integer] number of nodes of this class
         def count(distinct = nil, target = nil)
+          return 0 if unpersisted_start_object?
```
cheerfulstoic: Returning `0` feels like we've made a query and gotten an answer.  Maybe this should be `nil`?

lib/neo4j/active_node/query/query_proxy.rb
```diff
@@ -1,7 +1,9 @@
 module Neo4j
   module ActiveNode
     module Query
+      # rubocop:disable Metrics/ClassLength
```
cheerfulstoic: `rubocop:disable` makes @cheerfulstoic :sadpanda:

Not saying this can't be fixed later ;p

lib/neo4j/active_node/has_n.rb
ProGM: @cheerfulstoic  Really don't know. I left as it was 😆 We can change it, I think ;)

lib/neo4j/active_node/query/query_proxy_methods.rb
ProGM: @cheerfulstoic  This is how AR works... so, I wouldn't change that

lib/neo4j/active_node/query/query_proxy.rb
ProGM: You're right. Probably we should split the `QueryProxy` module a bit later ;P

lib/neo4j/active_node/query/query_proxy_methods.rb
cheerfulstoic: I thought it might be, thanks ;p

cheerfulstoic:
:+1:

ProGM:
Merged, thank you for your suggestions @klobuczek !

PR #1297, opened 2016-09-25
removing unnecessary limitation of scope to maximum 1 parameter

klobuczek:
Fixes #

This pull introduces/changes:
 * scope can now take more than 1 parameter
 * the previous error message was very confusing if defined scope with 2 parameters and called with 2:
` wrong number of arguments (given 0, expected 1)`

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/lib/neo4j/active_node/scope.rb
+++ b/lib/neo4j/active_node/scope.rb
@@ -39,7 +39,7 @@ module Neo4j::ActiveNode
 
         klass = class << self; self; end
         klass.instance_eval do
-          define_method(name) do |query_params = nil, _ = nil|
+          define_method(name) do |*query_params|
             eval_context = ScopeEvalContext.new(self, current_scope || self.query_proxy)
             proc = full_scopes[name.to_sym]
             _call_scope_context(eval_context, query_params, proc)
@@ -68,13 +68,8 @@ module Neo4j::ActiveNode
       end
 
       def _call_scope_context(eval_context, query_params, proc)
-        if proc.arity == 1
-          eval_context.instance_exec(query_params, &proc)
-        else
-          eval_context.instance_exec(&proc)
-        end
+        eval_context.instance_exec(*query_params.fill(nil, query_params.length..proc.arity - 1), &proc)
       end
-
 
       def current_scope #:nodoc:
         ScopeRegistry.value_for(:current_scope, base_class.to_s)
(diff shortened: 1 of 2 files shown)


spec/e2e/scope_spec.rb
```diff
@@ -97,6 +97,26 @@
     end
   end
 
+  describe 'Person.scope :great_students, -> (identifier, score) { where("#{identifier}.score > ?", score)' do
+    before(:each) do
+      Person.scope :great_students, -> (identifier, score) { where("#{identifier}.score > ?", score || 41) }
```
ProGM: This lambda syntax is not valid on jruby 1.7 https://github.com/jruby/jruby/issues/2704

It should just work by removing the empty space between the `->` and `(identifier, scope) ...`:

```ruby
Person.scope :great_students, ->(identifier, score) { where("#{identifier}.score > ?", score || 41) }
```

ProGM:
I noticed this problem as well, this is great :+1: 
I left a comment on it

lib/neo4j/active_node/scope.rb
```diff
@@ -68,14 +68,9 @@ def full_scopes
       end
 
       def _call_scope_context(eval_context, query_params, proc)
-        if proc.arity == 1
-          eval_context.instance_exec(query_params, &proc)
-        else
-          eval_context.instance_exec(&proc)
-        end
+        eval_context.instance_exec(*query_params.fill(nil, query_params.length..proc.arity - 1), &proc)
```
cheerfulstoic: Ah, nice, I didn't know about `Array#fill`!

:+1:

spec/e2e/scope_spec.rb
```diff
@@ -97,6 +97,26 @@
     end
   end
 
+  describe 'Person.scope :great_students, -> (identifier, score) { where("#{identifier}.score > ?", score)' do
+    before(:each) do
+      Person.scope :great_students, ->(identifier, score) { where("#{identifier}.score > ?", score || 41) }
```
cheerfulstoic: This is fine, though I think that in 8.0 we support calling `identity` from within a scope now

klobuczek: I was merely looking for a test with 2 parameters. :)

cheerfulstoic: I thought so ;)  Just wanted to make sure that you knew about it__

PR #1298, opened 2016-09-26
Fix `as` losing the current query chain scope

### LINKED ISSUE #1278: Aliasing QueryProxy without losing scope
I have the following ActiveNode class
````
[1] pry(main)> DD::Error.count
 CYPHER 17ms MATCH (n:`DD::Error`) RETURN count(n) AS count 
=> 30990
`````
a scope defined in order to filter the class:
````
  scope :active, -> { where(status: 'active') }
````
so that:
````
[4] pry(main)> DD::Error.active.count
 CYPHER 34ms MATCH (result_dderror:`DD::Error`) WHERE (result_dderror.status = {result_dderror_status}) RETURN count(result_dderror) AS result_dderror | {:result_dderror_status=>"active"}
=> 2646
[5] pry(main)> my_qp = DD::Error.active
=> #<QueryProxy  CYPHER: "MATCH (result_dderror:`DD::Error`) WHERE (result_dderror.status = {result_dderror_status})">
[6] pry(main)> my_qp.count
 CYPHER 33ms MATCH (result_dderror:`DD::Error`) WHERE (result_dderror.status = {result_dderror_status}) RETURN count(result_dderror) AS result_dderror | {:result_dderror_status=>"active"}
=> 2646
````
I need to alias the QueryProxy in order to do some dynamic filtering later on. However it doesn't seem to work.
````
[9] pry(main)> my_qp.as(:a).count
 DD::Error 13ms MATCH (a:`DD::Error`) RETURN count(a) AS a 
=> 30990
````
However using the following combination works properly:
````
[11] pry(main)> my_qp.query_as(:a).proxy_as(DD::Error, :a).count
 CYPHER 41ms MATCH (a:`DD::Error`) WHERE (a.status = {a_status}) RETURN count(a) AS a | {:a_status=>"active"}
=> 2646
````

Is this expected behaviour?
### Runtime information:

Neo4j database version:
`neo4j` gem version: 7.2.0
`neo4j-core` gem version: 6.1.4

### PR DESCRIPTION
**Pull request by ProGM**
Fixes #1278 

This pull introduces/changes:
 * `.as` doesn't break query chain anymore

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/lib/neo4j/active_node/query/query_proxy.rb
+++ b/lib/neo4j/active_node/query/query_proxy.rb
@@ -186,11 +186,8 @@ module Neo4j
         #
         # @return [QueryProxy] A new QueryProxy
         def branch(&block)
-          if block
-            instance_eval(&block).query.proxy_as(self.model, identity)
-          else
-            fail LocalJumpError, 'no block given'
-          end
+          fail LocalJumpError, 'no block given' unless block
+          instance_eval(&block).query.proxy_as(self.model, identity)
         end
 
         def [](index)
--- a/lib/neo4j/active_node/query/query_proxy.rb
+++ b/lib/neo4j/active_node/query/query_proxy.rb
@@ -186,7 +186,7 @@ module Neo4j
         #
         # @return [QueryProxy] A new QueryProxy
         def branch(&block)
-          fail LocalJumpError, 'no block given' unless block
+          fail LocalJumpError, 'no block given' if block.nil?
           instance_eval(&block).query.proxy_as(self.model, identity)
         end
 
--- a/spec/e2e/query_proxy_methods_spec.rb
+++ b/spec/e2e/query_proxy_methods_spec.rb
@@ -511,6 +511,26 @@ describe 'query_proxy_methods' do
       end
     end
 
+    describe 'as' do
+      it 'changes query variable' do
+        query = Student.as(:stud)
+        expect(query.to_cypher).to include('(stud:`Student`)')
+        query.to_a
+      end
+
+      it 'keeps the current context' do
+        query = Student.where(name: 'John').as(:stud)
+        expect(query.to_cypher).to include('(stud:`Student`)', ' WHERE ')
+        query.to_a
+      end
+
+      it 'keeps the current context with associations' do
+        query = Student.all.lessons.as(:less)
+        expect(query.to_cypher).to include('`Student`)', '(less:`Lesson`)')
+        query.to_a
+      end
+    end
+
     describe 'delete, destroy' do
       before { @john.lessons << @history unless @john.lessons.include?(@history) }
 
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: lib/neo4j/active_node/query/query_proxy.rb [hunk 1]
```diff
@@ -186,11 +186,8 @@ def <<(other_node)
         #
         # @return [QueryProxy] A new QueryProxy
         def branch(&block)
-          if block
-            instance_eval(&block).query.proxy_as(self.model, identity)
-          else
-            fail LocalJumpError, 'no block given'
-          end
+          fail LocalJumpError, 'no block given' unless block
```
cheerfulstoic: This is a pet peeve of mine, so be warned ;)

From a standpoint of language, saying "fail unless X" is weird because most of the time you won't be failing.  It feels somewhat better to say "fail if not X" because it's supposed to be a rare case

spec/e2e/query_proxy_methods_spec.rb
```diff
@@ -511,6 +511,26 @@ def destroy_called
...
+
+      it 'keeps the current context' do
+        query = Student.where(name: 'John').as(:stud)
+        expect(query.to_cypher).to include('(stud:`Student`)', ' WHERE ')
+        query.to_a
+      end
+
+      it 'keeps the current context with associations' do
+        query = Student.all.lessons.as(:less)
+        expect(query.to_cypher).to include('`Student`)', '(less:`Lesson`)')
```
cheerfulstoic: Is passing multiple arguments to `include` an RSpec thing?  Does it mean that we expect both of those strings to be included (like an `and`)?

lib/neo4j/active_node/query/query_proxy.rb
ProGM: @cheerfulstoic 
Rubocop 0.4 doesn't agree with you 
Just joking, I understand what you're saying 😜
I changed that thing because of a suggestion of it from my Editor, but no problem about restoring it ;)

spec/e2e/query_proxy_methods_spec.rb
ProGM: Yup! It works like that

lib/neo4j/active_node/query/query_proxy.rb
cheerfulstoic: I've ignored that rubocop rule in a number of projects, but I couldn't remember if I had done that here.  I like your `if block.nil?` suggestion from Gitter

I'm not a fanatic, exactly, but I have participated in conversations like this one: https://github.com/bbatsov/ruby-style-guide/issues/329

ProGM:
Changed to `if branch.nil?`. Waiting for CI, then merging! ;)

PR #1301, opened 2016-10-01
update_node_property/update_node_properties, AKA update_column(s)

### LINKED ISSUE #1285: add update_columns method to bypass timestamp update
ActiveRecord::Persistence implements an [update_column(s) method](http://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-update_column) to be able to update an attribute value directly in the database, bypassing validations,callbacks, and timestamp updates.

In my use case, this would be used to implement a last_access timestamp which is updated when the record is viewed, leaving the updated_at timestamp as-is.

subvertallchris:
This'll close #1285. I agree with @cheerfulstoic's statement that `update_column` is a little weird in our context, but I think that plain ole `update_property` is also potentially misleading because we have a very well-defined concept of "properties" within the gem. I made the distinction of _node_ properties because it strikes me as just a tiny bit clearer, but I'm open to something else if someone has a good suggestion.

Pings:
@cheerfulstoic
@ProGM

### DIFF
--- a/lib/neo4j/shared/persistence.rb
+++ b/lib/neo4j/shared/persistence.rb
@@ -189,6 +189,25 @@ module Neo4j::Shared
     end
     alias update_attributes update
 
+    def update_db_property(field, value)
+      update_db_properties(field => value)
+      true
+    end
+    alias update_column update_db_property
+
+    def update_db_properties(hash)
+      fail ::Neo4j::Error, 'can not update on a new record object' unless persisted?
+      self.class.run_transaction do
+        db_values = props_for_db(hash)
+        neo4j_query(query_as(:n).set(n: db_values))
+        db_values.each_pair { |k, v| self.public_send(:"#{k}=", v) }
+        _persisted_obj.props.merge!(db_values)
+        db_values.each_key { |k| changed_attributes.delete(k) }
+        true
+      end
+    end
+    alias update_columns update_db_properties
+
     # Same as {#update_attributes}, but raises an exception if saving fails.
     def update!(attributes)
       self.class.run_transaction do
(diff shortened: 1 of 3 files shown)


cheerfulstoic:
`update_node_property(ies)` doesn't make sense because it's in `shared` and thus on `ActiveRel` right?  I'm totally down with just `update_property(ies)`

lib/neo4j/shared/persistence.rb
```diff
@@ -189,6 +189,25 @@ def update(attributes)
...
+    alias update_column update_node_property
+
+    def update_node_properties(hash)
+      fail ::Neo4j::Error, 'can not update on a new record object' unless persisted?
+      self.class.run_transaction do
+        db_values = props_for_db(hash)
+        neo4j_query(query_as(:n).set(n: db_values))
+        db_values.each_pair { |k, v| self.public_send(:"#{k}=", v) }
+        _persisted_obj.props.merge!(db_values)
+        db_values.each_key { |k| changed_attributes.delete(k) }
```
cheerfulstoic: `ActiveModel::Dirty` has a `clear_attribute_changes`:

http://api.rubyonrails.org/classes/ActiveModel/Dirty.html#method-i-clear_attribute_changes

subvertallchris:
Ah right, ActiveRel. I think `update_property` is still potentially
misleading, can we think of something better?

cheerfulstoic:
`update_neo4j_property` or `update_db_property`?

subvertallchris:
Yeah, I was thinking db property. Let's go with that, I can do it tonight
and we can get this in!

ProGM:
Looks great! 👍 Not sure that "db properties" is clear, but every AR user would use `update_column(s)` without thinking too much about it.

subvertallchris:
Aaaand... four days later, let's make sure things are still passing. If so, I will merge.

ProGM:
👍

PR #1305, opened 2016-10-06
Implemented `distinct` in QueryProxy

### LINKED ISSUE #1217: `distinct` method for proxies
This has seemed unnecessary to me in the past, but something that I was just thinking about:

```ruby
class Model
  include Neo4j::ActiveNode

  def transitive_association
    association1.association2
  end
end
```

If you just called `Model.first.transitive_association`, you may very well want distinct results, but you could also want to chain the results like `Model.first.transitive_association.association3`.  To address that, I think it makes sense to have a `distinct` method so that enumerations will add a `DISTINCT`.

ProGM:
Fixes #1217

This pull introduces/changes:
 * Implements `distinct` in query proxy

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@ All notable changes to this project will be documented in this file.
 This file should follow the standards specified on [http://keepachangelog.com/]
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+# Added
+- `distinct` method for QueryProxy (thanks @ProGM / see #1305)
+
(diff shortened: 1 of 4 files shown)


cheerfulstoic:
Easy enough!  I love it, thanks!

I was thinking about the suggestion from @klobuczek in Gitter about being able to label the path variable for an association.  I think it could be just about as easy as this and also really useful (hint hint, if you have time ;) )

I should be finally winding down my current long-running side project that's been taking so much of my time, so hopefully I can help with some things a bit more in the near future.

PR #1309, opened 2016-10-10
Many migrations improvements and bugfixes

ProGM:
This pull introduces/changes:
 * `drop_constraint` and `drop_index` now accepts `force: true`
 * `PendingMigrationError` now shows which migrations are pending
 * Fixed `silenced: true`, not working properly
 * Removed "strange" inheritance between `Neo4j::Migrations::Base` and the legacy `Neo4j::Migration` class
 * Avoid creating the `SchemaMigration` model constraint when it already exists

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/lib/neo4j/migrations.rb
+++ b/lib/neo4j/migrations.rb
@@ -11,7 +11,8 @@ module Neo4j
     class << self
       def check_for_pending_migrations!
         runner = Neo4j::Migrations::Runner.new
-        fail ::Neo4j::PendingMigrationError if runner.pending_migrations?
+        pending = runner.pending_migrations
+        fail ::Neo4j::PendingMigrationError, pending if pending.any?
       end
 
       attr_accessor :currently_running_migrations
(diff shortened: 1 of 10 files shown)


lib/neo4j/migrations.rb
```diff
@@ -11,7 +11,8 @@ module Migrations
     class << self
       def check_for_pending_migrations!
         runner = Neo4j::Migrations::Runner.new
-        fail ::Neo4j::PendingMigrationError if runner.pending_migrations?
+        pending = runner.pending_migrations
+        fail ::Neo4j::PendingMigrationError, pending if pending.any?
```
cheerfulstoic: What does this output?  Perhaps `pending.inspect` to turn it into a string?

ProGM: @cheerfulstoic check here:
https://github.com/neo4jrb/neo4j/pull/1309/files#diff-3ced7dc77d960f7c24fdee3fe17efe5cR40
It's passed as a parameter to the error, and joined with "\n"

ProGM:
Merged!

lib/neo4j/migrations.rb
cheerfulstoic: Oh, nice ;)  Thanks for the clarification

PR #1324, opened 2016-12-05
Implementing after_commit hooks

### LINKED ISSUE #1218: Implement after_commit callback
Given that queries are wrapped in a transaction by default, would it be possible to implement the _commit and _rollback callbacks?

ProGM:
Fixes #1218

This pull introduces/changes:
 * `after_create_commit`, `after_update_commit` and `after_destroy` commit hooks

It's pretty raw, but functional (at least in my tests).

I made a PR on neo4j-core as well:
https://github.com/neo4jrb/neo4j-core/pull/277

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/spec/e2e/active_model_spec.rb
+++ b/spec/e2e/active_model_spec.rb
@@ -221,10 +221,6 @@ describe 'Neo4j::ActiveNode' do
           attr_reader :"before_#{verb}_called", :"after_#{verb}_called"
         end
 
-        %w(update_commit create_commit destroy_commit).each do |verb|
-          attr_reader :"after_#{verb}_called"
-        end
-
         attr_reader :after_find_called, :after_initialize_called
 
         property :name
@@ -242,10 +238,6 @@ describe 'Neo4j::ActiveNode' do
         after_destroy { @after_destroy_called = true }
         before_validation { @before_validation_called = true }
         after_validation { @after_validation_called = true }
-
-        after_create_commit { @after_create_commit_called = true }
-        after_update_commit { @after_update_commit_called = true }
-        after_destroy_commit { @after_destroy_commit_called = true }
       end
     end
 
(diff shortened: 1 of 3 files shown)


spec/shared_examples/after_commit.rb
```diff
@@ -0,0 +1,45 @@
...
+    else
+      company.save
+      transactions.last.mark_failed if fail_transaction
+      close_inner_transactions!
+      expect { transactions.first.close }.send(to_or_not_to, change { c.after_create_commit_called })
+    end
+  end
+
+  it 'handles after_update_commit callbacks' do
+    c
```
cheerfulstoic: Having the `c` inherited from the scope that the shared context is used in was really confusing to me.  Can we have something like:

```ruby
shared_context 'after_commit' do |company_variable, options|
  let(:company) { send(:company_variable) }
end
```

spec/e2e/active_model_spec.rb
```diff
@@ -221,6 +221,10 @@ def mark_saved
           attr_reader :"before_#{verb}_called", :"after_#{verb}_called"
         end
 
+        %w(update_commit create_commit destroy_commit).each do |verb|
+          attr_reader :"after_#{verb}_called"
+        end
```
cheerfulstoic:
```ruby
 %w(update create destroy).each do |verb|
  attr_reader :"after_#{verb}_commit_called"
end
```

spec/shared_examples/after_commit.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  let(:transactions) { Array.new(transactions_count) { Neo4j::ActiveBase.new_transaction } }
+  let(:close_inner_transactions!) { transactions.reverse.first(transactions_count - 1).each(&:close) }
+  let(:to_or_not_to) { options[:fail_transaction] ? :not_to : :to }
+
+  it 'handles after_create_commit callbacks' do
+    company = Company.new
+
+    if transactions.empty?
+      expect { company.save }.to change { c.after_create_commit_called }
```
cheerfulstoic: It also seems weird that you're defining the `after_*_commit_called` outside of the shared context and then using it within

spec/e2e/active_model_spec.rb
ProGM: Just right. ;)

ProGM:
@cheerfulstoic I think I fixed ;) What do you think?

cheerfulstoic:
❤️ it, thanks!

PR #1331, opened 2016-12-19
1330 fix

### LINKED ISSUE #1330: Missing rake task from pure Ruby implementation
I'm trying to use the `neo4j` gem in a non-Rails, non-Sinatra library. I keep getting the following message when I try to write to the database:

```
Neo4j::DeprecatedSchemaDefinitionError:           Some schema elements were defined by the model (which is no longer supported), but they do not exist in the database.  Run the following to create them:

rake neo4j:generate_schema_migration[constraint,AstVisualizer::AstNode,uuid]
```

I tried following the instructions and running that rake task:

```
$ bundle exec rake neo4j:generate_schema_migration[constraint,AstVisualizer::AstNode,uuid]
rake aborted!
Don't know how to build task 'neo4j:generate_schema_migration' (see --tasks)
```

My Rakefile just has `require 'neo4j'` in it.

The project is at https://github.com/CoralineAda/ast_visualizer

To reproduce:

`irb -r './lib/ast_visualizer.rb'`
`AstVisualizer.build_graph("lib/ast_visualizer/ast_node.rb")`

Any help would be greatly appreciated.

### Runtime information:

Neo4j database version: community-latest
`neo4j` gem version: 8.0.0
`neo4j-core` gem version: 7.0.0

cheerfulstoic:
Fixes #1330

This pull introduces/changes:
 * Migration rake tasks are always loaded since they are needed for anybody using the `neo4j` gem (not just rails apps)
 * Removed dependency on the `Railtie` for loading an environment for non-Rails apps (I'd still like this to be able to look for the `neo4j.yml`, but this should work for now)

Pings:
@subvertallchris
@ProGM

### DIFF
--- a/lib/neo4j/tasks/migration.rake
+++ b/lib/neo4j/tasks/migration.rake
@@ -9,7 +9,7 @@ if !defined?(Rails) && !Rake::Task.task_defined?('environment')
     neo4j_url = ENV['NEO4J_URL'] || 'http://localhost:7474'
     $:.unshift File.dirname('./')
     Neo4j::ActiveBase.on_establish_session do
-      type = neo4j_url.match(/^bolt/) ? :bolt : http
+      type = neo4j_url.match(/^bolt/) ? :bolt : :http
       Neo4j::SessionManager.open_neo4j_session(type, neo4j_url)
     end
   end
--- a/lib/neo4j/tasks/migration.rake
+++ b/lib/neo4j/tasks/migration.rake
@@ -6,7 +6,12 @@ if !defined?(Rails) && !Rake::Task.task_defined?('environment')
   task :environment do
     require 'neo4j/session_manager'
     require 'ostruct'
-    Neo4j::Railtie.setup!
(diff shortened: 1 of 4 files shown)


lib/neo4j/tasks/migration.rake
```diff
@@ -6,7 +6,12 @@ if !defined?(Rails) && !Rake::Task.task_defined?('environment')
   task :environment do
     require 'neo4j/session_manager'
     require 'ostruct'
-    Neo4j::Railtie.setup!
+    neo4j_url = ENV['NEO4J_URL'] || 'http://localhost:7474'
+    $:.unshift File.dirname('./')
+    Neo4j::ActiveBase.on_establish_session do
+      type = neo4j_url.match(/^bolt/) ? :bolt : http
```
phreakocious: I think http here should be :http ..  
This commit resolves my issues if I set that environment variable to a bolt:// URL in the Rakefile but fails with this otherwise:
'NameError: undefined local variable or method `http' for main:Object'

cheerfulstoic:
Fixes #1353

Issue #1338, opened 2016-12-31
QueryProxy `exists?` uses `COUNT`, could be faster


subvertallchris:
This Ruby...

```ruby
Student.lessons.exists?
```

...generates this Cypher...

```
 Student#lessons 
  MATCH (node2:`Student`)
  MATCH (node2)-[rel1:`ENROLLED_IN`]->(result_lessons:`Lesson`)
  RETURN COUNT(result_lessons) AS count
```

...but it could generate this...

```
Student#lessons
  MATCH (node2:`Student`)
  MATCH (node2)-[rel1:`ENROLLED_IN`]->(result_lessons:`Lesson`)
  RETURN ID(result_lessons) LIMIT 1
```

We don't need to count, we just need to know that there is one to prove existence. Counting everything can be quite slow.

thefliik:
looks like this issue has been addressed

cheerfulstoic:
🎉

PR #1339, opened 2016-12-31
`exists?` performance, primary key, response, tests

subvertallchris:
* Look for presence of a single value instead of counting all matches
* Allow String arguments, which will check for primary_key
* Fix response, only return boolean, no more `nil`

#1338

### DIFF
--- a/lib/neo4j/active_node/query/query_proxy_methods.rb
+++ b/lib/neo4j/active_node/query/query_proxy_methods.rb
@@ -94,12 +94,13 @@ module Neo4j
         end
 
         def exists?(node_condition = nil, target = nil)
-          unless node_condition.is_a?(Integer) || node_condition.is_a?(Hash) || node_condition.nil?
-            fail(Neo4j::InvalidParameterError, ':exists? only accepts neo_ids')
+          unless [Integer, String, Hash, NilClass].include?(node_condition.class)
+            fail(Neo4j::InvalidParameterError, ':exists? only accepts ids or conditions')
           end
           query_with_target(target) do |var|
             start_q = exists_query_start(node_condition, var)
-            start_q.query.reorder.return("COUNT(#{var}) AS count").first.count > 0
+            result = start_q.query.reorder.return("ID(#{var}) AS found_ids LIMIT 1").first
+            !!result && result.found_ids > 0
           end
         end
 
@@ -284,6 +285,8 @@ module Neo4j
             self.where("ID(#{target}) = {exists_condition}").params(exists_condition: condition)
           when Hash
             self.where(condition.keys.first => condition.values.first)
+          when String
+            self.where(model.primary_key => condition)
           else
             self
           end
--- a/lib/neo4j/active_node/query/query_proxy_methods.rb
+++ b/lib/neo4j/active_node/query/query_proxy_methods.rb
@@ -99,8 +99,8 @@ module Neo4j
           end
           query_with_target(target) do |var|
             start_q = exists_query_start(node_condition, var)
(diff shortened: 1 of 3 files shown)


subvertallchris:
Pings:
@cheerfulstoic 
@ProGM

lib/neo4j/active_node/query/query_proxy_methods.rb
```diff
@@ -94,12 +94,13 @@ def include?(other, target = nil)
...
-          unless node_condition.is_a?(Integer) || node_condition.is_a?(Hash) || node_condition.nil?
-            fail(Neo4j::InvalidParameterError, ':exists? only accepts neo_ids')
+          unless [Integer, Fixnum, String, Hash, NilClass].include?(node_condition.class)
+            fail(Neo4j::InvalidParameterError, ':exists? only accepts ids or conditions')
           end
           query_with_target(target) do |var|
             start_q = exists_query_start(node_condition, var)
-            start_q.query.reorder.return("COUNT(#{var}) AS count").first.count > 0
+            result = start_q.query.reorder.return("ID(#{var}) AS found_ids LIMIT 1").first
+            !!result && result.found_ids > 0
```
cheerfulstoic: Shouldn't `found_ids` be `found_id`?  We're returning one row with one column and then I think we should be checking to see if the size of `result` is > 0, right?

lib/neo4j/active_node/query_methods.rb
```diff
@@ -2,12 +2,13 @@ module Neo4j
...
       def exists?(node_condition = nil)
-        unless node_condition.is_a?(Integer) || node_condition.is_a?(Hash) || node_condition.nil?
+        unless [Integer, Fixnum, String, Hash, NilClass].include?(node_condition.class)
           fail(Neo4j::InvalidParameterError, ':exists? only accepts ids or conditions')
         end
         query_start = exists_query_start(node_condition)
         start_q = query_start.respond_to?(:query_as) ? query_start.query_as(:n) : query_start
-        start_q.return('COUNT(n) AS count').first.count > 0
+        result = start_q.return('ID(n) AS found_ids LIMIT 1').first
+        !!result && result.found_ids > 0
```
cheerfulstoic: Same comment as above

cheerfulstoic:
Ah, I get it now.  Couldn't the first ID be 0, though, since the ID is an offset which starts at the beginning of a file?

You could even probably get a boolean from Cypher to make the Ruby code a bit simpler.  Something like:

```ruby
!!start_q.limit(1).with('ID(n) AS found_id').return('count(found_id) > 0').first
```

subvertallchris:
That was one of the reasons why it was a "funny thing to check for!" I'm leaving the Cypher as is but should be good to go momentarily.

cheerfulstoic:
:+1:

subvertallchris:
Oof, I thought this was merged ages ago. Is the 8.0 branch still the best target, @cheerfulstoic?

cheerfulstoic:
Yeah, `8.0.x` is still the running branch for the latest gem releases.  Make sure to merge to `master` afterward and push (I didn't do that one time and somebody else made a PR to solve part of the same problem)

cheerfulstoic:
Since this is on `8.0.x` could you put in a `CHANGELOG` entry (either as `Unreleased` or with a new version)?

subvertallchris:
Already pushed that straight to the branch and released as 8.0.8!

cheerfulstoic:
Ah, right, I suppose I should have actually checked the `CHANGELOG` ;) . Thanks

PR #1350, opened 2017-01-24
Remove before and after association callbacks

cheerfulstoic:
This would technically be a breaking change, though I don't know if anybody uses this feature anymore

Pings:
@progm
@subvertallchris

### DIFF
--- a/lib/neo4j/active_node/query/query_proxy.rb
+++ b/lib/neo4j/active_node/query/query_proxy.rb
@@ -201,13 +201,9 @@ module Neo4j
             other_nodes.each do |other_node|
               other_node.save unless other_node.neo_id
 
-              return false if @association.perform_callback(@start_object, other_node, :before) == false
-
               @start_object.association_proxy_cache.clear
 
               _create_relationship(other_node, properties)
-
-              @association.perform_callback(@start_object, other_node, :after)
             end
           end
         end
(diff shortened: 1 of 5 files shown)


subvertallchris:
Total support.

ProGM:
I added a couple of notes 👍

cheerfulstoic:
Merged to `master` and I've made a `8.1.x` branch with the changes.  Haven't pushed to 8.1.x because we might want more changes (maybe?)

ProGM:
@cheerfulstoic  Ach, I didn't pressed "submit review", so I think you haven't seen my comments D:

No prob, just a couple of things.
👍  for the 8.1.x branch

lib/neo4j/active_node/query/query_proxy.rb
```diff
@@ -258,6 +254,7 @@ def new_link(node_var = nil)
           self.clone.tap do |new_query_proxy|
             new_query_proxy.instance_variable_set('@result_cache', nil)
             new_query_proxy.instance_variable_set('@node_var', node_var) if node_var
+            new_query_proxy.instance_variable_set('@distinct', @distinct) if @distinct
```
cheerfulstoic: Good catch, I think it was a bit of code from something that I was playing with.  It wasn't something that was complete so I just removed from the `8.1.x` branch

PR #1390, opened 2017-06-04
Model generator update to fix #1387

### LINKED ISSUE #1387: Superclass mismatch when Neo4jrb generates model migration w/ trailblazer-rails gem
On a new install of Rails 5.1.1 with Neo4jrb (installed using the `rails new myapp -m http://neo4jrb.io/neo4j/neo4j.rb -O` option described in the setup docs), running `rails g model User` creates a `user.rb` file in the `models` folder containing

    class User
      include Neo4j::ActiveNode
    end

and a migration file containing `class User < Neo4j::Migrations::Base` and results in a `superclass mismatch error` when I try and run `rake neo4j:migrate`.

If I edit the migration file to update its class name to `CreateUser`, the `superclass mismatch error` is eliminated but I get a new `ArgumentError: wrong number of arguments (given 2, expected 0..1)`.

I can't seem to find much information on migration files in the docs, so I'm not sure what the argument error is referring to.

### Runtime information:

Neo4j database version: 3.1.4
`neo4j` gem version: 8.0.17
`neo4j-core` gem version: 7.1.2
`neo4j-rake_tasks` gem version: 0.7.12

thefliik:
Aims to fix #1387 by prefixing the migration automatically created when generating a model with "Constrain" and adding `force: true` to the migration. i.e.:

`rails g model Person` will create
```
app/models/person.rb
db/neo4j/migrate/xxxxxx_constrain_person.rb
```
with `db/neo4j/migrate/xxxxxx_constrain_person.rb` containing
```
class ConstrainPerson < Neo4j::Migrations::Base
  def up
    add_constraint :Person, :uuid, force: true
  end

  def down
    drop_constraint :Person, :uuid
  end
end
```

This is incomplete. So far all I've done is add `force: true` to the model migration template. I imagine I can update the name of the migration file created by `rails g model` by updating the model generator code, but I haven't found where in the code the model generator calls the migration generator.

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,8 @@ docs_site/
 
 # dotenv
 .env
+
+# Ignore vagrant files
+Vagrantfile
+/.vagrant/
+ubuntu-xenial*
--- a/lib/rails/generators/neo4j/model/templates/migration.erb
+++ b/lib/rails/generators/neo4j/model/templates/migration.erb
@@ -1,6 +1,6 @@
 class <%= @migration_class_name.underscore.camelize %> < Neo4j::Migrations::Base
   def up
-    add_constraint :<%= class_name %>, :uuid
+    add_constraint :<%= class_name %>, :uuid, force: true
   end
 
   def down


cheerfulstoic:
Unfortunately `force: true` shouldn't be the default behavior (otherwise we might make it the default value in the code accepting the value).  The more fundamental problem is that migrations should not be created with the same name as models (again, I'd want to know how `ActiveRecord` handles this)

thefliik:
@cheerfulstoic,
When creating a constraint migration manually (i.e. `rake neo4j:generate_schema_migration[constraint,Person,uuid]`), neo4j already adds `force: true` by default. Why would it be ok there but not when running `rails g model Person`? Or am I missing something

PR #1395, opened 2017-06-26
1376 dependent delete orphans

### LINKED ISSUE #1376: `other_rel` not defined with dependent: :delete_orphans
Just got an error using dependent: :delete_orphans. :destroy_orphans works fine.

``` ruby
Neo.ClientError.Statement.SyntaxError: Variable `other_rel` not defined (line 1, column 237 (offset: 236))
"MATCH (self:`LabelOne`) WHERE (ID(self) = {ID_self}) WITH self OPTIONAL MATCH (self)-[other_rel:`REL_ONE`]->(n:`LabelTwo`) WITH n MATCH ()-[orphan_rel:`REL_ONE`]->(
n) WITH n, count(*) AS count WHERE (count = {one}) DELETE n, other_rel"
```
Looks like it is defined, but don't have time to dig into this right now, but wanted to mention if it sounds familiar to others....

``` ruby
LabelOne.last.destroy
````

### Runtime information:

Neo4j database version:
  * neo4j (8.0.15)
  * neo4j-core (7.1.0)

### PR DESCRIPTION
**Pull request by TyGuy**
Fixes #1376 

This pull introduces/changes:
 * Fixes the `dependent: :delete_orphans` option so that it deletes the orphaned node if it exists, without raising an error.

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/lib/neo4j/active_node/dependent/association_methods.rb
+++ b/lib/neo4j/active_node/dependent/association_methods.rb
@@ -28,7 +28,8 @@ module Neo4j
         end
 
         def dependent_delete_orphans_callback(object)
-          object.as(:self).unique_nodes(self, :self, :n, :other_rel).query.delete(:n, :other_rel).exec
+          unique_query = object.as(:self).unique_nodes(self, :self, :n, :other_rel)
+          unique_query.query.detach_delete(:n).exec if unique_query
         end
 
         def dependent_destroy_callback(object)
--- a/spec/e2e/association_dependency_spec.rb
+++ b/spec/e2e/association_dependency_spec.rb
@@ -58,7 +58,7 @@ describe 'association dependent delete/destroy' do
     stub_active_node_class('Comment') do
       property :note
       # In the real world, there would be no reason to setup a dependency here since you'd never want to delete
-      # the topic of a comment just because the topic is destroyed.
+      # the topic of a comment just because the comment is destroyed.
       # For the purpose of these tests, we're setting this to demonstrate that we are protected against loops.
       has_one :in, :topic, model_class: false, type: 'HAS_COMMENT', dependent: :destroy
     end
@@ -95,7 +95,7 @@ describe 'association dependent delete/destroy' do
       end
     end
 
-    context 'a node with relationshpis' do
+    context 'a node with relationships' do
       let(:band) { Band.create }
       before { node.bands << band }
 
--- a/spec/e2e/association_dependency_spec.rb
+++ b/spec/e2e/association_dependency_spec.rb
@@ -36,7 +36,7 @@ describe 'association dependent delete/destroy' do
       after_destroy lambda { CALL_COUNT[:called] += 1 }
       property :city
       has_many :in, :routes, model_class: 'Route', origin: :stops
-      has_many :out, :comments, model_class: 'Comment', type: 'HAS_COMMENT', dependent: :destroy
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: lib/neo4j/active_node/dependent/association_methods.rb [hunk 1]
```diff
@@ -28,7 +28,8 @@ def dependent_delete_callback(object)
         end
 
         def dependent_delete_orphans_callback(object)
-          object.as(:self).unique_nodes(self, :self, :n, :other_rel).query.delete(:n, :other_rel).exec
+          unique_query = object.as(:self).unique_nodes(self, :self, :n, :other_rel)
+          unique_query.query.detach_delete(:n).exec if unique_query
```
subvertallchris: Is truthiness of `unique_query` enough to prove that it should execute the query?

subvertallchris:
Thanks for checking this. Added a comment, as I'm sure you see. Can you also add a couple quick tests?

lib/neo4j/active_node/dependent/association_methods.rb
TyGuy: Yes I think so, because even in the event where the `unique_query` is true and there are no orphan nodes to delete, executing the query will simply not match any orphan nodes (none will match `:n`) and thus not delete any.

TyGuy:
@subvertallchris thanks for the comments. The original test would break with _just_ the spec change and not the code change, so I assumed that would work as a red/green kinda thing. But definitely understand wanting this to be tested more explicitly. Hopefully that additional test case will suffice!

leehericks:
@subvertallchris Just ran into this delete_orphans error. What's the state of this?

subvertallchris:
Sorry for not following up, guys. @cheerfulstoic, could you review? I'm not setup to do any gem development at the moment.

cheerfulstoic:
Sorry, this one got dropped.  I'll merge it it now and let it run the latest tests to see how it goes

lib/neo4j/active_node/dependent/association_methods.rb
cheerfulstoic: A couple of things I'm looking into (again, sorry for the delay in looking at this):

 * We can't use `detach_delete` because we need to support versions of Neo4j before 3.0
 * `unique_nodes` returns false for the case of `send(association.name).empty?`, otherwise is returns a `QueryProxy` (which would always be considered truthy).  I'm actually a bit puzzled because before if `unique_nodes` returned false we wouldn't be able to call `query` on it, so that situation was probably never coming up in the tests.  But if that's the case, I don't yet understand how the result of `unique_query` could not be truthy.

Still digging on that second one...

cheerfulstoic: Ah, I see now that the relationship variable in the `delete` was the problem.  Cool, continuing to look...

cheerfulstoic:
Released as 9.0.7, give it a try, and again, sorry for the delay!

FYI, I basically replaced the `detach_delete` call from this PR's original code with Cypher which is compatible with older versions of Neo4j.  I also needed to add in a `break` so that the clauses go in the right order (I think my next priority when I have time is to make it so that clauses don't re-order themselves.  That was such a confusing thing that I did...)

leehericks:
I've updated my bundle and re-enabled `dependent: :delete_orphans`.

I've assigned the reading of one kanji node to another and then deleted the original kanji node. The one reading remained. When I destroyed the other kanji node, the reading node was deleted as expected.

Thanks @cheerfulstoic!

cheerfulstoic:
🎉

PR #1396, opened 2017-06-27
fixing unreliable order of relationships in labels

klobuczek:
Fixes #
unreliable order of relationships in the enterprise version, relationship ids not necessarily assigned in ascending order.
labels appear to be sorted alphabetically in enterprise-3.2.1
This pull introduces/changes:
 * weakened specs for first and last on ActiveRel. Those methods can no longer be relied on. 
 * sorting labels before adding to cache
 * additionally proposing to limit the build matrix to the neo4j versions which are available on https://neo4j.com/download/other-releases/ and rather test significant functional differences between community and enterprise editions

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,8 @@ env:
   global:
     - JRUBY_OPTS="-J-Xmx1280m -Xcompile.invokedynamic=false -J-XX:+TieredCompilation -J-XX:TieredStopAtLevel=1 -J-noverify -Xcompile.mode=OFF"
   matrix:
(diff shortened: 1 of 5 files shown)


subvertallchris:
It's been a while but I think some of us were proponents of removing `first` and `last` from ActiveRel entirely but left it behind because some people were relying on them. Should those methods be deprecated?

klobuczek:
They might be still helpful in some testing situations, but I don't think anything useful can be achieved with them. I think we should make at least clear that they return random relationship.

cheerfulstoic:
Released in 8.1.3, thanks!

PR #1397, opened 2017-06-28
Fix JSON serialization of Strings

leviwilson:
Fixes #

This pull introduces/changes:
 * adds ability to serialize `String` value only
 * adds specs around leveraging `JSONConverter`

Prior to this change, if you had a model defined like this:

```ruby
class MuhData
  include Neo4j::ActiveNode
  property :thing
  serialize :thing
end
```

If you were to `MuhData.create(thing: 'explodo')` it would be very angry with you. This is because the `db_type` of the `JSONConverter` was _already_ a `String` and the `BaseConverter` thought it didn't have to do anything with it. This is problematic when you try to:

```ruby
JSON.parse 'explodo'
JSON::ParserError: 765: unexpected token at 'explodo'
```

Rather than bypassing, I've modified the `JSONConverter` to `def converted?(_value); false; end` so that it _always_ does a `to_json` in the `to_db`.

The other issue I noticed is that `converted?` was _always_ being bypassed because it was doing `found_converter.respond_to(:converted)` vs. `found_converter.respond_to(:converted?)`.

Feedback welcome.

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/docs/ActiveNode.rst
+++ b/docs/ActiveNode.rst
@@ -137,7 +137,7 @@ If you would like to use multiple labels you can use class inheritance.  In the 
 Serialization
 ~~~~~~~~~~~~~
 
-Pass a property name as a symbol to the serialize method if you want to save a hash or an array with mixed object types* to the database.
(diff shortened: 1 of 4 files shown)


cheerfulstoic:
Released in 8.1.3, thanks!

PR #1398, opened 2017-07-07
deep eager loading

klobuczek:
Fixes #
- fixes issue when there is a relationships type and label with the same name
- fixes broken result caching on QueryProxyEagerLoading

This pull introduces/changes:
 * with_association now accepts hashes of symbol instead of only symbols allowing deep eager loading
 * 

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/lib/neo4j/active_node/query/query_proxy_eager_loading.rb
+++ b/lib/neo4j/active_node/query/query_proxy_eager_loading.rb
@@ -2,58 +2,153 @@ module Neo4j
   module ActiveNode
     module Query
       module QueryProxyEagerLoading
-        def each(node = true, rel = nil, &block)
+        class IdentityMap < Hash
+          def add(node)
+            self[node.neo_id] ||= node
+          end
+        end
+
+        class AssociationTree < Hash
+          attr_accessor :model, :name, :association, :path
+
+          def initialize(model, name = nil)
+            super()
+            self.model = name ? target_class(model, name) : model
+            self.name = name
+            self.association = name ? model.associations[name] : nil
+          end
+
+          def clone
+            super.tap { |copy| copy.each { |key, value| copy[key] = value.clone } }
+          end
+
+          def add_association(spec)
+            if spec.is_a?(Array)
+              spec.each { |s| add_association(s) }
+            elsif spec.is_a?(Hash)
+              spec.each { |k, v| (self[k] ||= AssociationTree.new(model, k)).add_association(v) }
+            else
+              self[spec] ||= AssociationTree.new(model, spec)
+            end
+          end
+
+          def paths(*prefix)
+            values.flat_map { |v| [[*prefix, v]] + v.paths(*prefix, v) }
+          end
+
+          private
+
+          def target_class(model, key)
+            association = model.associations[key]
+            fail "Invalid association: #{[*path, key].join('.')}" unless association
+            model.associations[key].target_class
+          end
+        end
+
+        def pluck_vars(node, rel)
           return super if with_associations_spec.size.zero?
 
-          query_from_association_spec.pluck(identity, "[#{with_associations_return_clause}]").map do |record, eager_data|
-            eager_data.each_with_index do |eager_records, index|
-              record.association_proxy(with_associations_spec[index]).cache_result(eager_records)
+          perform_query
+        end
+
+        def perform_query
+          @_cache = IdentityMap.new
+          query_from_association_spec
+            .map do |record, eager_data|
+            cache_and_init(record, with_associations_spec)
+            eager_data.zip(with_associations_spec.paths.map(&:last)).each do |eager_records, element|
+              eager_records.first.zip(eager_records.last).each do |eager_record|
+                add_to_cache(*eager_record, element)
+              end
             end
 
-            yield(record)
+            record
+          end
+        end
+
+        def with_associations(*spec)
+          new_link.tap do |new_query_proxy|
+            new_query_proxy.with_associations_spec = with_associations_spec.clone
+            new_query_proxy.with_associations_spec.add_association(spec)
           end
         end
 
         def with_associations_spec
-          @with_associations_spec ||= []
+          @with_associations_spec ||= AssociationTree.new(model)
         end
 
-        def with_associations(*spec)
-          invalid_association_names = spec.reject do |association_name|
-            model.associations[association_name]
-          end
-
-          if !invalid_association_names.empty?
-            fail "Invalid associations: #{invalid_association_names.join(', ')}"
-          end
-
-          new_link.tap do |new_query_proxy|
-            new_spec = new_query_proxy.with_associations_spec + spec
-            new_query_proxy.with_associations_spec.replace(new_spec)
-          end
+        def with_associations_spec=(tree)
+          @with_associations_spec = tree
         end
 
         private
 
-        def with_associations_return_clause(variables = with_associations_spec)
-          variables.map { |n| "#{n}_collection" }.join(',')
+        def add_to_cache(rel, node, element)
+          direction = element.association.direction
+          if rel.is_a?(Neo4j::ActiveRel)
+            rel.instance_variable_set(direction == :in ? '@from_node' : '@to_node', node)
+          end
+          @_cache[direction == :out ? rel.start_node_neo_id : rel.end_node_neo_id]
+            .association_proxy(element.name).add_to_cache(cache_and_init(node, element), rel)
+        end
+
+        def init_associations(node, element)
+          element.keys.each { |key| node.association_proxy(key).init_cache }
+        end
+
+        def cache_and_init(node, element)
+          @_cache.add(node).tap { |n| init_associations(n, element) }
+        end
+
+        def with_associations_return_clause(variables = path_names)
+          variables.map { |n| escape("#{n}_collection") }.join(',')
+        end
+
+        def escape(s)
+          s =~ /\./ ? "`#{s}`" : s
+        end
+
+        def path_name(path)
+          path.map(&:name).join('.')
+        end
+
+        def path_names
+          with_associations_spec.paths.map { |path| path_name(path) }
         end
 
         def query_from_association_spec
           previous_with_variables = []
-          with_associations_spec.inject(query_as(identity).with(identity)) do |query, association_name|
-            with_association_query_part(query, association_name, previous_with_variables).tap do
-              previous_with_variables << association_name
+          with_associations_spec.paths.inject(query_as(identity).with(identity)) do |query, path|
+            with_association_query_part(query, path, previous_with_variables).tap do
+              previous_with_variables << path_name(path)
             end
-          end.return(identity)
+          end.pluck(identity, "[#{with_associations_return_clause}]")
         end
 
-        def with_association_query_part(base_query, association_name, previous_with_variables)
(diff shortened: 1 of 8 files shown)


lib/neo4j/active_node/query/query_proxy_eager_loading.rb
```diff
@@ -2,58 +2,153 @@ module Neo4j
...
         private
 
-        def with_associations_return_clause(variables = with_associations_spec)
-          variables.map { |n| "#{n}_collection" }.join(',')
+        def add_to_cache(rel, node, element)
+          direction = element.association.direction
+          if rel.is_a?(Neo4j::ActiveRel)
+            rel.instance_variable_set(direction == :in ? '@from_node' : '@to_node', node)
+          end
+          @_cache[direction == :out ? rel.start_node_neo_id : rel.end_node_neo_id]
```
cheerfulstoic: In testing this out, I got an error that `start_node_neo_id` doesn't exist.  This means that the `gemspec` for this change would need to be changed to require `7.2.2` of the `neo4j-core` gem.  Probably we can release as 8.2.x of the `neo4j` gem, so that shouldn't be a big deal

lib/neo4j/active_node/query/query_proxy_eager_loading.rb
```diff
@@ -2,58 +2,153 @@ module Neo4j
...
+            new_query_proxy.with_associations_spec.add_association(spec)
           end
         end
 
+        def with_associations_spec
+          @with_associations_spec ||= AssociationTree.new(model)
+        end
+
+        def with_associations_spec=(tree)
+          @with_associations_spec = tree
```
cheerfulstoic: Probably we can change all `with_associations_spec` references to something like `with_association_tree`

lib/neo4j/active_node/query/query_proxy_eager_loading.rb
```diff
@@ -2,58 +2,153 @@ module Neo4j
...
+
+          def add_association(spec)
+            if spec.is_a?(Array)
+              spec.each { |s| add_association(s) }
+            elsif spec.is_a?(Hash)
+              spec.each { |k, v| (self[k] ||= AssociationTree.new(model, k)).add_association(v) }
+            else
+              self[spec] ||= AssociationTree.new(model, spec)
             end
+          end
```
cheerfulstoic: Probably this method could be called `add_spec`.  The singular "association" seems weird when you're often adding more than one via array or hash

spec/e2e/association_proxy_spec.rb
```diff
@@ -107,6 +115,26 @@
...
+    end
+  end
+
+  it 'Queries only one time when there are some empty associations' do
+    Student.create.lessons << science
+    Student.create.lessons += [science, Lesson.create]
+    expect_queries(1) do
+      science.students.with_associations(lessons: :exams_given).flat_map(&:lessons).flat_map(&:exams_given)
+    end
+  end
```
cheerfulstoic: I see that you are getting the relationships in the query as well.  Would these be available by doing an `each_rel`?  If so, I think there's at least one spec missing here.

I also wonder if it makes sense to have another function which doesn't return the relationship for the sake of not loading as much data.  Probably that's a project for another day...

lib/neo4j/active_rel/rel_wrapper.rb
```diff
@@ -11,8 +11,9 @@ class << self
       def wrapper(rel)
         rel.props.symbolize_keys!
         begin
-          most_concrete_class = class_from_type(rel.rel_type)
-          most_concrete_class.constantize.new
+          most_concrete_class = class_from_type(rel.rel_type).constantize
+          return rel unless most_concrete_class < Neo4j::ActiveRel
+          most_concrete_class.new
```
cheerfulstoic: :+1: Cheers for this

lib/neo4j/active_node/query/query_proxy_eager_loading.rb
```diff
@@ -2,58 +2,153 @@ module Neo4j
...
+          @_cache.add(node).tap { |n| init_associations(n, element) }
+        end
+
+        def with_associations_return_clause(variables = path_names)
+          variables.map { |n| escape("#{n}_collection") }.join(',')
+        end
+
+        def escape(s)
+          s =~ /\./ ? "`#{s}`" : s
+        end
```
cheerfulstoic: I assume that this isn't trying to avoid Cypher injection vulnerabilities, right?  I suppose generally we should trust the programmers to give us safe values in the `with_associations` argument(s)

lib/neo4j/active_node/query/query_proxy_eager_loading.rb
```diff
@@ -2,58 +2,153 @@ module Neo4j
...
+        def cache_and_init(node, element)
+          @_cache.add(node).tap { |n| init_associations(n, element) }
+        end
+
+        def with_associations_return_clause(variables = path_names)
+          variables.map { |n| escape("#{n}_collection") }.join(',')
+        end
+
+        def escape(s)
+          s =~ /\./ ? "`#{s}`" : s
```
cheerfulstoic: You could use `s['.']`. Both are very very fast, though `s['.']` is about 1.8 times faster in my Benchmarking

cheerfulstoic:
Hey @klobuczek just checking to see what's up with this.

klobuczek:
@cheerfulstoic sorry I dropped the ball on this as I was swamped with other stuff. Thank you very much for the feedback, which I am going to work through right now.

lib/neo4j/active_node/query/query_proxy_eager_loading.rb
klobuczek: Good point about cypher injection. Maybe we should unconditionally escape all aliases as they are not simple generated ones but driven by the parameters to the `with_associations` method. In my case the names are passed via query parameters in jsonapi specifying what sideloading should be included. The names can be verified otherwise, whitelisting only good associations but it is easy to make a mistake and forget about it.

klobuczek:
@cheerfulstoic could you give the performance test you ran and which had the anomaly of having 1000 objects at the 2nd level while total objects were only 7?
I sort of can imagine where the weak point is. The algorithm works nicely for trees, but for arbitrary graphs, it will still traverse all possible paths as if it were a tree and thus it possibly touches and returns the same nodes multiple times before finally reducing the result to the actual object graph in memory.
There are few different strategies for doing I might explore.

spec/e2e/association_proxy_spec.rb
klobuczek: Currently, on master `each_rel` called after `with_association` does not work properly. It wrongfully behaves like `each` passing a node to the block. Here I was actually not attempting to support `each_rel`, but maybe we should fix that for the sake of consistency.

cheerfulstoic:
No problem, thanks for the update.

I can't share my case, but I'm sure it's a pathological one, and no, I believe the performance was the same or better.  I still think that this will be a good option for users to be able to choose to use and, unless it's super easy, I don't think there's a need to explore deeper options yet.  I was mainly curious at the sort of speed-ups you've seen.

klobuczek:
@cheerfulstoic in my case I have a performance improvement of factor 50. But this was an extreme case of a team trying to download an entire database branch for 1 client over jsonapi. This took initially 7 minutes with thousands of queries and now results in 1 query of 2 seconds and 20 second json serialization due to the enormous json size. But the good aspect of this is that this feature has been really tested under a huge load.

lib/neo4j/active_node/query/query_proxy_eager_loading.rb
```diff
@@ -2,58 +2,158 @@ module Neo4j
...
+          @with_associations_tree ||= AssociationTree.new(model)
+        end
+
+        def with_associations_tree=(tree)
+          @with_associations_tree = tree
+        end
+
+        def first
+          limit(1).to_a.first
+        end
```
klobuczek: this is bad and the reason for some random spec failures, I need to correct it

cheerfulstoic:
Awesome, I definitely think that this would be a great improvement, then!  I see that there are some builds failing on CI and you made a comment on the `#first` method.  How close would you say that this is?

Thanks again

klobuczek:
@cheerfulstoic the `first` is the last missing part that I recently discovered. Do you have any recommendation how to fix that? `first` breaks `with_associations` on master as well, so were several other methods. I could fix most of the problems by not overriding `each` but only `pluck_vars`, but the `first` needs different handling.

cheerfulstoic:
I'm not sure I understand the problem.  Is it about ordering?  In the `ActiveNode` `.first` method we order by the primary key:

https://github.com/neo4jrb/neo4j/blob/d3a2412a45c551225b24f4a21760183c3f7e561e/lib/neo4j/active_node/query_methods.rb#L15

klobuczek:
The problem is that a lot of the methods in the query chain are ignored (including `with_associations`) when `first` is called as the `each` method which handles many of them is skipped.

cheerfulstoic:
Ah, I think I see now.  So did d61cd5a43474b86e54f9c102771f34d4e6fcfef0 fix it?

cheerfulstoic:
If so, would you say everything is done?  If so I'm happy to merge / release it

klobuczek:
Yes d61cd5a did fix the problem for `first`. The last commit c83e84 is a workaround for a bug in neo4j 2.1.8 if we still want to support it.
And yes, I think we are done.

cheerfulstoic:
I would like to try to support old versions as is reasonable, but maybe it would be reasonable to drop older versions in version `9.x` of the gem

cheerfulstoic:
Thanks so much for this!

pepesenaris:
@cheerfulstoic  in gem version will be available this?

cheerfulstoic:
Just released as `8.2.1` (I forgot to get everything from the `8.1.x` branch, so I had to release twice ;)

Joshfindit:
If I'm not just missing it somewhere, could we please get documentation on how to do deep eager loading?

cheerfulstoic:
@Joshfindit I just added a note at the bottom of [this section](https://neo4jrb.readthedocs.io/en/latest/ActiveNode.html#eager-loading)

Joshfindit:
Excellent. Would it be reasonable to also go over the trickier usage as well? Like `person.blog_posts.with_associations(:tags, comments: :hashtags)` where we also want `comments: :authors`?

Joshfindit:
Excellent. Would it be reasonable to also go over the trickier usage as well? Like `person.blog_posts.with_associations(:tags, comments: :hashtags)` where we also want `comments: :authors`?

cheerfulstoic:
@Joshfindit I checked the tests and didn't see deeper examples, so I didn't want to document when I didn't know how well it would work.  Unfortunately I don't have time to test it out, but if you have time you can change the docs here: https://github.com/neo4jrb/neo4j/blob/master/docs/ActiveNode.rst

thefliik:
@Joshfindit `person.blog_posts.with_associations(:tags, comments: [:hashtags, :authors])`  will retrieve hashtags and authors

cheerfulstoic:
@Joshfindit I generally just watch the query log.  It should come up in the rails console or you can tail under `log/development.log`.  If you're not using Rails you can pass a `logger` option into the adaptor options using a `Logger` object.

PR #1404, opened 2017-08-17
Neo4j::Session.current.query changed

nossila:
Neo4j::Session.current.query changed to Neo4j::ActiveBase.current_session.query like its mentioned in Upgrade section

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/docs/Querying.rst
+++ b/docs/Querying.rst
@@ -257,7 +257,7 @@ The ``neo4j-core`` gem provides a ``Query`` class which can be used for building
 
 .. code-block:: ruby
 
-  Neo4j::Session.current.query # Get a new Query object
+  Neo4j::ActiveBase.current_session.query # Get a new Query object
 
   # Get a Query object based on a scope
   Student.query_as(:s) # For a


thefliik:
As long as it's being updated, referencing `Neo4j::ActiveBase.query` is probably more useful for the average `Neo4j` user (Which itself is an alias for `Neo4j::ActiveBase.current_transaction_or_session.query`).

cheerfulstoic:
Good points on both counts.  Going to merge and then change to @thefliik 's suggestion.  Thanks!

cheerfulstoic:
Also, if you want to get a `Query` object (which is what the docs are describing), you should call `Neo4j::ActiveBase.new_query` because `Neo4j::ActiveBase.query` is for submitting a string and parameters.  You can use `Neo4j::ActiveBase.magic_query` to do either, depending on arguments

thefliik:
@cheerfulstoic ah, I was wondering what made `magic_query` unique

cheerfulstoic:
MAGIC!

thefliik:
LOL

PR #1411, opened 2017-09-17
Allow enums to be created without associated indexes

### LINKED ISSUE #1259: Enums should suggest adding an index on initialization
In neo4j 8.0, enums should not try to define an index anymore. Instead, there should be a warning after defining an enum that suggest to setup an index using migrations.

Or maybe nothing at all, like AR does.
### Runtime information:

Neo4j database version: 3.0.3
`neo4j` gem version: 8.0.0
`neo4j-core` gem version: 7.0.0

thefliik:
As far as I can tell, the `_index` option for enums doesn't currently do anything. I'm using it to suppress the `Neo4j::ModelSchema.add_required_index` exception, which IMO is expected behavior.

Fixes #1259

This pull introduces/changes:
 * ` enum _index: false` now suppresses `Neo4j::ModelSchema.add_required_index`

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/lib/neo4j/active_node/enum.rb
+++ b/lib/neo4j/active_node/enum.rb
@@ -8,7 +8,7 @@ module Neo4j::ActiveNode
 
       def define_property(property_name, *args)
         super
-        Neo4j::ModelSchema.add_required_index(self, property_name)
+        Neo4j::ModelSchema.add_required_index(self, property_name) unless args[1][:_index] == false
       end
 
       def define_enum_methods(property_name, enum_keys, options)


thefliik:
One potential issue with this change is if `Enum::ClassMethods#define_property` is expected to have a non-nil return value, as it's return will be `nil` unless `args[1][:_index] == false`.

cheerfulstoic:
I think the return value is fine because this seems to be the only place where `define_property` is called:

https://github.com/neo4jrb/neo4j/blob/2b7caed832b0db7ff7b578fd48d5066d17350043/lib/neo4j/shared/enum.rb#L49

Since that's not the last line, the return value should be ignored.

I like this change, but I don't like the `_index` because it makes it feel like this option is secret or something which might change later.  Happy to throw this in if it were something like `require_index`

Thanks!

thefliik:
@cheerfulstoic, I can change it to `require_index` later today or tomorrow. Just to be clear though, [`_index` is already declared in the docs](http://neo4jrb.readthedocs.io/en/8.2.x/ActiveNode.html#enums) as a valid option for `enum`(it just appears to currently be useless) and the `_` formatting matches the format of the other options that can be passed to `enum`. Still want the name changed?

> By default, every enum property will be defined as unique, to improve query performances. If you want to disable this, simply pass _index: false to enum

thefliik:
And, if you do want the name changed to `require_index`, should I remove the `_index` option? (also, it seems like the name of the index option should follow the conventions established by the other options).

cheerfulstoic:
Oh, right, I had forgotten about that.  Looking back, it looks like this was because [this is how ActiveRecord does it](https://github.com/neo4jrb/neo4j/pull/1129#r52731033).

I will happily merge this, then.  Thanks again!

cheerfulstoic:
Released as 8.2.2

PR #1414, opened 2017-09-18
add: additional_mapped_label_names option to ActiveNode

jorroll:
## New feature:
This pull adds an ActiveNode class method `self.additional_mapped_label_names=` that maps additional labels to a model.

### Before this patch
The only way you can add multiple labels to a model is through inheritance or with module mixins (though I didn't actually realize the module option existed until I was writing the specs for this pull request). 

### After this patch
```
class User
  include Neo4j::ActiveNode
  self.additional_mapped_label_names = ['Author', 'Admin']

  class Person < User
  end
end

user = User.create
user.labels
#=> [:User, :Author, :Admin]

person = Person.create
person.labels
#=> [:Person, :User, :Author, :Admin]
```

There are multiple ways a feature like this could be implemented. The way I've implemented it leaves `mapped_label_name`, and the existing DSL, untouched. 

Each model still needs a single `mapped_label_name` and the `mapped_label_name` requires a constraint. `additional_mapped_label_names` are in addition to the `mapped_label_name`, and they do not require constraints. 

### The need
In my app, I have several models that need multiple labels such as `Role::Template::Current` (which has labels `Role`, `Current`, `Role::Current`, `Role::Template`, `Role::Template::Current`. Currently, I need to jump through some hoops to get there...

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/lib/neo4j/active_node/labels.rb
+++ b/lib/neo4j/active_node/labels.rb
@@ -131,12 +131,24 @@ module Neo4j
 
         # @return [Array{Symbol}] all the labels that this class has
         def mapped_label_names
-          self.ancestors.find_all { |a| a.respond_to?(:mapped_label_name) }.map { |a| a.mapped_label_name.to_sym }
+          self.ancestors.find_all { |a| a.respond_to?(:mapped_label_name) }.inject([]) do |labels, a|
+            labels.push *a.all_mapped_label_names.map(&:to_sym)
+          end
         end
 
         # @return [Symbol] the label that this class has which corresponds to a Ruby class
         def mapped_label_name
           @mapped_label_name || label_for_model
+        end
+
+        # @return [Array{Symbol}] the additional, user given labels that this class has
+        def additional_mapped_label_names
+          @additional_mapped_label_names || []
+        end
+
+        # @return [Array{Symbol}] the mapped_label_name and additional_mapped_label_names of this class
+        def all_mapped_label_names
+          [mapped_label_name, additional_mapped_label_names].flatten
         end
 
         # @return [Neo4j::Label] the label for this class
@@ -164,6 +176,17 @@ module Neo4j
 
         def mapped_label_name=(name)
           @mapped_label_name = name.to_sym
+        end
+
+        def additional_mapped_label_names=(given_names)
+          case given_names
+          when Array then names = given_names
+          when String then names = [given_names]
+          else
+            fail '"additional_mapped_label_names" must be a string or array of strings'
+          end
+          
+          @additional_mapped_label_names = names.map(&:to_sym)
         end
 
         # rubocop:disable Style/AccessorMethodName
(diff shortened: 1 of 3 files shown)


thefliik:
I figure folks might have some opinions on if / how this feature is implemented. If this pull is accepted, I can update the docs to reference the new option.

lib/neo4j/active_node/labels.rb
```diff
@@ -166,6 +178,17 @@ def mapped_label_name=(name)
           @mapped_label_name = name.to_sym
         end
 
+        def additional_mapped_label_names=(given_names)
+          case given_names
+          when Array then names = given_names
+          when String then names = [given_names]
+          else
+            fail '"additional_mapped_label_names" must be a string or array of strings'
+          end
```
cheerfulstoic: I think you could do:

```ruby
names = case given_names
        when Array then given_names
        when String then [given_names]
        else
          fail '"additional_mapped_label_names" must be a string or array of strings'
        end
```

thefliik:
@cheerfulstoic, updated formatting

thefliik:
My thinking with `additional_mapped_labels_for_names` is to basically treat them like inherited labels. The `mapped_label_name` is the, theoretically unique, label which identifies a particular model. The `additional_mapped_label_names` only come into play at the times when inherited `mapped_label_name`s would.

thefliik:
While this PR still seems like a valuable feature to have, I think #1424 provides a better solution to my particular problem, and would largely negate my need for this.

leehericks:
The inheritance-based label system currently in place is very understandable. It also allows for a separation of logic code, as an Admin class can inherit from User and have admin specific logic kept separate from the basic User class.

```
class User
  include Neo4j::ActiveNode
  self.additional_mapped_label_names = ['Author', 'Admin']

  class Person < User
  end
end
```

I understand it's example code, but when are all users by default Authors and Admins?

So I agree with Brian that some other way of implementing cross-cutting concerns should be investigated because inheritance isn't reusable.

Neo4j is pretty flexible in how you choose to model the data. I'm not an authoritative source, but I find [this GraphAware article](https://graphaware.com/neo4j/2015/01/16/neo4j-graph-model-design-labels-versus-indexed-properties.html) to be an interesting look at the use of multiple labels.

> In my app, I have several models that need multiple labels such as Role::Template::Current (which has labels Role, Current, Role::Current, Role::Template, Role::Template::Current.

Without more information, I can't fully understand why you are relying on so many combination of labels logically. The above article shows incredible performance hits for matching multiple labels or using negation.

All that said, labels are indeed an integral part of the the Neo4j property graph model. If possible, there should be a labels property and the ability to add labels and call save to persist them. Then before a save you could apply labels based on some logic, you could also mix in a module with shared logic based on labels?

cheerfulstoic:
Thanks @leehericks  Definitely check out #1424 because we have conversation going on about some possibilities over there.  I'd be curious to know what you think about the discussion / proposals

PR #1416, opened 2017-09-19
add: Vagrantfile to make contributing easier

thefliik:
I'm not sure if this is something folks want added to the project, but seeing as I already set up a Vagrantfile for myself...

This pull introduces/changes:
 * adds a Vagrantfile to make setting up a dev environment easier for contributors

After this change, new folks can simply clone the repo, install vagrant, cd into the root directory, run `vagrant up`, and `vagrant ssh` to get started. Installation and configuration of all gems as well as Neo4j (community-latest) is handled by vagrant.

See `CONTRIBUTING.mb` for more information

Pings:
@cheerfulstoic
@subvertallchris


thefliik:
Just learned that Neo4jrb doesn't support Neo4j 3.2.x so, at minimum, I'll need to update the Vagrantfile to use community-3.1.4. I'll wait for feedback before making any changes though.

cheerfulstoic:
This is great, thanks!

Personally since I use `tmux`, I set up a [config file](https://gist.github.com/cheerfulstoic/c2bb5c4a1eb5e8c19c570d0da151c7a5) with [https://github.com/tmuxinator/tmuxinator](https://github.com/tmuxinator/tmuxinator), which is kind of portable but it requires you to have a database set up in your main folder which has specific strings inside the configuration so that it can copy the DB twice and change it for `neo4j` and `neo4j-core`.  I've since learned a lot more about Docker and I'm wondering if it wouldn't be pretty easy to have a `docker-compose.yml` file to set these things up for the projects that could be combined with a `tmuxinator` file if somebody desired to use `tmux`.

I only used Vagrant once before (with a `Vagrantfile` that was already set up).  Do you have any options on Vagrant vs Docker?  If I were to use Docker I would probably set up two containers for two Neo4j DBs and two containers to run specs rubocop.

thefliik:
I'm not familiar with `tmux`, so I just [googled it](https://www.sitepoint.com/tmux-a-simple-start/). From what I can tell, it's solving a different problem than vagrant / docker would. It appears to just be supercharging / configuring your terminal window, rather than actually installing / configuring an environment?

Docker could definitely be adopted instead of vagrant. Before adopting vagrant, I made a halfhearted attempt to learn docker compose. I found it to be more difficult, which, combined with the fact that my computer is (apparently) too old to run docker compose, pushed me towards vagrant. If docker was already configured however, that would probably solve the first problem and I imagine most people won't have the second problem.

For its part, Vagrant just automatically installs a VirtualBox VM running headless ubuntu, ssh's into it, and runs the terminal commands specified in the Vagrantfile to install java, rvm, run bundler, install neo4j, etc. It then sym links the project folder on the host machine to the project folder in the virtual machine, so edits on the host machine are edits in the VM. Then, running `vagrant ssh` from the project's root ssh's into the VM where you can execute all commands normally within the virtual machine.

Vagrantfile
```diff
@@ -0,0 +1,157 @@
...
+# backwards compatibility). Please don't change it unless you know what
+# you're doing.
+Vagrant.configure("2") do |config|
+  # The most common configuration options are documented and commented below.
+  # For a complete reference, please see the online documentation at
+  # https://docs.vagrantup.com.
+
+  # Every Vagrant development environment requires a box. You can search for
+  # boxes at https://atlas.hashicorp.com/search.
+  config.vm.box = "ubuntu/xenial64"
```
thefliik: The base ubuntu installation to use for the VM

Vagrantfile
```diff
@@ -0,0 +1,157 @@
...
+
+  # Create a forwarded port mapping which allows access to a specific port
+  # within the machine from a port on the host machine and only allow access
+  # via 127.0.0.1 to disable public access
+  config.vm.network "forwarded_port", guest: 7474, host: 7474, host_ip: "127.0.0.1"
+  config.vm.network "forwarded_port", guest: 7473, host: 7473, host_ip: "127.0.0.1"
+  config.vm.network "forwarded_port", guest: 7472, host: 7472, host_ip: "127.0.0.1"
```
thefliik: This config links a specific port on the VM to a port on the host. Because Neo4j will be running inside a VM, ordinarily visiting `localhost:7474` in your browser would produce nothing. But this config forwards the VM 7474 port to the host machine allowing you to visit `localhost:7474` in your browser to get the neo4j server app.

By default though, Neo4j rejects non-local connections. And the port forwarding is a non-local connection. To get around this, I update neo4j.conf below to accept non-local connections.

Vagrantfile
```diff
@@ -0,0 +1,157 @@
...
+
+  # Share an additional folder to the guest VM or configure . The first argument is
+  # the path on the host to the actual folder. The second argument is
+  # the path on the guest to mount the folder. And the optional third
+  # argument is a set of non-required options.
+
+  project_name = "neo4j"
+  fail "Vagrantfile project_name variable cannot be blank" if project_name.nil? || project_name == ""
+
+  config.vm.synced_folder ".", "/home/ubuntu/#{project_name}"
```
thefliik: On the host machine, the project folder could be located anywhere. Inside the VM, the project folder is located here. This config sym links the two.

Vagrantfile
```diff
@@ -0,0 +1,157 @@
...
+  #   # Customize the amount of memory on the VM:
+  #   vb.memory = "2048"
+  # end
+  #
+  # View the documentation for the provider you are using for more
+  # information on available options.
+
+  # Disable notify-forwarder plugin, if present, because it was messing with Neo4j and causing a database panic
+  # https://discuss.elastic.co/t/es-cluster-in-docker-containers-alreadyclosedexception-underlying-file-changed-by-an-external-force/48874/8
+  if Vagrant.has_plugin?("vagrant-notify-forwarder")
```
thefliik: vagrant-notify-forwarder is a useful plugin that vagrant users may already have installed. It enables things like `rails s`'s automatic reload on file changes. Unfortunately, it doesn't play nice with neo4j and causes deadlocks, so it has been disabled (if it exists).

Vagrantfile
```diff
@@ -0,0 +1,157 @@
...
+    config.notify_forwarder.enable = false
+  end
+
+  # Enable provisioning with a shell script. Additional provisioners such as
+  # Puppet, Chef, Ansible, Salt, and Docker are also available. Please see the
+  # documentation for more information about their specific syntax and use.
+
+  config.vm.provision "bootstrapping", type: "shell", privileged: false,
+    keep_color: true, inline: <<-SHELL
```
thefliik: Provisioning configs are where Vagrant runs your custom setup commands. Vagrant supports many different provisioners (puppet, chef, etc), but I'm just using shell commands (`type: "shell"`), without sudo (`privilaged: false`), and I want the output to be colored (`keep_color: true`). The actual commands are given in the `inline:` string, and they're just standard shell commands that you would execute in your own terminal (on a machine running Ubuntu).

Vagrantfile
```diff
@@ -0,0 +1,157 @@
...
+      # also install libcurl3, libcurl3-gnutls, & libcurl4-openssl-dev
+      # which are needed by the "curb" gem, used in some 3rd party rake tasks--tho not actually sure about neo4j
+      yes | sudo apt-get install libcurl3 libcurl4-openssl-dev
+      bundle install
+      gem install overcommit --no-rdoc --no-ri
+      overcommit --install
+      echo "${GREEN}COMPLETED INSTALLING GEMS${NC}"
+  SHELL
+
+  unless Dir["./db/neo4j/development"].any?
```
thefliik: If the project folder doesn't already have neo4j installed `db/neo4j/development`, run the neo4j installation scripts.

Vagrantfile
```diff
@@ -0,0 +1,157 @@
...
+    # With default configuration, Neo4j only accepts local connections (which, in this case, means local to the VM).
+    # Vagrant has been set up to forward connections local to the VM to the appropriate ports on the host machine.
+    # As such, we need to change neo4j's conf to accept non-local connections if we want to be able to connect to
+    # neo4j from a browser on the host machine:
+    update_neo4j_conf = <<-TEXT
+      require 'fileutils'
+      require 'tempfile'
+
+      t_file = Tempfile.new('neo4j.conf')
+      File.open('./db/neo4j/development/conf/neo4j.conf', 'r+').each do |line|
```
thefliik: edit the neo4j.conf file to accept non-local connections (necessary for port forwarding)

Vagrantfile
```diff
@@ -0,0 +1,157 @@
...
+        echo "${PURPLE}INSTALLING NEO4J INTO 'db/neo4j/development'${NC}"
+        cd ~/#{project_name}
+        rake neo4j:install[community-latest]
+        ruby -e "#{update_neo4j_conf}"
+        echo "${GREEN}COMPLETED INSTALLING NEO4J INTO 'db/neo4j/development'${NC}"
+    SHELL
+  end
+
+  config.vm.provision "bootstrapping complete", type: "shell", privileged: false,
+    keep_color: true, run: "always", inline: <<-SHELL
```
thefliik: Finally, output a VM start up message. The `run: "always"` option tells vagrant to run these shell scripts each time `vagrant up` is used to start the VM. If `run: "always"` isn't included, vagrant will only run the commands the first time you `vagrant up` as part of the initial provisioning process.

Vagrantfile
```diff
@@ -0,0 +1,157 @@
...
+  # end
+  #
+  # View the documentation for the provider you are using for more
+  # information on available options.
+
+  # Disable notify-forwarder plugin, if present, because it was messing with Neo4j and causing a database panic
+  # https://discuss.elastic.co/t/es-cluster-in-docker-containers-alreadyclosedexception-underlying-file-changed-by-an-external-force/48874/8
+  if Vagrant.has_plugin?("vagrant-notify-forwarder")
+    config.notify_forwarder.enable = false
+  end
```
thefliik: vagrant-notify-forwarder is a useful plugin that vagrant users may already have installed. It enables things like `rails s`'s automatic reload on file changes. Unfortunately, it doesn't play nice with neo4j and causes exceptions, so it has been disabled (if it exists).

Vagrantfile
```diff
@@ -0,0 +1,157 @@
...
+    # As such, we need to change neo4j's conf to accept non-local connections if we want to be able to connect to
+    # neo4j from a browser on the host machine:
+    update_neo4j_conf = <<-TEXT
+      require 'fileutils'
+      require 'tempfile'
+
+      t_file = Tempfile.new('neo4j.conf')
+      File.open('./db/neo4j/development/conf/neo4j.conf', 'r+').each do |line|
+        case line.chomp
+        when '#dbms.connectors.default_listen_address=0.0.0.0'
```
thefliik: edit the neo4j.conf file to accept non-local connections (necessary for port forwarding)

cheerfulstoic:
Yep, you're right about tmux.  You could set up tmux to open up various things in either vagrant or docker (or something else).  Since I'm a lot more comfortable with docker I created some `Dockerfile`s in `neo4j` and `neo4j-core` and updated my [tmuxinator script](https://gist.github.com/cheerfulstoic/c2bb5c4a1eb5e8c19c570d0da151c7a5).  Even though you can't run docker compose, you could try running those individual commands (`tmuxinator` is also pretty great because I can have my `vim` pane on the left while `rspec` and `rubocop` run on the right.  I also have a [tmux configuration](https://gist.github.com/cheerfulstoic/b4d31e5e5750769376b36a786b700c0f) which allows me to move panes around with shift-arrow keys and so that I can use my mouse to change windows or move things around.

All of that said, I'm happy to merge this to provide another option for anybody interested in collaborating.  I'll work up some text to describe the docker side too.  Thanks!

PR #1419, opened 2017-09-20
breaking change: enums case insensitive by default

### LINKED ISSUE #1418: feature: case-insensitive enums
In my app, I'd like clients to provide enum params in all caps ([a convention](https://stackoverflow.com/a/3069863/5490505) I think is fairly common?). Currently, in order to accomplish this most easily, I'd need to have `enum property: [:VALUE_ONE, :VALUE_TWO]` which results in methods `Model.VALUE_ONE` or `Model.prefix_VALUE_ONE` which doesn't feel very Ruby esque. 

Would you accept a pull request to add a new option to `enum` of `_case_insensitive: true`? If enabled, this would downcase params before matching them against possible enum values. It would allow `enum property: [:value_one, :value_two], _case_insensitive: true` to match against `Model.create(property: 'VALUE_TWO')`.

Looks like this change would involve updates to [EnumConverter](https://github.com/neo4jrb/neo4j/blob/2b7caed832b0db7ff7b578fd48d5066d17350043/lib/neo4j/shared/type_converters.rb#L298) and ~[Neo4j::Shared::Enum::ClassMethods](https://github.com/neo4jrb/neo4j/blob/2b7caed832b0db7ff7b578fd48d5066d17350043/lib/neo4j/shared/enum.rb#L85)

thefliik:
Enums are now by default case insensitive. To make things easy for everyone, I've added a global config option `enums_case_sensitive` if people want to revert to the old standard. This being said, this is still a breaking change to Neo4jrb. I've also updated the docs.

Fixes #1418

This pull introduces/changes:
 * enums setters are now case insensitive by default 
 * to override this locally, you can pass `_case_sensitive: true` when defining an enum.
 * to override this default globally, you can set `config.neo4j.enums_case_sensitive = true`
 * unless the enum is case sensitive, it's keys must be downcased or an exception will be raised

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/docs/ActiveNode.rst
+++ b/docs/ActiveNode.rst
@@ -190,7 +190,7 @@ Or by using ``where``:
     Media.as(:m).where('m.type <> ?', Media.types[:image])
     # => CYPHER: "MATCH (result_media:`StoredFile`) WHERE (result_media.type <> 0)"
 
-By default, every ``enum`` property will be defined as ``unique``, to improve query performances. If you want to disable this, simply pass ``_index: false`` to ``enum``:
+By default, every ``enum`` property will require you to add an associated index to improve query performance. If you want to disable this, simply pass ``_index: false`` to ``enum``:
 
 .. code-block:: ruby
 
@@ -200,16 +200,24 @@ By default, every ``enum`` property will be defined as ``unique``, to improve qu
       enum type: [:image, :video, :unknown], _index: false
     end
 
-Sometimes it is desirable to have a default value for an ``enum`` property.  To acheive this, you can simply define a property with the same name which defines a default value:
-
+Sometimes it is desirable to have a default value for an ``enum`` property.  To acheive this, you can simply pass the `_default` option when defining the enum:
 
 .. code-block:: ruby
 
     class Media
       include Neo4j::ActiveNode
 
-      enum type: [:image, :video, :unknown]
-      property :type, default: :video
+      enum type: [:image, :video, :unknown], _default: :video
+    end
+
+By default, enum setters are case insensitive (in the example below, `Media.create(type: 'VIDEO').type == :video). If you wish to disable this for a specific enum, pass the `_case_sensitive: true` option. if you wish to change the global default for `_case_sensitive` to `true`, use Neo4jrb's `enums_case_sensitive` config option (detailed in the :ref:`configuration-variables` section).
+
+.. code-block:: ruby
+
+    class Media
+      include Neo4j::ActiveNode
+
+      enum type: [:image, :video, :unknown], _case_sensitive: false
     end
 
 .. _activenode-scopes:
--- a/docs/ActiveNode.rst
+++ b/docs/ActiveNode.rst
@@ -395,6 +395,13 @@ You can query associations:
     comment.post.comments       # Original comment and all of it's siblings.  Makes just one query
     post.comments.authors.posts # All posts of people who have commented on the post.  Still makes just one query
 
+When querying ``has_one`` associations, by default ``.first`` will be called on the result. This makes the result non-chainable if the result is ``nil``. If you want to ensure a chainable result, you can call ``has_one`` with a ``chainable: true`` argument.
+
+.. code-block:: ruby
+
+    comment.post                    # Post object
+    comment.post(chainable: true)   # Association proxy object wrapping post
+
 You can create associations
 
 .. code-block:: ruby
@@ -455,9 +462,9 @@ The two orphan-destruction options are unique to Neo4j.rb. As an example of when
 
 
 .. seealso::
-  :ref:`#has_many <Neo4j/ActiveNode/HasN/ClassMethods#has_many>`
+  #has_many http://www.rubydoc.info/gems/neo4j/Neo4j/ActiveNode/HasN/ClassMethods#has_many-instance_method
   and
-  :ref:`#has_one <Neo4j/ActiveNode/HasN/ClassMethods#has_one>`
+  #has_one http://www.rubydoc.info/gems/neo4j/Neo4j/ActiveNode/HasN/ClassMethods#has_one-instance_method
 
 
 Creating Unique Relationships
--- a/docs/ActiveNode.rst
+++ b/docs/ActiveNode.rst
@@ -404,6 +404,25 @@ You can create associations
 
     comment.post = post1                  # Removes all existing relationships
 
+Updating Associations
+~~~~~~~~~~~~~~~~~~~~~
+
+You can update attributes for objects of an association like this:
+
+.. code-block:: ruby
+
+    post.comments.update_all(flagged: true)
+    post.comments.where(text: /.*cats.*/).update_all(flagged: true)
+
+You can even update properties of the relationships for the associations like so:
+
+.. code-block:: ruby
+
+    post.comments.update_all_rels(flagged: true)
+    post.comments.where(text: /.*cats.*/).update_all_rels(flagged: true)
+    # Or to filter on the relationships
+    post.comments.where(flagged: nil).update_all_rels(flagged: true)
+
 Dependent Associations
 ~~~~~~~~~~~~~~~~~~~~~~
 
--- a/docs/ActiveNode.rst
+++ b/docs/ActiveNode.rst
@@ -200,7 +200,7 @@ By default, every ``enum`` property will require you to add an associated index 
       enum type: [:image, :video, :unknown], _index: false
     end
 
-Sometimes it is desirable to have a default value for an ``enum`` property.  To acheive this, you can simply pass the `_default` option when defining the enum:
+Sometimes it is desirable to have a default value for an ``enum`` property.  To acheive this, you can simply pass the ``_default`` option when defining the enum:
 
 .. code-block:: ruby
 
@@ -210,7 +210,7 @@ Sometimes it is desirable to have a default value for an ``enum`` property.  To 
       enum type: [:image, :video, :unknown], _default: :video
     end
 
-By default, enum setters are case insensitive (in the example below, `Media.create(type: 'VIDEO').type == :video). If you wish to disable this for a specific enum, pass the `_case_sensitive: true` option. if you wish to change the global default for `_case_sensitive` to `true`, use Neo4jrb's `enums_case_sensitive` config option (detailed in the :ref:`configuration-variables` section).
+By default, enum setters are `case insensitive` (in the example below, ``Media.create(type: 'VIDEO').type == :video``). If you wish to disable this for a specific enum, pass the ``_case_sensitive: true`` option. if you wish to change the global default for ``_case_sensitive`` to ``true``, use Neo4jrb's ``enums_case_sensitive`` config option (detailed in the :ref:`configuration-variables` section).
 
 .. code-block:: ruby
 
--- a/docs/Configuration.rst
+++ b/docs/Configuration.rst
@@ -21,16 +21,16 @@ Variables
   **association_model_namespace**
     **Default:** ``nil``
 
-    Associations defined in node models will try to match association names to classes. For example, `has_many :out, :student` will look for a `Student` class. To avoid having to use `model_class: 'MyModule::Student'`, this config option lets you specify the module that should be used globally for class name discovery.
+    Associations defined in node models will try to match association names to classes. For example, ``has_many :out, :student`` will look for a ``Student`` class. To avoid having to use ``model_class: 'MyModule::Student'``, this config option lets you specify the module that should be used globally for class name discovery.
 
-    Of course, even with this option set, you can always override it by calling `model_class: 'ClassName'`.
+    Of course, even with this option set, you can always override it by calling ``model_class: 'ClassName'``.
 
   **class_name_property**
     **Default:** ``:_classname``
 
-    Which property should be used to determine the `ActiveNode` class to wrap the node in
+    Which property should be used to determine the ``ActiveNode`` class to wrap the node in
 
-    If there is no value for this property on a node the node`s labels will be used to determine the `ActiveNode` class
+    If there is no value for this property on a node the node`s labels will be used to determine the ``ActiveNode`` class
 
     .. seealso:: :ref:`activenode-wrapping`
 
--- a/docs/Configuration.rst
+++ b/docs/Configuration.rst
@@ -18,12 +18,12 @@ Variables
 
 .. glossary::
 
-  **skip_migration_check**
-    **Default:** ``false``
+  **association_model_namespace**
+    **Default:** ``nil``
 
-    Prevents the ``neo4j`` gem from raising ``Neo4j::PendingMigrationError`` in web requests when migrations haven't been run.  For environments (like testing) where you need to use the ``neo4j:schema:load`` rake task to build the database instead of migrations.  Automatically set to ``true`` in Rails test environments by default
+    Associations defined in node models will try to match association names to classes. For example, `has_many :out, :student` will look for a `Student` class. To avoid having to use `model_class: 'MyModule::Student'`, this config option lets you specify the module that should be used globally for class name discovery.
 
-  .. _configuration-class_name_property:
+    Of course, even with this option set, you can always override it by calling `model_class: 'ClassName'`.
 
   **class_name_property**
     **Default:** ``:_classname``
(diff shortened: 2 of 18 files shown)


docs/Configuration.rst
```diff
@@ -18,40 +18,40 @@ Variables
 
 .. glossary::
 
-  **skip_migration_check**
-    **Default:** ``false``
+  **association_model_namespace**
```
thefliik: I took a moment to rearrange the Configuration::Variables section of the docs to be in alphabetical order

docs/ActiveNode.rst
```diff
@@ -190,7 +190,7 @@ Or by using ``where``:
     Media.as(:m).where('m.type <> ?', Media.types[:image])
     # => CYPHER: "MATCH (result_media:`StoredFile`) WHERE (result_media.type <> 0)"
 
-By default, every ``enum`` property will be defined as ``unique``, to improve query performances. If you want to disable this, simply pass ``_index: false`` to ``enum``:
+By default, every ``enum`` property will require you to add an associated index to improve query performance. If you want to disable this, simply pass ``_index: false`` to ``enum``:
```
thefliik: Also updated the enum `_index` option docs

docs/ActiveNode.rst
```diff
@@ -200,16 +200,24 @@ By default, every ``enum`` property will be defined as ``unique``, to improve qu
       enum type: [:image, :video, :unknown], _index: false
     end
 
-Sometimes it is desirable to have a default value for an ``enum`` property.  To acheive this, you can simply define a property with the same name which defines a default value:
+Sometimes it is desirable to have a default value for an ``enum`` property.  To acheive this, you can simply pass the ``_default`` option when defining the enum:
```
thefliik: this was outdated

brandon-zeeb-nw:
Is someone able to describe the use case for this change beyond it being "nice to do"?

thefliik:
@brandon-zeeb-nw, in languages with an enum type, [convention is for enum variables](https://stackoverflow.com/a/3069863/5490505) be given in screaming case (which is case insensitive / only one case). Currently, if you want your neo4jrb enums to accept screaming case params, you need to define the enums in screaming case `enum property: [:OPTION_ONE, :OPTION_TWO]` which produces related methods in the form `Model.OPTION_ONE` or `Model.prefix_OPTION_ONE` which break ruby conventions.

This change allows you to define `enum, property: [:option_one, :option_two]`, still accept `Model.create(property: 'OPTION_ONE')` or `Model.create(property: 'option_one')`, and interact with it `Model.prefix_option_one`. Given that the property will be persisted to the DB as an Integer, the capitalization of the name really shouldn't matter unless you have options that are only differentiated by capitalization--which I imagine is an unusual use case.

For folks who reject this change, a global config is available for them to change the default back to case sensitive.

Personally, I would describe the current behavior of allowing case sensitive enums to be a bug.

thefliik:
@brandon-zeeb-nw do you have a use case for case sensitive enums?

brandon-zeeb-nw:
thanks @thefliik, that helps.  I fully support the idea of sticking to conventions, but I'm not sure this is the way.  This feels more like a mapping problem to me, to map the ruby conventions into a normalized form and back.  Case insensitivity is a way around it, Imo.  That said I do fully support the spirit of the change.

thefliik:
I thought about simply normalizing the method names and not the enums themselves, but enums should be case insensitive, per convention (well, all caps). This strategy encourages developers to use enums that fall in line with other languages--especially useful if 3rd party clients will be sending the params and you can't guarantee their formatting.

brandon-zeeb-nw:
@thefliik I had a somewhat incomplete view of the angle on this change, I retract my objection, I think your change is pretty legit.  Thanks for working with me.

thefliik:
@brandon-zeeb-nw for sure!

lib/neo4j/shared/enum.rb
```diff
@@ -53,18 +53,29 @@ def enum(parameters = {})
 
       protected
 
-      def normalize_key_list(enum_keys)
+      def normalize_key_list(enum_keys, options)
+        case_sensitive = options[:_case_sensitive]
+        case_sensitive = Neo4j::Config.enums_case_sensitive if case_sensitive.nil?
```
cheerfulstoic: You could do

```ruby
case_sensitive = options[:_case_sensitive] || Neo4j::Config.enums_case_sensitive
```

or

```ruby
options.fetch(:_case_sensitive, Neo4j::Config.enums_case_sensitive)
```

cheerfulstoic: The difference between the two is that the first would fall back to `Neo4j::Config.enums_case_sensitive` if you get a `nil` value from `options[:_case_sensitive]` regardless of if there was no key or if there was a key but it just happened to be `nil`.  In the second one if `options[:_case_sensitive]` is set to `nil` then the result will be `nil` because `fetch` only falls back to the default when the key isn't set in the hash.

lib/neo4j/shared/enum.rb
```diff
@@ -91,6 +101,12 @@ def build_property_options(_enum_keys, options = {})
         }
       end
 
+      def build_enum_options(enum_keys, options = {})
+        enum_options = build_property_options(enum_keys, options)
+        enum_options[:case_sensitive] = options[:_case_sensitive]
+        enum_options
```
cheerfulstoic: Maybe a bit cleaner:

```ruby
build_property_options(enum_keys, options).tap do |enum_options|
  enum_options[:case_sensitive] = options[:_case_sensitive]
end

File: spec/e2e/enum_spec.rb [hunk 6]
```diff
@@ -32,10 +33,15 @@
       expect(StoredFile.types).to eq(unknown: 0, image: 1, video: 2)
       expect(StoredFile.sizes).to eq(big: 100, medium: 7, small: 2)
       expect(StoredFile.flags).to eq(clean: 0, dangerous: 1)
+      expect(StoredFile.type_formats).to eq(Mpeg: 0, Png: 1)
       expect(UploaderRel.origins).to eq(disk: 0, web: 1)
     end
   end
```
cheerfulstoic: There should be a spec somewhere is this file which tests that the `ArgumentError` gets raised when a non-lowercase option is passed to `enum`.  Probably you can dynamically, in the spec, call `StoredFile.enum something: [:value1, :Value1]`

cheerfulstoic:
Oh, also!  The build is failing because `rubocop` is failing:

https://travis-ci.org/neo4jrb/neo4j/builds/278210221?utm_source=github_status&utm_medium=notification

If you run `rubocop -aD` locally it will fix any rubocop issues that it can automagically and tell you about the ones that it couldn't fix.  They have pretty good [documentation](http://rubocop.readthedocs.io/en/latest/) on what the different rules are about, but let me know if you have any questions or need any help.

lib/neo4j/shared/enum.rb
thefliik: I'll update it to use `fetch`, the first one doesn't quite work because if `options[:_case_sensitive] = false` and `Neo4j::Config.enums_case_sensitive = true`, then `case_sensitive = true`, when it should equal `false`.

thefliik:
Ya, I have `overcommit` installed locally so I was under the impression that `rubocop` was automatically run on commit? It's certainly stopped a few commits because of whitespace errors. This commit ran fine locally though, so I was a bit surprised to see the Travis failure. Or does `overcommit` run a stripped down version of the `rubocop` checks?

docs/ActiveNode.rst
thefliik: You're totally right. A `_default` option is whitelisted in the source and I assumed it worked. Just testing though, I see it does not work. I'll change it back to the old documentation.

thefliik: Wait no, the `_default` option does work. There's just a bug that if someone submits an enum param that isn't among the list of options, then the record is persisted with the first option. It should either be persisted with the `_default` option or it should raise an exception. Do you have an opinion either way?

thefliik: Also, regarding `_default` vs `default`, personally, I don't really like the idea of accepting both. More code would need to be added to make sure only one option was given and it feels like clutter. Given that this is already a breaking change, now might be the time to remove `_` from all these options. 

This being said, I'm happy to do it your way if that's what you decide. It would probably involve updating `normalize_key_list` to both normalize the new keys and spit out depreciation warnings.

thefliik: So looking into persistence a bit more. Currently, if a `_default` option isn't set, and a submitted value doesn't match a permitted value, then the node will be persisted with a value of `0`. This is a **bug** if someone has defined their enums using a hash like `enum property: {one: 'ONE', two: 'TWO'}`. If you have an enum like that, and execute `Model.create(property: 'TEST')`, then the object will be persisted with `property = 0` when the only allowed values are `'ONE'` or `'TWO'`. It's a little hard to diagnose because the `Neo4jrb` object shows `property = nil` (but the db has `property = 0`).

This is what I think the fix should be: 
1. If `_default` isn't specified, and a given param doesn't match one of the values in the enum set, an exception should be raised. It could be an `ArgumentError`, but since these happen for a variety of reasons, I'm thinking a special `InvalidEnumValueError` would be better (allowing someone wants to rescue that, specifically).
2. If `_default` is specified, and a given param doesn't match one of the keys in the enum set, then the `_default` value is used as if `Model.create(property: default_value)` had been called (i.e. `default_value` will be matched against the enum keys to determine what value should be persisted to the db).

cheerfulstoic: Yeah, fair enough about supporting both.  Let's stick with the underscore for now because it's what `ActiveRecord` does (even though it's really strange...)

Really good catch about it defaulting to `0` in all cases.  That's pretty ugly...

One thing that I thought of: if we have a `_default` we should raise an exception if that isn't in the list of `enum` options.

I like 1), though regarding 2) I think that if you specify a `_default` and you give a value that isn't in the list of enums it should raise an exception.

In the end, I think you're absolutely right.  If they give something that's not in the params list we should raise an exception.  I definitely like the idea of having an `InvalidEnumValueError` which inherits from `ArgumentError` so that users can account for this.

Just checking in really quickly (still working through the backlog of issues and PRs, don't worry ;) ).  Let me know if I missed anything here

cheerfulstoic:
That's certainly odd... `overcommit` should catch `rubocop` issues.  It basically just runs on the files that you have staged when you go to commit.  I think I've noticed that when I have multiple versions of `rubocop` on my system `overcommit` doesn't do a `bundle exec` and so it might pull in the wrong version (but that's a long shot guess)

thefliik:
Well I confirmed that `overcommit` is letting stuff through when I committed #1422. I first ran the commit. `overcommit` fixed some stuff then successfully committed. I then ran `rubocob` and ran into errors that I needed to fix.

It's definitely not a `rubocob` version problem, and this is running inside a VM containing nothing but the Neo4j repo.

docs/ActiveNode.rst
thefliik: Will do. And no rush! I've thrown a lot at you this week.

lib/neo4j/shared/enum.rb
```diff
@@ -3,6 +3,7 @@ module Enum
     extend ActiveSupport::Concern
 
     class ConflictingEnumMethodError < Neo4j::Error; end
+    class InvalidEnumValueError < Neo4j::InvalidParameterError; end
```
thefliik: I'm inheriting from `InvalidParameterError` rather than `ArgumentError`. When I suggested `ArgumentError`, I didn't know about `InvalidParameterError`. Seeing as we're talking about passing an invalid enum value through a param, this seemed more appropriate. Let me know if you prefer `ArgumentError` though.

cheerfulstoic:
I just resolved a conflict in the `Vagrantfile` that looked like this:

```
<<<<<<< case_insensitive_enums
  config.vm.network 'forwarded_port', guest: 7474, host: 7474, host_ip: '127.0.0.1'
  config.vm.network 'forwarded_port', guest: 7473, host: 7473, host_ip: '127.0.0.1'
  config.vm.network 'forwarded_port', guest: 7472, host: 7472, host_ip: '127.0.0.1'
=======
  config.vm.network 'forwarded_port', guest: 7474, host: 9474, host_ip: '127.0.0.1'
  config.vm.network 'forwarded_port', guest: 7473, host: 9473, host_ip: '127.0.0.1'
  config.vm.network 'forwarded_port', guest: 7472, host: 9472, host_ip: '127.0.0.1'
>>>>>>> master
```

I went with the `master` block, but I wanted to let you know in case that was wrong.

cheerfulstoic:
Let me know if you're done with this PR and I'll take another look!

thefliik:
> I just resolved a conflict in the Vagrantfile that looked like this:

Ooopss!! The change from `7474` --> `9474` etc wasn't suppose to be committed! I just changed that on my personal computer to avoid a port conflict with multiple instances of Neo4j.

lib/neo4j/shared/type_converters.rb
```diff
@@ -271,6 +271,10 @@ class EnumConverter
       def initialize(enum_keys, options)
         @enum_keys = enum_keys
         @options = options
+
+        return unless @options[:case_sensitive].nil?
+
+        @options[:case_sensitive] = Neo4j::Config.enums_case_sensitive
```
cheerfulstoic: Instead of these two lines, what about just:

```ruby
@options[:case_sensitive] ||= Neo4j::Config.enums_case_sensitive
```

spec/e2e/enum_spec.rb
```diff
@@ -68,6 +82,31 @@
...
+      expect(file.reload.type_format).to eq(:Png)
+    end
+
+    it 'respects global _case_sensitive = false default' do
+      file = StoredFile.new
+      file.type = :VIdeO
+      file.save!
+      expect(StoredFile.as(:f).pluck('f.type')).to eq([2])
+      expect(file.reload.type).to eq(:video)
+    end
```
cheerfulstoic: You could test the configuration by doing:

```ruby
context 'global enums_case_sensitive config is set to true' do
  let_config(:enums_case_sensitive) { true }
  # spec here testing, for example, that you can't set `type` to `VIdeO`
end
```

We define the `let_config` helper here:

https://github.com/neo4jrb/neo4j/blob/master/spec/spec_helper.rb#L135

spec/e2e/enum_spec.rb
```diff
@@ -68,6 +82,31 @@
...
+    it "raises error if value doesn't match an enum key" do
+      file = StoredFile.new
+      file.type = :audio
+      expect { file.save! }.to raise_error Neo4j::Shared::Enum::InvalidEnumValueError
+    end
+
+    it 'respects local _case_sensitive option' do
+      file = StoredFile.new
+      file.type_format = :png
+      expect { file.save! }.to raise_error(Neo4j::Shared::Enum::InvalidEnumValueError)
```
cheerfulstoic: For a stronger assertion this could be:

```ruby
expect { file.save! }.to raise_error(Neo4j::Shared::Enum::InvalidEnumValueError, 'Value passed to an enum property must match one of the enum keys')
```

spec/e2e/enum_spec.rb
```diff
@@ -68,6 +82,31 @@
       expect(StoredFile.as(:f).where(id: file.id).pluck('f.flag')).to eq([nil])
       expect(file.reload.flag).to eq(nil)
     end
+
+    it "raises error if value doesn't match an enum key" do
+      file = StoredFile.new
+      file.type = :audio
+      expect { file.save! }.to raise_error Neo4j::Shared::Enum::InvalidEnumValueError
```
cheerfulstoic: For a stronger assertion this could be:

```ruby
expect { file.save! }.to raise_error(Neo4j::Shared::Enum::InvalidEnumValueError, 'Case-insensitive (downcased) value passed to an enum property must match one of the enum keys')
```

lib/neo4j/shared/type_converters.rb
thefliik: This doesn't work because if `@options[:case_sensitive] == false` and `Neo4j::Config.enums_case_sensitive == true`, then `@options[:case_sensitive]` will be set to `true` when it should remain false (i.e. local settings should always take precedence over global settings). As it stands, global settings only kick in if `@options[:case_sensitive].nil?`.

spec/e2e/enum_spec.rb
thefliik: Oh thats cool

thefliik:
Ok, I think its done. Let me know if there's anything else though.

lib/neo4j/shared/type_converters.rb
cheerfulstoic: Ah, right, good point ;)

cheerfulstoic:
Excellent, thanks!

cheerfulstoic:
I believe that #1414 and #1424 (however we go with that) would be backwards-compatible, right?  If so I'll release this as 9.0.0 separately

thefliik:
Yes, both of those would be backwards-compatible. I'm not sure if #1428 is backwards compatible or not. Depends on whether the current behavior is considered a bug (though, even if its a bug, its so pervasive fixing it might be considered a breaking change).

PR #1422, opened 2017-09-22
Add: has_one(chainable: true) option

thefliik:
By default, `model.has_one` calls `.first` on the result. If the result is `nil`, this means the method is not chainable. I'd like to have the option of getting an association proxy back from a `has_one` getter. Having [dug into the source](https://github.com/neo4jrb/neo4j/blob/d5dc490349d24a6748f77fe445f0d2ff664bf8fd/lib/neo4j/active_node/has_n.rb#L472), I see that `has_one` methods actually take arguments and options, and one of those options (`rel_length`) can return a chainable result, but
1. the method name `rel_length` is not an intuitive argument to pass if you specifically want a chainable result.
2. `rel_length` must have a non-integer value to return an association proxy (e.g. `rel_length: ''`) which adds to the complexity / is ugly.

This pull introduces/changes:
 * adds a `chainable: true` option to `has_one` methods that returns an association proxy instead of an ActiveNode object

```
comment.post(chainable: true).author
```

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/docs/ActiveNode.rst
+++ b/docs/ActiveNode.rst
@@ -387,6 +387,13 @@ You can query associations:
     comment.post.comments       # Original comment and all of it's siblings.  Makes just one query
     post.comments.authors.posts # All posts of people who have commented on the post.  Still makes just one query
 
+When querying ``has_one`` associations, by default ``.first`` will be called on the result. This makes the result non-chainable if the result is ``nil``. If you want to ensure a chainable result, you can call ``has_one`` with a ``chainable: true`` argument.
+
+.. code-block:: ruby
+
+    comment.post                    # Post object
+    comment.post(chainable: true)   # Association proxy object wrapping post
+
 You can create associations
 
 .. code-block:: ruby
@@ -428,9 +435,9 @@ The two orphan-destruction options are unique to Neo4j.rb. As an example of when
 
 
 .. seealso::
-  :ref:`#has_many <Neo4j/ActiveNode/HasN/ClassMethods#has_many>`
+  #has_many http://www.rubydoc.info/gems/neo4j/Neo4j/ActiveNode/HasN/ClassMethods#has_many-instance_method
   and
-  :ref:`#has_one <Neo4j/ActiveNode/HasN/ClassMethods#has_one>`
+  #has_one http://www.rubydoc.info/gems/neo4j/Neo4j/ActiveNode/HasN/ClassMethods#has_one-instance_method
 
 
 Creating Unique Relationships
--- a/lib/neo4j/active_node/has_n.rb
+++ b/lib/neo4j/active_node/has_n.rb
@@ -467,9 +467,10 @@ module Neo4j::ActiveNode
         define_method(name) do |node = nil, rel = nil, options = {}|
           options, node = node, nil if node.is_a?(Hash)
 
-          # Return all results if a variable-length relationship length was given
           association_proxy = association_proxy(name, {node: node, rel: rel}.merge!(options))
-          if options[:rel_length] && !options[:rel_length].is_a?(Integer)
+
+          # Return all results if options[:chainable] == true or a variable-length relationship length was given
+          if options[:chainable] || (options[:rel_length] && !options[:rel_length].is_a?(Integer))
             association_proxy
           else
             target_class = self.class.send(:association_target_class, name)
--- a/lib/neo4j/active_node/has_n.rb
+++ b/lib/neo4j/active_node/has_n.rb
@@ -463,7 +463,7 @@ module Neo4j::ActiveNode
         end
       end
 
-      def define_has_one_getter(name)
+      def define_has_one_getter(name) # rubocop:disable Metrics/PerceivedComplexity
         define_method(name) do |node = nil, rel = nil, options = {}|
           options, node = node, nil if node.is_a?(Hash)
 
--- a/spec/e2e/has_one_spec.rb
+++ b/spec/e2e/has_one_spec.rb
@@ -24,6 +24,12 @@ describe 'has_one' do
       it 'returns a nil object' do
         expect(unsaved_node.parent).to eq nil
       end
+
+      describe 'with chainable: true option' do
+        it 'returns an empty association proxy object' do
+          expect(unsaved_node.parent(chainable: true).inspect).to eq "#<AssociationProxy HasOneB#parent []>"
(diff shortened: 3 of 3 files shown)


docs/ActiveNode.rst
```diff
@@ -428,9 +435,9 @@ The two orphan-destruction options are unique to Neo4j.rb. As an example of when
 
 
 .. seealso::
-  :ref:`#has_many <Neo4j/ActiveNode/HasN/ClassMethods#has_many>`
+  #has_many http://www.rubydoc.info/gems/neo4j/Neo4j/ActiveNode/HasN/ClassMethods#has_many-instance_method
   and
-  :ref:`#has_one <Neo4j/ActiveNode/HasN/ClassMethods#has_one>`
+  #has_one http://www.rubydoc.info/gems/neo4j/Neo4j/ActiveNode/HasN/ClassMethods#has_one-instance_method
```
thefliik: Currently, in the docs this `see also` block links to nothing. I've updated it with appropriate links to the API on rubydoc.

lib/neo4j/active_node/has_n.rb
```diff
@@ -463,13 +463,14 @@ def define_has_one_id_methods(name)
         end
       end
 
-      def define_has_one_getter(name)
+      def define_has_one_getter(name) # rubocop:disable Metrics/PerceivedComplexity
```
thefliik: Rubocop was failing this with PerceivedComplexity as too high. I didn't know what to do with that, so I suppressed the warning to make it pass 😕 ...

cheerfulstoic:
I could see the use for this, but what about:

```ruby
comment.association_proxy(:post).author
```

I think I like your solution better because it's a bit more obvious, but I wanted to mention that

cheerfulstoic:
We could even probably `alias` `association_proxy` as something like `start_chain`.  I'm good either with something like that or this PR

thefliik:
I find having the method name describe the function rather then the argument describe the function to be more readable and intuitive. The decision to have the result wrapped vs unwrapped seems to be fundamentally a param choice, rather than a method choice.

For example:
```
reply.comment(chainable: true).post(chainable: true).author(chainable: true).posts.comments
```
vs
```
reply.start_chain(:comment).start_chain(:post).start_chain(:author).posts.comments
```

thefliik:
Especially when I'm already conditioned by Neo4j to see an association chain as `model.path.path.path` if it suddenly became `model.method(:path).method(:path).method(:path)` it feels like its breaking convention.

cheerfulstoic:
Yeah, I actually agree a lot.  We actually have two options for specifying `OPTIONAL MATCH` for associations:

```ruby
object.association(optional: true)
object.optional(:association)
```

I prefer the first one as well.  I'll have a look through this PR this, thanks!

docs/ActiveNode.rst
cheerfulstoic: Thanks!  I had actually written some code a while back to turn the `YARD` documentation into `rst` files for readthedocs (you can see the `rst` formatting on rubydoc.info), but re-implementing rubydoc.info was probably largely a waste of my time ;) . I think these were links to that internal API documentation, but glad to have these fixed now

lib/neo4j/active_node/has_n.rb
cheerfulstoic: It could probably just do with an extraction of a method.  I'll fix it and link to the commit here

spec/e2e/has_one_spec.rb
```diff
@@ -24,6 +24,12 @@
       it 'returns a nil object' do
         expect(unsaved_node.parent).to eq nil
       end
+
+      describe 'with chainable: true option' do
+        it 'returns an empty association proxy object' do
+          expect(unsaved_node.parent(chainable: true).inspect).to eq '#<AssociationProxy HasOneB#parent []>'
```
cheerfulstoic: I think this would be a bit better as:

```
expect(unsaved_node.parent(chainable: true)).to be_a Neo4j::ActiveNode::HasN::AssociationProxy
```

I can make that change

cheerfulstoic:
Looks great.  Since this isn't a bugfix but a backwards compatible change I'll release this in `8.3.0`, though let's get #1414 and #1419 done and we can release them all together.

#1419 is indeed a breaking change, though I doubt it will affect anybody so I'm tempted to just put it into 8.3 (especially since we have good exceptions around it)

lib/neo4j/active_node/has_n.rb
cheerfulstoic: Meh, not the best: https://github.com/neo4jrb/neo4j/commit/1122ee7b302b1d1d7b29b56d313987ac95913abe

spec/e2e/has_one_spec.rb
cheerfulstoic: https://github.com/neo4jrb/neo4j/commit/a34cb4cc49cde56a73d5974315e45526356d9a76

thefliik:
I imagine the semantic versioning standard was created _specifically_ to prevent an update like this from being released as a minor point increase. Major updates pretty obviously demand someone's attention. I imagine the little updates are the ones that break something because no one expects them to. Definitely understand the temptation, but can't say I support it.

thefliik:
Yeah I get that. If you wanted, I think it would be possible to achieve both by creating a duel versioning scheme (e.g. Google Chrome does that). The semver version is what people would spec against. The branding(?) version is what people could look at to get a sense of how large the update is. In the readme you might put:

### Neo4jrb 3
**semver 13.4.3**

Of course you'd need to explain this.

And ya, large open source projects with paid maintainers will manage releases to limit the semver point increase, but I think its totally ok for volunteer projects to ignore that. I don't see anything wrong with the version blowing up to 20/30/40.x.x -- and if devs do have a problem with it, I imagine they'll give feedback.

At its core: ya, as a dev its annoying when a dependency demands your attention during an update and forces you to look at the changelog, but for breaking changes like this, people _really should_ look at the changelog.

This being said, you're doing a **kick-ass** job managing this repo as is, so ultimately, do what you gotta do.

cheerfulstoic:
Yup, thanks ;)  I think I'll just go with the simplest "upgrade whatever is appropriate" for each change for now, at least ;)

I just realized I didn't release this as 8.3.0 previous, so I just did that now

PR #1426, opened 2017-09-22
If the model isn't defined, fallback to the key

### LINKED ISSUE #1425: bug: pluck() fails with has_many() assoc thru another model
Tested in a new, blank app. Given the following models:

```
class Form 
  include Neo4j::ActiveNode

  has_many :out, :owner_edgs, type: :OWNER, model_class: "Owner"

  def owners
    owner_edgs.to_node
  end
end

class Owner 
  include Neo4j::ActiveNode

  has_one :in, :from_node, type: :OWNER, model_class: "Form"
  has_one :out, :to_node, type: :OWNER, model_class: ["Person", "Organization"]
end

class Person 
  include Neo4j::ActiveNode

  has_many :in, :owner_edgs, type: :OWNER, model_class: "Owner"
end
```

# Expected Result

```
form = Form.create
owner = Owner.create
person = Person.create
owner.from_node = form
owner.to_node = person

form.owners.pluck(:id)
#=> ["uidaf-olkcjn-aeklnc-dadf;lcea"] # pretend its a uuid
```

# Actual Result

```
form = Form.create
owner = Owner.create
person = Person.create
owner.from_node = form
owner.to_node = person

form.owners.pluck(:id)
#=> NoMethodError: undefined method `id_property_name' for nil:NilClass
```

# Additional Information

The query works normally if `class Owner` `has_one :out, :to_node, type: :OWNER, model_class: "Person"`

I can also work around the issue by calling `form.owners.to_a.pluck(:id)`

### Runtime information:

Neo4j database version: 3.1.5
`neo4j` gem version: 8.2.4
`neo4j-core` gem version: 7.2.3

cheerfulstoic:
Fixes #1425

This pull introduces/changes:
 * When there is an association with an array for the `model_class` argument, you can get an error when plucking `:id`.

### DIFF
--- a/lib/neo4j/active_node/query/query_proxy_link.rb
+++ b/lib/neo4j/active_node/query/query_proxy_link.rb
@@ -115,7 +115,11 @@ module Neo4j
             end
 
             def converted_key(model, key)
-              (model && key.to_sym == :id) ? model.id_property_name : key
+              if key.to_sym == :id
+                model ? model.id_property_name : :uuid
+              else
+                key
+              end
             end
 
             def converted_value(model, key, value)
--- a/lib/neo4j/active_node/query/query_proxy_link.rb
+++ b/lib/neo4j/active_node/query/query_proxy_link.rb
@@ -115,7 +115,7 @@ module Neo4j
             end
 
             def converted_key(model, key)
-              key.to_sym == :id ? model.id_property_name : key
+              (model && key.to_sym == :id) ? model.id_property_name : key
             end
 
             def converted_value(model, key, value)


lib/neo4j/active_node/query/query_proxy_link.rb
```diff
@@ -115,7 +115,11 @@ def converted_keys(model, arg)
             end
 
             def converted_key(model, key)
-              key.to_sym == :id ? model.id_property_name : key
+              if key.to_sym == :id
+                model ? model.id_property_name : :uuid
```
thefliik: Haven't actually tested this yet but ❗️ 
I imagine this solution will work for my use case, but won't this fail if someone has defined a custom `id_property`? I don't think there's anyway you can work around that case for everyone, as, as you said, you don't know what the model is so you don't know what the` id_property` is. I think this could be addressed by adding a warning in the docs to users of the custom `id_property` option.

This being said, if someone makes uniform use of a custom `id_property` in their app, then they should be able to use the `pluck(:id)` feature. The docs reference something about setting a custom `id_property` globally (though they don't appear to actually explain how). Maybe, instead of `:uuid`, point to the global config and have the default for that be `:uuid`.

thefliik:
This fixed the issue I was experiencing

lib/neo4j/active_node/query/query_proxy_link.rb
cheerfulstoic: Yeah, it could make sense to have a warning both under the documentation for `id_property` as well as the documentation for `model_class`.  It could be that we assume the `model` for the `QueryProxy` is the first model specified and for things like getting the `id_property` both would need to have the same ID property.  But for now I think that this will take care of almost all cases.

I'll add a spec and release it and it could be re-visited later as needed

cheerfulstoic:
Released as 8.2.5, thanks!

PR #1428, opened 2017-09-24
Fix / upd: properties declared on parent after inheritance takes place are now passed to children

thefliik:
I'm not sure if this is a bug or intended behavior, but currently: if an ActiveNode class `A` is inherited by `B`, and then an ActiveNode `property` is subsequently declared on `A`, `B` will not have that property (i.e. properties are only be transferred at inheritance).

This PR makes it so that every property declared on a parent is also present on the children (unless `undef_property` was explicitly called on a child).
 
In achieving this, I plucked `ActiveSupport/core_ext/class/subclasses` and added the `subclasses` method to `Class`.

This pull introduces/changes:
 * properties declared on parent classes after inheritance has taken place are properly inherited by subclasses
 * all classes now have `subclasses` method

### Example

```
class Person
  include Neo4j::ActiveNode

  class Superhero < Person
  end

  property :name
end

hero = Person::Superhero.new
hero.name = 'John'
hero.name
#=> 'John'
```

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/lib/neo4j.rb
+++ b/lib/neo4j.rb
@@ -11,6 +11,7 @@ require 'neo4j/active_base/session_registry'
 require 'active_model'
 require 'active_support/concern'
 require 'active_support/core_ext/class/attribute.rb'
+require 'active_support/core_ext/class/subclasses.rb'
 require 'json'
 
 require 'neo4j/errors'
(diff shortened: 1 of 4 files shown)


thefliik:
This PR has a minor conflict with #1419

thefliik:
I suppose this update has the capacity to break things... If it is something you're interested in accepting, might be a good idea to release it along with #1419

cheerfulstoic:
Thanks so much for this!  I would say this falls into the bugfix camp, but I'll release it in 9.0.0 just to make sure.  I resolved the conflict and now waiting for Travis to finish before merging (though I might not get back to it this evening)

thefliik:
awesome

cheerfulstoic:
9.0.0 released

thefliik:
Whoo hoo

PR #1435, opened 2017-09-26
Added note about global case sensitive option

thefliik:
included note about global case sensitive option

Pings:
@cheerfulstoic
@subvertallchris


cheerfulstoic:
:+1: thanks

PR #1437, opened 2017-10-08
ensure branch method propagets context like with_association_tree

klobuczek:
Fixes # 
makes sure `branch` method does not erase context created by `with_associations`
This pull introduces/changes:
 *  the `branch` method is implemented using the chain `query.proxy_as`. This construct does not carry over the context of the original QueryProxy. Should it? If so then this solution is probably not the best one as it only tackles the `branch` method.
 * 

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/lib/neo4j/active_node/query/query_proxy.rb
+++ b/lib/neo4j/active_node/query/query_proxy.rb
@@ -187,7 +187,9 @@ module Neo4j
           # `as(identity)` is here to make sure we get the right variable
           # There might be a deeper problem of the variable changing when we
           # traverse an association
-          as(identity).instance_eval(&block).query.proxy_as(self.model, identity)
+          as(identity).instance_eval(&block).query.proxy_as(self.model, identity).tap do |new_query_proxy|
+            propagate_context(new_query_proxy)
+          end
         end
 
         def [](index)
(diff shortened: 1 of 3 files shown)


cheerfulstoic:
This is perfect, thank you!  I like the `propagate_context` method because there certainly might be other things that need to be propagated.

But you're point is well taken about having a more general solution to `propagating` `QueryProxy` information more generally when going in `Query` and back out to `QueryProxy`.  The problem, as I see it, is that `Query` objects in `neo4j-core` generate new `Query` objects for every chained method call.  So the state (at least a reference to the original `QueryProxy`) needs to be propagated through the `Query` objects so that it can be used when doing the `proxy_as`.  Probably it wouldn't be good for the `Query` object to have any idea what a `QueryProxy` is, but it would probably make sense when creating a `Query` object to pass in a generic hash of data which `Query` will pass on blindly and which can be retrieved at the end.

Thanks again, I'll put this fix out now!

cheerfulstoic:
Released as 8.3.1 and 9.0.2

klobuczek:
@cheerfulstoic I think you forgot to merge master to both branches because the PR is not included in either release.

cheerfulstoic:
Oy, thank you.  I don't merge `master` to the branches because there is stuff in master that (at least for 8.3.x) shouldn't be backported.  I must have forgotten to cherry-pick, though...

cheerfulstoic:
That should do it, thanks again

PR #1438, opened 2017-10-13
ensure branch and with_association methods propagate distinct

klobuczek:
Fixes #
As suspected there are more cases where context has to be propagated. Here distinct was suppressed by both branch and with_associations.
This pull introduces/changes:
 * 
 * 

Pings:
@cheerfulstoic
@subvertallchris

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -79,3 +79,8 @@ Style/Lambda:
 # Reason: I'm proud to be part of the double negative Ruby tradition
 Style/DoubleNegation:
   Enabled: false
+
+# Reason: It's OK if the spec files get long as long as they're well factored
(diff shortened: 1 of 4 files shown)


klobuczek:
@cheerfulstoic what do you suggest for the rubocop: too long block in spec?

cheerfulstoic:
Thanks for this!

Actually, in another project we have this in the `.rubocop.yml`:

```yaml
Metrics/BlockLength:
  Exclude:
    - 'spec/**/*'
```

I think it's OK if the spec files get long as long as they're well factored.

cheerfulstoic:
Released in 8.3.4 and 9.0.5

PR #1442, opened 2017-11-15
make RecordNotFound consistent in providing the model

leviwilson:
Fixes inconsistency between `.find_by!` and `.find` when a `RecordNotFound` is thrown so that they both provide the `model`.

This pull introduces/changes:
 * consistency with `RecordNotFound` having a `#model`

Pings:
@cheerfulstoic
@subvertallchris


cheerfulstoic:
Thanks!  Released as 9.0.6

PR #1454, opened 2017-12-16
fixes ordering when with_association used

klobuczek:
Fixes #
when `with_associations` is used in any query chain any order clause (before or after with association) generates cypher which makes ordering ineffective.
This pull introduces/changes:
 * makes sure that ordering happens as the last step in the query after `with_association` has been processed
 *

### DIFF
--- a/lib/neo4j/active_node/query/query_proxy.rb
+++ b/lib/neo4j/active_node/query/query_proxy.rb
@@ -95,13 +95,16 @@ module Neo4j
         #
         #   student.lessons.query_as(:l).with('your cypher here...')
         def query_as(var, with_labels = true)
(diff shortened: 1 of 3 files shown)


cheerfulstoic:
I can't say that I like the side-effect of `@order_chain` being set when you call `chain`, though I think it's probably a symptom of a larger problem and is probably a bit out of the scope of this PR.  Probably the `Query` class shouldn't put `ORDER` into the list of links but hold onto it until a `with`, `return`, or `pluck` and automatically add it at that point.

Anyway, thanks for this!

cheerfulstoic:
Released as 9.1.2

Issue #1453, opened 2017-12-15
feature: an ActiveLabel / ActiveModule module for sharing ActiveNode logic


thefliik:
I think issue #1424 has started to bloom into multiple issues. I'm opening this up to continue the discussion of creating some sort of includeable ActiveLabel module functionality.

#### Preface
@cheerfulstoic began this conversation in #1424 but I didn't really "get it" (see the use case) until, unrelated, I just embarked on a side quest to benchmark the effect of label ordering on query time. As part of this, I downloaded the [Neo4j example movie database](https://neo4j.com/developer/example-data/) to test against. Lo-and-behold, as I tried to use ActiveNode to model the data, I begun to see problems....

I initially thought to model the database like so:

```
class Person 
  include Neo4j::ActiveNode

  property :name

  class User < Person
    self.mapped_label_name = "User"

    property :login    
    property :password    
    property :roles    
  end

  class Actor < Person
    include InShowbusiness

    self.mapped_label_name = "Actor"
  end

  class Director < Person
    include InShowbusiness  

    self.mapped_label_name = "Director"
  end
end

module InShowbusiness
  extend ActiveSupport::Concern

  included do
    property :biography    
    property :lastModified    
    property :version    
  end
end
```

If you check out the example movie database, you'll notice problems with the above though. This is because a node can have `:Person:Actor:User` labels associated with it (i.e. can be both an Actor and a User at the same time). So inheritance doesn't work, because then a node would be wrapped as _either_ an Actor or a User, but would not be both (when it should be both). 

Adding multiple labels to `Person` (such as using modules) doesn't work either though, as then `Person` will _always_ have those labels (which is inappropriate). So `:User`, `:Actor`, and `:Director` labels are truly optional to the person node, except their presence alter's the properties of the person node. In the case of `:User`, it adds properties.

#### The exciting part...

So after reading the ideas in #1424 and trying to solve the movie database problem, here's my take which, the more I think about, the more I like. (it's also possible that this is exactly what @cheerfulstoic and @leehericks where getting at over in #1424)

```
class Person
  include Neo4j::ActiveNode
  include User
  include Actor
  include Director

  property :name

  if_is(User) do
    def admin_powers(id)
      Person.find(id).destroy
    end
  end

  if_is(Actor, Director) do
    def favorite_tv_show
      puts "3rd rock from the sun"
    end
  end
end

module User
  include Neo4j::ActiveModule

  property :login    
  property :password    
  property :roles    

  def stats
[...]

PR #1457, opened 2017-12-19
``ActiveLabel`` documentation proposal

cheerfulstoic:
Just docs so far as a way to think through the notion of ``ActiveLabel``.  See also #1453

### DIFF
--- a/docs/ActiveLabel.rst
+++ b/docs/ActiveLabel.rst
@@ -1,0 +1,135 @@
+ActiveLabel
+===========
+
+``ActiveNode`` provides an ability to define inheritance of models which also gives subclasess the labels of their parent models.  In Ruby, however, inheritence of classes is not sufficient.  Sometimes is makes more sense to be able to build a module which defines behavior (or "concerns") which could be applied to any model.  This is what ``ActiveLabel`` provides.
+
+``ActiveLabel`` modules can be defined in two ways:
+
+ * Default: Where the module's behavior is always defined on the ``ActiveNode`` model and the model's nodes always have a corresponding label in Neo4j
+ * Optional: Where the module's behavior is defined on the class only when the model's nodes have a corresponding label in Neo4j
+
+.. code-block:: ruby
+
+  class Person
+    include Neo4j::ActiveNode
+
+    property :name, type: String
+
+    label :HasAddress
+    label :Destroyable
+  end
+
+  class Organization
+    include Neo4j::ActiveNode
+
+    property :title, type: String
+
+    label :HasAddress
+    label :Destroyable
+  end
+
+
+.. code-block:: ruby
+
+  class Address
+    property :line1, type: String
+    property :line2, type: String
+    property :country, type: String
+    property :postal_code, type: String
+  end
+
+  module HasAddress
+    include Neo4j::ActiveLabel
+
+    included do
+      has_one :out, :address, type: :HAS_ADDRESS
+    end
+
+    module InstanceMethods
+      def distance_from(has_address_object)
+        address.distance_from(has_address_object.address)
+      end
+    end
+  end
+
+
+.. code-block:: ruby
+
+  module Destroyable
+    include Neo4j::ActiveLabel
+
+    follows_label :Destroyed
+
+    included do
+      property :destroyed_at, type: DateTime
+    end
+
+    module InstanceMethods
+      def destroy
+        destroyed_at = Time.now
+
+        super
+      end
+    end
+
+    module ClassMethods
+      def destroyed_recently
+        all.where("#{identity}.destroyed_at > ?", 1.week.ago)
+      end
+    end
+  end
+
+
+Creating
+--------
+
+If an ``ActiveLabel`` does not declare ``follows_label``, creating a node will attach the corresponding label.  Otherwise you must trigger the attachment of the label:
+
+.. code-block:: ruby
+
+  # Node gets both `Person` and `HasAddress` labels
+  person = Person.create
+
+  # `Destroyed' label is added.  `mark_destroyed` method is automatically defined via `follows_label` definition
+  person.label_as_destroyed
+
+  # `Destroyed' label is removed
+  person.label_as_not_destroyed
+
+Querying
+--------
+
+``ActiveLabel`` allows your Ruby module to act like a model class.  However, since you can add a label to any module, you can query for nodes across modules:
+
+.. code-block:: ruby
+
+  Destroyable.all
+
+  HasAddress.as(:obj).address.where(postal_code: '12345').pluck('DISTINCT obj')
+
+By default this returns all nodes for all models where the ``ActiveLabel`` module is defined.  If ``follows_label`` is declared, this returns just those nodes which have the label.
+
+By defining the ``follows_label``, some methods are automatically provided to allow you to filter and interrogate:
+
+.. code-block:: ruby
+
+  Person.labeled_as_destroyed
+
+  Person.first.labeled_as_destroyed?
+
+Associations
+~~~~~~~~~~~~
+
+You can even create associations to traverse to labels:
+
+.. code-block:: ruby
+
+  class Organization
+    include Neo4j::ActiveNode
+
+    has_many :out, :addressables, type: :HAS_ADDRESSABLE_OBJECT, label_class: :HasAddress
+
+    # `model_class` acts as a filter to the `label_class` argument.  Both `model_class` and `label_class` can be arrays
+    has_many :out, :addressable_people, type: :HAS_ADDRESSABLE_OBJECT, label_class: :HasAddress, model_class: :Person
+  end
+
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -19,6 +19,7 @@ Contents:
    RakeTasks
 
    ActiveNode
+   ActiveLabel
    ActiveRel
 
    Properties


leehericks:
I sat down this evening to contemplate this documentation based proposal about labels and what it might allow to be accomplished, because I think good support for labels overall is important.

Overall, there are many ways one can mark and match nodes. Labels provide an automatic index on a set of nodes. Properties can be set like flags and indexes can be created for a set of a label and property together. But matching multiple labels together has known performance issues. Also the more labels and indexes declared, the more the database size grows. And the biggest concern I realized was actually that the more nodes you put the same label on, the more performance problems I think will occur. Imagine if all of your models can have the :Destroyed label through the Destroyable example. Just searching for :Person:Destroyed will actually invoke an index that includes ALL nodes in the graph which were destroyed, and then have some mathematical comparison agains all nodes in the :Person label index. I don't think this is going to encourage good modeling, and if the goal of this is shared, reusable coding features, perhaps something should be implemented that doesn't rely on labels in the graph.

Focusing specifically on the proposed documentation:

I think this was a miss? (mark_destroyed and label_as_destroyed)
```Ruby
# `Destroyed' label is added.  `mark_destroyed` method is automatically defined via `follows_label` definition
person.label_as_destroyed
```
Expanding on that, the following method doesn't actually apply a label :NotDestroyed, but the action wording implies that:
```Ruby
person.label_as_not_destroyed
```
which is why I proposed the boolean format which I think provides the correct context:
```Ruby
label :Destroyed

# Check existence
person.labelled_destroyed?

# Add or remove named label
person.labelled_destroyed=(Boolean)

# Querying
Person.labelled_destroyed
```
This has the added bonus of being one naming style for each named label which has the same grammatical structure, so there is very little to learn or remember. It's consistent and it can be used on any ActiveNode with or without additional modules.

Regarding the ActiveLabel naming, if you don't by default infer the label name from the module, then I think it leads to confusion, because you made a module called `Destroyable` following the label `:Destroyed` and called `Destroyable.all` <- What does this return? All possibly destroyable nodes? Or all nodes with the :Destroy label?

I also feel confused how it can be written `label :Destroyable` on the ActiveNode, and what if the Destroyable module doesn't exist or has some namespace clash? In fact it is written `label :Destroyable` but a `:Destroyed` label will be applied to the node. This doesn't end up being very self-documenting code. Which is why if an association uses the term `label_class`, then the ActiveNode declaration should be `label :Destroyed, label_class: :Destroyable`, which would also need a `mapped_label_name`.

leehericks:
I just realized, if you implement the core label support in ActiveNode as I outlined in the github issue, then anyone could just create a concern and add the label. That then makes less reason to have ActiveLabel separately because it’s just relying on concerns.

```Ruby
Module Destroyable
  include ActiveSupport::Concern

  included do
    label :Destroyed
  end
end
```

cheerfulstoic:
Regarding querying for multiple labels, you are correct in some cases.  It could certainly be the case that when doing a `Person.all` we would only need to use the `Person` label still because adding the labels to the query wouldn't really filter it, unless you were doing a scope like `Person.all.labeled_as_destroyed`, in which case you would need to query on both.  Since that's a case where the user is specifically asking for it I think it's OK.  Similarly if you did `HasAddress.all` it would only be searching on one label.  So I think that this is the ideal way of using labels because is the general cases you're using specific labels until you have to use multiple labels.

Of course when you create a new object with a "default" `ActiveLabel` you'd be adding multiple labels, but that's so that you could search for either the model's label or the module's label.

I'll add my other comments inline on the code

docs/ActiveLabel.rst
```diff
@@ -0,0 +1,135 @@
...
+.. code-block:: ruby
+
+  # Node gets both `Person` and `HasAddress` labels
+  person = Person.create
+
+  # `Destroyed' label is added.  `mark_destroyed` method is automatically defined via `follows_label` definition
+  person.label_as_destroyed
+
+  # `Destroyed' label is removed
+  person.label_as_not_destroyed
```
cheerfulstoic: @leehericks Responding to your comment:

I can see what you're saying about the magic prefixes of `label_as_` and `label_as_not_`, though I feel like this is a pretty common Ruby practice.  To me, the act of changing the state of the node to add / remove a label is semantically an action and not setting a property.  Setting a boolean value seems confusing to me because that boolean value isn't actually going anywhere

I definitely see the appeal of the consistent naming (though again, the Ruby convention that I've seen is to generally go with different phrasings of English even if they aren't exactly the same).  Are you saying that we might have `method_missing` magic like `labelled_foo_bar = true` / `label_as_foo_bar` which would add a `FooBar` label even if there wasn't an `ActiveLabel` module?

docs/ActiveLabel.rst
```diff
@@ -0,0 +1,135 @@
...
+    end
+  end
+
+
+.. code-block:: ruby
+
+  module Destroyable
+    include Neo4j::ActiveLabel
+
+    follows_label :Destroyed
```
cheerfulstoic: @leehericks Responding to your comment about following the module name

I'm definitely a bit on the fence about having a `Destroyable` module which defines a `Destroyed` label, but it was a choice that I made on purpose because I'm trying to keep a Ruby style.  In Ruby you often have module names which are adjectives like `Destroyable` /`Touchable` / `Whateverable`, but that doesn't make sense as a label name.

The question of what `Destroyable.all` returns definitely makes me take a step back.  Maybe it's the case that all of these nodes get a `Destroyable` label but then you can define something like `boolean_label :Destroyed` instead of `follows_label :Destroyed` would let you optionally add the `Destroyed` label to nodes and would also allow you to query for `Destroyed` nodes.

As I'm thinking about this more and more, this might actually be best implemented as two features.  `ActiveLabel` might just be something which always adds the label to the module that it's defined on and we have a method like `boolean_label` which you can apply to `ActiveNode` modules (perhaps even via the `included` block of a module) which allows you to define labels which can be added, removed, and queried/filtered upon.

That could actually simplify defining `ActiveLabel`s as well because rather than doing `label :Destroyable` in the model we could simply do `include Destroyable` and use Ruby's normal module logic to bring in the behavior.  I was using the `label` method so that methods could be defined optionally.

So maybe then it does make sense to think in terms similar to what @thefliik proposed to have something like:

```ruby
optional_label :Destroyed do
  property :destroyed_at, type: DateTime

  def destroy
    destroyed_at = Time.now

    super
  end
end
```

That could be defined on the model or inside of a `include`d module

cheerfulstoic:
Hey @leehericks 

I just saw your latest comment here.  I think it's similar to what I was talking about here:

https://github.com/neo4jrb/neo4j/pull/1457/files#r157831679

docs/ActiveLabel.rst
leehericks: @cheerfulstoic The boolean value needs to go somewhere. Adding and removing labels should fit within the persistence workflow just like properties and be able to check them or enforce some logic in callbacks. If something fails, rolling back the save or update should also make sure the labels are not set. If labels are applied as individual actions, how do you rollback or save errors to the model? This is the same problem I have with associations autosaving right now.

leehericks:
@cheerfulstoic 
What I'm saying is that I've come to realize/assume that developers already have the tools to do the job without this added API.

ActiveNode should give the appropriate means to add and remove labels and query for them. No module should be necessary to add labels to nodes and have the synthesized abilities to query based on them.

What you wanted modules for was shared code, and this example shows that developers can already use ActiveSupport::Concern for that.

```Ruby
module Destroyable
  include ActiveSupport::Concern

  included do
    label :Destroyed
  end

  def destroy!
    self.labelled_destroyed = true
    self.save
  end
end

class Case
  include Neo4j::ActiveNode
  include Destroyable

  property :name, type: String
end

case = Case.create(name: "New Case")
case.labelled_destroyed = true

if case.save
  puts "Case closed!"
end
# OR
case.destroy!

# Query
Case.labelled_destroyed.all
Case.where(has_labels: [:Destroyed]) # Because this literally translates to WHERE n:Destroyed and it's readable
```

Please go back to #1453 and consider my documentation-style example more. That's the core label functionality that ActiveNode needs.

thefliik:
Oh geez, there's so much I want to say all at the same time.

1. As I said in #1453, I think an `ActiveLabel` module should be an all or nothing affair: everything in the module is added if the label is present, or nothing in the module is added. If you want something to _always_ be included in a class, make a separate concern for that. Yes, this potentially makes things a little more verbose, but it will really simplify the developer's understanding of what an `ActiveLabel` module is, and it will simplify the DSL as well. In my scenario, the `ActiveLabel` DSL can be the same as that of `ActiveSupport::Concern`.

This simplifies the `Destroyable` module to
```
module Destroyable
  include Neo4j::ActiveLabel

  follows_label :Destroyed

  def destroy
    destroyed_at = Time.now
    super
  end

  included do
    property :destroyed_at, type: DateTime
  end

  module ClassMethods
    def destroyed_recently
      all.where("#{identity}.destroyed_at > ?", 1.week.ago)
    end
  end
end
```

2. You're going to have to explain your attraction to the `label :Destroyed` DSL. As a developer, what I'm trying to do is conditionally add a module to a class. The condition is "if this label is present, add what's in the module to the object instance". I am NOT adding a label to the class. `Neo4j-core` already provides an api for adding labels to objects, and I'd argue any discussion about Neo4jrb's class label API should be separate from this (for example, #1414).

Since what I'm trying to do is add a _module_ to a class, not a _label_, why not simply use `include TheModule` rather than this custom `label :TheModule` abstraction? Alternatively, and perhaps better, use something like `if_label_present_include ::TheModule`, though it's a bit wordy.

3. If you accept that `ActiveLabel` is just a way of conditionally specifying functionality to add from a module, then `ActiveLabel` shouldn't know, or care, what labels are "optional" to instances of a class vs "always" added to instances of a class. I suppose an argument could be made that `label :HasAddress` could be used as the DSL for adding labels to a class, but that's a separate conversation (in my mind) from `ActiveLabel` (I'd say put that idea over in #1414).

There's more that I could say, such as a discussion of what singleton methods including an `ActiveLabel` module might add to a class, but I'm going to wait to see what you think of the above points, as they really define the conversation (in my mind).

docs/ActiveLabel.rst
```diff
@@ -0,0 +1,135 @@
+ActiveLabel
+===========
+
+``ActiveNode`` provides an ability to define inheritance of models which also gives subclasess the labels of their parent models.  In Ruby, however, inheritence of classes is not sufficient.  Sometimes is makes more sense to be able to build a module which defines behavior (or "concerns") which could be applied to any model.  This is what ``ActiveLabel`` provides.
```
thefliik: I'd replace with:
The Neo4jrb gem provides several ways to keep your code DRY. Subclasses can inherit from an ``ActiveNode`` model, which gives subclasses the labels of their parent models in addition to methods from the parent models. However, inheritance alone is not always sufficient.  Sometimes what you want to do is conditionally add a module to an ``ActiveNode`` model if a specific label is present on an instance of that model. For an example of when this is needed, look at the Neo4j's example movie database (https://neo4j.com/developer/movie-database/). ``ActiveLabel`` allows you to accomplish this.

docs/ActiveLabel.rst
thefliik: Oops, I should have read this first.
> That could actually simplify defining ActiveLabels as well because rather than doing label :Destroyable in the model we could simply do include Destroyable and use Ruby's normal module logic to bring in the behavior.

Looks like we've arrived at similar conclusions

docs/ActiveLabel.rst
leehericks: I still don't understand the conditional part.

Suppose you have a person model who may be an Actor or a Director or both.

1. You still have to be careful not to have namespace clash.
Regardless of having the label or not, you still need associations like directed_movies, acted_movies for each relationship type. If each module called the association `movies`...

2. Is there any benefit to hiding these chunks of code when the label is not set??
Class methods for associations are called when loading the models, right? and then instances are made. Why bother conditionally adding in the the methods? Won't you have to pepper your code with checks if the methods exist and if the label is set anyway?

thefliik: Regarding your first point, you'd only need a `directed_movies` association if someone was a director (i.e. has the `:Director` label). Put another way, if `person.director? == false` then `person.directed_movies` would return an unknown method error.

Regarding your second point, if you don't see a benefit to conditionally adding code based on the presence of a label, what have you been arguing for this whole time? Honestly curious. It seems clear that we've been operating with different goals.

thefliik: Perhaps I need to go back and re-read some of your comments, given your second point above...I've missed something.

leehericks: You're absolutely correct. I never really caught the conditional argument. My bad.

But my concern is, as a developer one has to check `person.director?` before called these methods or one has to handle the unknown method error. What I'm trying to understand then is why conditionally include the code if making the checks anyway?

thefliik: Oh ya, you absolutely need to check for `person.director?` Or in my case, I'd need to check for `person.current?` or `person.destroyed?` and handle these two cases differently. I always need to do this though. A current person node is conceptually different from a destroyed person node and needs to be handled different. Its appropriate that different methods exist on each. 

It's conceptually similar to checking what type of object a variable contains.

As far as why I'd want to conditionally include the code, well mainly because a `current` node might have a method of the same name as a `destroyed` node, but the two methods do completely different things. I'd want to check to see what type of node I was dealing with, so I knew what the method did. Obviously you could achieve similar results with some type of method overloading, but there are always multiple ways to accomplish something in ~~ruby~~ programming.

thefliik: Also worth noting: if your query only returns `:Person:Actor` nodes, then you _know_ that the resulting objects have the `:Actor` methods.

leehericks: Ok, but in the case of :Actor and :Director, a :Person could be both, and now if each one has a method with the same name, you have a clash.

thefliik:
So I made a PR against this PR in #1458

docs/ActiveLabel.rst
thefliik: True, but I think that's a limitation of Ruby modules, rather than a flaw with this implementation strategy. If two modules define a method with the same name, and you include both of them in a class, the last one will overwrite the others.

thefliik: Put another way, if a method might apply to either an :Actor or :Director, you need to be more thoughtful about the naming than if it just applies to one or the other.

klobuczek:
I think this is related. I would like to be able to define:
```has_one :out, :part_of, type: :part_of, model_class: :Specializable```
where `Specializable` is a  module. 
Then
`object.part_of` ideally returns an object of an anonymous class including `Specializable` or with more hints object of an existing class.
The currently required definition:
```has_one :out, :part_of, type: :part_of, model_class: [:Class1, Class2, ...Classn]```
breaks couple of OO paradigms where a class needs to know all classes which include a module `Specializable`.

PR #1458, opened 2017-12-28
Active label API Proposals

thefliik:
After working on the neo4j-core `.union` patch, I wanted to keep up the momentum. This PR contains proposed `ActiveLabel` API updates. The main change from what you had (I think) is that I've separated ``ActiveLabel`` modules, which simply respond to labels, from how labels are added to nodes.

I've also made a few other changes, which were the result of thinking through the implementation process, and realizing that I don't think some of my previous ideas were readily implementable. (for example, if `Actor` is an `ActiveLabel` module, in order to allow `Actor.acted_in` queries, I don't think the `has_many :out, :acted_in` association can be called inside a `included do` block, but rather should be called directly on the `Actor` module. Ideally, I'd have liked it to be called inside an `included do` block to mimic `ActiveSupport::Concern`)

### DIFF
--- a/docs/ActiveLabel.rst
+++ b/docs/ActiveLabel.rst
@@ -3,7 +3,7 @@ ActiveLabel
 
 As you build out your application's models, you likely will want to share code between them.
 Neo4jrb's ``ActiveNode`` module supports class inheritance, allowing you to create "submodels" which 
-inherit the methods and labels of their ``ActiveNode parents`` while also adding their own submodel specific
+inherit the methods and labels of their ``ActiveNode`` parents while also adding their own submodel specific
 label & methods. This code sharing strategy should be familiar to anyone coming from the ActiveRecord world.
 
 Sometime's however, inheritance is not always appropriate. Sometimes what you want to do is conditionally add 
@@ -12,16 +12,18 @@ For an example of when this is needed, look at the Neo4j's example movie databas
 In this example, a Person node is sometimes an Actor, sometimes a Director, sometimes a User, and sometimes a 
 combination of Actor, Director, and/or User.
 
-Multiple inheritance such as this is not possible using ``ActiveNode`` inheritence (or with ActiveRecord). This
+Multiple inheritance such as this is not possible with ``ActiveNode`` (or with ActiveRecord). This
 is where ``ActiveLabel`` comes to the rescue! ``ActiveLabel`` allows you to create a Ruby module which is only
 applied to an ``ActiveNode`` model when a specific label is attached to an instance of the model. Using our
 example movie database from above, you could create an ``ActiveLabel`` module which only adds Actor methods and
-properties to an instance of Person if a Person node also has an Actor label.
+properties to an instance of Person if a Person node also has an Actor label. Or only adds InShowbusiness methods
+and properties to an instance of Person if a Person node has `either` Actor or Director labels.
 
 ``ActiveLabel`` can fully replace ``ActiveNode`` inheritence, but it involves a different way of thinking
 then what many ActiveRecord developers might be used to. If you're just starting out with Neo4jrb, you might find
-it easiest to stick with ``ActiveNode`` and inheritence. As you get more comfortable with Neo4j's flexibility and
-easy polymorphism, you'll find that ``ActiveLabel`` can be incredibly powerful and flexible.
+it easiest to stick with the "ActiveRecord" like workflow provided by ``ActiveNode`` and inheritence.
+But as you get more comfortable with Neo4j's flexibility and 
+polymorphism, you'll might find that ``ActiveLabel`` is the better option for many tasks.
 
 .. code-block:: ruby
 
@@ -106,13 +108,13 @@ include an ``ActiveLabel`` module in an ``ActiveNode`` class if you want the cla
       # When a node is retrieved from the database, it is mapped to an ``ActiveNode`` class and a new
       # instance of that class is created. We'll call this created object obj A.
       # If obj A's class includes this ``ActiveLabel``, and, additionally, obj A has the label associated
-      # with this ``ActiveLabel``, then this included block will be evaluaded
+      # with this ``ActiveLabel``, then this included block will be evaluated
       # within the context of obj A.
     end
 
     module InstanceMethods
-      # After obj A have been found and initialized, before the included block is evaluated, obj A will
-      # be extended with the InstanceMethods (e.g. obj.extend(InstanceMethods))
+      # After obj A has been found and initialized, before the included block is evaluated, obj A will
+      # be extended with these InstanceMethods (e.g. obj.extend(InstanceMethods))
 
       def act
         puts "I acted!"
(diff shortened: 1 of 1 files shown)


docs/ActiveLabel.rst
```diff
@@ -1,135 +1,320 @@
...
     include Neo4j::ActiveNode
 
-    has_many :out, :addressables, type: :HAS_ADDRESSABLE_OBJECT, label_class: :HasAddress
+    has_many :in, :actors, type: :ACTS_IN, label_module: :Actor
 
-    # `model_class` acts as a filter to the `label_class` argument.  Both `model_class` and `label_class` can be arrays
-    has_many :out, :addressable_people, type: :HAS_ADDRESSABLE_OBJECT, label_class: :HasAddress, model_class: :Person
+    # `label_module` acts as a filter to the `model_class` argument.
+    # Both `model_class` and `label_module` can be arrays
+    has_many :in, :human_actors, type: :ACTS_IN, label_module: :Actor, model_class: :Person
```
thefliik: Where you had *"model_class" acts as a filter...* I changed it to *"label_module" acts as a filter* because the class is what is instantiated and so seems like the more "primary" descriptor (in my mind). Put another way, `ActiveLabel` is marking up `ActiveNode`. Also I changed from *"label_class"* to *"label_module"*, because the labels are modules rather than classes.

docs/ActiveLabel.rst
```diff
@@ -1,135 +1,320 @@
...
+    self.associated_labels_matcher = :any
+
+    # OR
+
+    self.associated_labels = [:Actor, :Director]
+    self.associated_labels_matcher = :all
+    
+  end
+
+By default, ``self.associated_labels_matcher == :any``
```
thefliik: Allowing ActiveLabels to have multiple associated labels would also be the primary way of sharing code between multiple `ActiveLabel` modules, rather than using something like `ActiveSupport::Concern`. I realized that sharing code using a concern wouldn't work, because there isn't a good (any?) way of specifying that the concern's code should only apply to certain instances of a class.

docs/ActiveLabel.rst
```diff
@@ -1,135 +1,320 @@
...
+
+    # only run if a Person node also has the Actor or Director labels
+    included_if_any :Actor, :Director do
+      property :medium_ego
+    end
+
+    # only run if a Person node also has the Actor AND Director labels
+    included_if_all :Actor, :Director do
+      property :large_ego
+    end
```
thefliik: I'm thinking that maybe this should be refactored as just `included_if() do`. If an argument is an array, then each label in the array must be present. So the above examples could be refactored as:
```
included_if :Actor, :Director do
  property :medium_ego
end

included_if [:Actor, :Director] do
  property :large_ego
end
```
This format would also be more powerful and allow
```
included_if [:Actor, Director], :Animal do
  property :name
end
```

thefliik: For that matter, the `self.associated_labels =` option could similarly be changed to accept a 2 dimensional array and eliminate `self.associated_labels_matcher =`

docs/ActiveLabel.rst
```diff
@@ -1,135 +1,320 @@
...
   end
 
 
+Creating
+--------
+
+``ActiveLabel`` modules are defined by creating a standard ruby module with ``include Neo4j::ActiveLabel``.
+By convention, the ``ActiveLabel`` module will be associated with a label equal to the module name. For example,
+in the example above, the ``Actor`` ``ActiveLabel`` module is associated with the ``:Actor`` label. You can 
+customize the label(s) which an ``ActiveLabel`` module is associated with using ``self.associated_labels =``. You must also
```
thefliik: This may just come down to personal preference. I originally wrote 

>in the example above, the ``Actor`` ``ActiveLabel`` module follows the ``:Actor`` label. You can customize the label(s) which an ``ActiveLabel`` module follows using ``self.associated_labels =``.

But I found the `follows` metaphor to be less intuitive for me than the `associated` metaphor, which resulted in the change to `self.associated_labels =`.

I don't feel strongly about this though.

docs/ActiveLabel.rst
```diff
@@ -1,135 +1,320 @@
...
+    # to every instance of the Person class. The :Actor label is technically
+    # optional, even though it is always added, because a node will still be mapped
+    # to the Person class even if you manually remove the :Actor label from it.
+    label :Actor, optional: true
+
+    # If you call the ``label`` method without the ``optional: true`` argument,
+    # then nodes will only be mapped to the Person class if the label is
+    # also present on the node. (i.e. removing the :User label from a node will
+    # mean that that node is no longer considered a Person)
+    label :User
```
thefliik: Note, if the `Person` class has `label :User` (or `label :User, optional: true`), then there would be no difference between `Person.create` and `Person.user.create`. This should probably be explicitly mentioned.

docs/ActiveLabel.rst
```diff
@@ -1,135 +1,320 @@
...
+.. code-block:: ruby
+
+  # Initializes a Person with additional Actor label
+  Person.actor.new
+
+  # Creates a Person with additional Actor label
+  Person.actor.create
+
+  # Creates a Person with additional Actor AND Director labels
+  Person.actor.director.create
```
thefliik: I find this api to be more expressive (and concise) than 
```
person = Person.create
person.add_label(:Actor)
```
The above is also two DB queries, where (`Person.actor.create`) could be one DB query.

docs/ActiveLabel.rst
thefliik: Another issue with any potential `included_if() do` block, I'm not sure how the following should be handled
```
class Person
  include Neo4j::ActiveNode
  scope :actors, ->{ where("#{identity}:Actor") }

  included_if :Actor do
    has_many :out, :acts_in, type: :ACTS_IN, model_class: :Movie
  end
end
```
In the above, devs could
```
movie = Movie.create
actor = Person.create
actor.add_label :Actor

actor.acts_in << movie
```
But I could see a dev also wanting to do something like `Person.actors.acts_in`, which they couldn't do because the `has_many :out, :actors` block would never be evaluated on `Person` (only on instances of person).

Maybe this is fine, and would simply be a known limitation of `included_if()` blocks. Devs could still use the blocks to add properties, define methods, or call custom methods. Or maybe there's another option.

I definitely view anything like `included_if()` as a "nice to have" feature. Maybe it would be more trouble than its worth.

Another note: `ActiveLabel` will, I imagine, make devs want the ability to filter an association proxy by label. Something like `Person.all.has_label(:Actor).acts_in`.

docs/ActiveLabel.rst
thefliik: This being said, I'm wondering what happens given the following
```
module Hollywood
  include Neo4j::ActiveLabel

  self.associated_labels = [:Actor, :Director]
  self.associated_labels_matcher = :any
end

class Person
  include Neo4j::ActiveNode
  include Hollywood
end

person = Person.hollywood.create
```
What labels does `person` have? It could be that we, optionally, let the dev define their own label helper methods. Something like
```
module Hollywood
  include Neo4j::ActiveLabel

  self.associated_labels = [:Actor, :Director]
  self.associated_labels_matcher = :any

  class_label_maker :actor, :Actor
  class_label_maker :director, :Director
end

actor = Person.actor.create
director = Person.director.create
```

Below, I added a comment suggesting that `self.associated_labels` accept a two dimensional array instead of `self.associated_labels_matcher`. What would happen if
`self.associated_labels = [[:Actor, :Director], :Animal]`?

I'm thinking that, if `self.associated_labels = []`, then devs must define their own class label makers. Or maybe we could simply always make devs define their own class label makers, and make it an `ActiveNode` method.
```
module InShowbusiness
  include Neo4j::ActiveLabel

  self.associated_labels = [[:Actor, :Director], :Animal]

  class_label_maker :hollywood, [:Actor, :Director]
  class_label_maker :animal, :Animal
end
```

klobuczek: @thefliik again you are bringing up something I had on my mind for a very long time. My idea was to do all this natively with ruby without any changes to the DSL or the api methods.
If each Label maps to an ActiveLabel module to create a node with a set of labels you would construct an object of a class which includes all the ActiveLabels you are interested in. When you read a node from the database and the node's combination of Labels does not correspond to any class you would return an object of an anonymous class which has all the corresponding ActiveLabels included. This feels very natural and intuitive and completely in harmony with the ruby language and how neo4j labels work.

thefliik: @klobuczek Interesting. I think it's important for me to say that this proposal only _adds_ to the existing API, and, even if this were implemented, anyone could continue to use the neo4j gem exactly as they have been without changes.

That being said, your suggestion sounds pretty similar to mine, except that you are getting rid of individual ActiveNode classes and simply using one, anonymous class. Are you suggesting this because you think it would be more flexible?

One downside of your approach, is that (and correct me if I'm wrong), but in your approach a specific label (e.g. `:Person`) could only be association with *one* ActiveLabel module, as oppose to my API in which the `:Person` label could have totally different meanings depending on the ActiveNode class of the object. It also _seems_ like my approach would be easier to integrate into existing applications (simply mixin an ActiveLabel module to your existing ActiveNode class). I could also envision many devs, coming from an ActiveRecord background, ignoring the ActiveLabel functionality in the beginning and building out their app just using ActiveNode (ActiveNode is much more similar, conceptually, to ActiveRecord, after all). At some point in the future when they got more comfortable with Neo4j's polymorphism, they could start mixing ActiveLabel modules into their classes.

I guess what I'm trying to say, is that both strategies are basically the same: neo4jrb gets a node from the database and then wraps that node in a class based on the labels the node has. Then, based on which of those labels are determined to be ActiveLabels, the object gets extended with one or more modules. Why toss `ActiveNode` classes from this process?

klobuczek: @thefliik Please note the ActiveNode is not a class but a module, so this is already different from active record, but does offer a huge advantage. What I am suggesting is to not be required to define classes including ActiveNode with properties and associations (which would be standard way like today), but allow defining just modules (including ActiveNode or ActiveLabel or similar,) which you can combine to classes, but which could live alone if an unknown combination of labels is retrieved from the database. This is the only case where I would return anonymous class. This would work nicely for creation and retrieval. I have not thought through the promotion (adding a label to existing node) and demotion (removing a label), which does not seem to have a native counterpart in ruby.

PR #1459, opened 2017-12-29
Fix: out of date API usage

thefliik:
`Neo4j::Transaction.run` no longer works (at least generally) in the neo4j gem. `Neo4j::ActiveBase.run_transaction` does.

However, I'm not sure if `Neo4j::ActiveBase` is present in neo4j-core. Maybe in neo4j-core `Neo4j::Transaction` is still correct?

Fixes @JakubA3informatics's question that came up in Gitter

### DIFF
--- a/docs/Testing.rst
+++ b/docs/Testing.rst
@@ -82,6 +82,15 @@ If you are using RSpec you can perform tests in a transaction as you would using
 
 .. code-block:: ruby
 
+  # For the `neo4j-core` gem
+  config.around do |example|
+    session.transaction do |tx|
+      example.run
+      tx.mark_failed
+    end
+  end
+  
+  # For the `neo4j` gem
   config.around do |example|
     Neo4j::ActiveBase.run_transaction do |tx|
       example.run
--- a/docs/Testing.rst
+++ b/docs/Testing.rst
@@ -83,7 +83,7 @@ If you are using RSpec you can perform tests in a transaction as you would using
 .. code-block:: ruby
 
   config.around do |example|
-    Neo4j::Transaction.run do |tx|
+    Neo4j::ActiveBase.run_transaction do |tx|
       example.run
       tx.mark_failed
     end


cheerfulstoic:
You're right that `Neo4j::ActiveBase` doesn't exist in `neo4j-core`, though `Neo4j::Transaction.run` is supposed to work generally.  Though honestly, I'm not sure how, come to think of it.  In the old versions of `neo4j-core` there was a concept of a "current" session, but at one point that was elevated so that in `neo4j-core` you needed to manage your own sessions and the `neo4j` gem is now the one with the concept of a current session.

The reason that `Neo4j::Transaction.run` is around is because we were trying to preserve the old transaction API (this was a request of @subvertallchris I think), though the code to support it in the new transactions API in `neo4j-core` is really complicated and I bang my head every time I try to read it.  Maybe we should bite the bullet and remove it because I think it is based around the idea of a current session in `neo4j-core`.

IIRC (and from looking at the code), the way to make a transaction in the `neo4j-core` gem without the `neo4j` gem is to do something like:

```ruby
tx = session.transaction
# Do something with the session
tx.commit

# or
session.transaction do
  # Do something with the session
end

# where `session` can be gotten from:

Neo4j::Core::CypherSession.new(http_or_bolt_or_embedded_adaptor)

# or as `Neo4j::ActiveBase.current_session`
```

Though come to think of it, I don't know that that API far making a transaction on a `neo4j-core` session is documented anywhere, so this could help with that

thefliik:
So I believe the correct usage of `Neo4j::Transaction` is `Neo4j::Transaction.run(current_session, boolean)` where current session is the cypher session and boolean indicates if it should actually be run in a transaction (I'm assuming this is a legacy thing). https://github.com/neo4jrb/neo4j/blob/c270799880d60090a67e5f77b936283eeceecda7/lib/neo4j/active_base.rb#L38-L42

If I was to guess, I'd say it wasn't that you kept `Neo4j::Transaction` in for legacy reasons, rather you've kept the awkward params around for legacy reasons. Though I'm just guessing looking at the code: https://github.com/neo4jrb/neo4j-core/blob/00190f40810b8fbb559ede1592273501d1230349/lib/neo4j/transaction.rb#L127

In neo4jrb though, the current session is managed by `ActiveBase` (I think) so the public API can be / is `Neo4j::ActiveBase.run_transaction`. In Neo4j-Core, I'm assuming you need to manage everything yourself.

However you wish to move forward, I think it would make sense to update the docs (as time allows) to, in general, reference `ActiveBase` transactions, since I'm assuming most people read the docs as users of Neo4jrb. I also feel like I've seen a few people posting about issues stemming from direct usage of `Neo4j::Transaction`

cheerfulstoic:
Regarding "neo4jrb", that's something that I've been trying to clarify for a long time.  For a long time people have referred to the `neo4j` gem as Neo4j.rb, though that doesn't really make sense to me since you can use either the `neo4j` or `neo4j-core` gems to use Neo4j with Ruby, so I've been referring to both gems (and even including `neo4j-rake_tasks`) as the Neo4j.rb project.  Honestly, we probably should have called the `neo4j-core` gem `neo4j` and the `neo4j` gem `active_node` (or `active_neo4j` since there's also `ActiveRel`) or something to make things more clear.

That aside ;)

I think it's true that most people use the `neo4j` gem, but there are definitely people that only use `neo4j-core` on purpose because they want to manage the queries themselves (for this reason a lot of people preferred the `neography` gem for a long time because it didn't have the additional baggage on top, but that gem hasn't been much maintained for a while).  So I definitely try to present both options in the docs for using either just `neo4j-core` or using `neo4j` on top of it

thefliik:
I've always assumed that most people, like myself, reference the gem as Neo4jrb to differentiate it from Neo4j the database (and because saying "neo4j the gem" takes too long). If neo4j-core was named neo4j, I'd refer to it as "neo4jrb". Personally, and I assume this holds true for others, "neo4j" will always, by default, be referring to the database. I think most people understand that `neo4jrb` refers to the github organization. Since there's no `neo4jrb/neo4jrb` project however, its easiest to just simplify `neo4jrb/neo4j` as `neo4jrb`.

re: docs. That makes sense. Though currently the issue this patch is trying to fix isn't described correctly for `neo4j-core` or `neo4jrb/neo4j`. I'm not sure what the correct instructions would be for `neo4j-core`

cheerfulstoic:
Yeah, saying "the `neo4j` gem" is definitely more onerous, but it's what I tend to do because I don't have a better way to make it clear what I'm talking about.  `¯\_(ツ)_/¯`

But regarding this PR I'd like to make it so that the two options for using the `neo4j` gem and only using `neo4j-core` are clear.  I could see doing this a couple of ways.  The simplest would be to just change it to:

```ruby
# For the `neo4j` gem, `session` can be retrieved via `Neo4j::ActiveBase.current_session`
Neo4j::Transaction.run(session) do |tx|
```

But probably the idea thing for people to use (even though it makes the docs a bit more verbose) would be:
```ruby
    # For the `neo4j-core` gem
    session.transaction do |tx|
      example.run
      tx.mark_failed
    end

    # For the `neo4j` gem
    Neo4j::ActiveBase.run_transaction do |tx|
      example.run
      tx.mark_failed
    end
```

thefliik:
Updated

cheerfulstoic:
❤️ Thanks!

PR #1460, opened 2017-12-31
Add: bang setters for associations

jorroll:
This PR adds bang setters for simple (non-ActiveRel) associations which raise an error if the association fails to persist (without an extra DB call). In my testing, creating an association with `ActiveRel` objects using either `.create!` or `.save!` already raises an error if it fails to persist.

```
person = Person.create
owner = Owner.create

Neo4j::ActiveBase.query("
  MATCH (n:`Person` {uuid: '#{person.id}'})
  DELETE n
")

owner.set_person! person
#=> Neo4j::ActiveNode::HasN::Association::PersistanceError: It appears the relation failed to persist in the database
```

All the tests are passing, but I haven't added any new tests yet. Implementing this ended up being pretty confusing and touching a lot of stuff, so I wanted to run it by you first. The implementation, while working, feels pretty ugly. Though it could be, in part, a result of how complicated the association logic is.

has_one associations now have:
* `object.set_#{association}!`

has_many associations now have:
* `object.set_#{associations}!`
* `object.add_#{association}!`

### DIFF
--- a/lib/neo4j/active_node/unpersisted.rb
+++ b/lib/neo4j/active_node/unpersisted.rb
@@ -7,9 +7,14 @@ module Neo4j
         @deferred_create_cache ||= {}
       end
 
+      def deferred_options_cache
+        @deferred_options_cache ||= {}
+      end
+
       def defer_create(association_name, object, options = {})
         clear_deferred_nodes_for_association(association_name) if options[:clear]
 
+        deferred_options_for_association(association_name) << {verify: options[:verify]}
         deferred_nodes_for_association(association_name) << object
       end
 
@@ -17,12 +22,17 @@ module Neo4j
         deferred_create_cache[association_name.to_sym] ||= []
       end
 
+      def deferred_options_for_association(association_name)
+        deferred_options_cache[association_name.to_sym] ||= []
+      end
+
       def pending_deferred_creations?
         !deferred_create_cache.values.all?(&:empty?)
       end
 
       def clear_deferred_nodes_for_association(association_name)
-        deferred_nodes_for_association(association_name.to_sym).clear
+        deferred_options_for_association(association_name).clear
+        deferred_nodes_for_association(association_name).clear
       end
 
       private
@@ -31,13 +41,21 @@ module Neo4j
         deferred_create_cache.dup.each do |association_name, nodes|
           association_proxy = association_proxy(association_name)
 
-          nodes.each do |node|
+          options_array = deferred_options_for_association(association_name)
+
+          nodes.each_with_index do |node, index|
+            options = options_array[index]
+
             if node.respond_to?(:changed?)
               node.save if node.changed? || !node.persisted?
               fail "Unable to defer node persistence, could not save #{node.inspect}" unless node.persisted?
             end
 
-            association_proxy << node
+            if options[:verify]
+              association_proxy.add_relation! node
+            else
+              association_proxy << node
+            end
           end
         end
 
--- a/lib/neo4j/shared/query_factory.rb
+++ b/lib/neo4j/shared/query_factory.rb
@@ -29,6 +29,10 @@ module Neo4j::Shared
       graph_object.persisted? ? match_query : create_query
     end
 
+    def query!
+      graph_object.persisted? ? match_query! : create_query!
+    end
+
     # @param [Neo4j::Core::Query] query An instance of Neo4j::Core::Query upon which methods will be chained.
     def base_query=(query)
       return if query.blank?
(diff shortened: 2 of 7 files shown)


lib/neo4j/active_node/unpersisted.rb
```diff
@@ -7,22 +7,32 @@ def deferred_create_cache
...
 
+      def deferred_options_cache
+        @deferred_options_cache ||= {}
+      end
+
       def defer_create(association_name, object, options = {})
         clear_deferred_nodes_for_association(association_name) if options[:clear]
 
+        deferred_options_for_association(association_name) << {verify: options[:verify]}
         deferred_nodes_for_association(association_name) << object
```
thefliik: I spent a few hours trying to get the `deferred_create_cache` to hold both the deferred nodes and options for each node (currently, options is just a hash which indicates whether the node's persistence should be verified, but in the future it could hold other stuff). My general (preferred) strategy would be `deferred_options_for_association(association_name) << [object, {verify: options[:verify]}]` rather than storing the options in a separate cache. I couldn't get it working though. Somewhere in the code base, the `@deferred_create_cache` is used for constructing associations in a manner I couldn't find. The following would fail:
```
person = Person.new
email = EmailAddress.new
person.email_addresses << email

person.email_addresses # <- this would get messed up by the `[object, {verify: options[:verify]}]` in the deferred_create_cache
```

lib/neo4j/shared/query_factory.rb
```diff
@@ -89,6 +97,12 @@ def create_query
                 .params(:"#{identifier}_create_props" => create_props)
     end
 
+    def create_query!
+      return match_query if graph_object.persisted?
```
thefliik: So this line:
```
return match_query if graph_object.persisted?
```
exists in the, already existing, `create_query` method. `match_query` simply matches on,
```
"(#{graph_object.from_node_identifier})-[#{identifier}]->()"
```
It's unclear to me why you can get away without also specifying the `to_node`. In my testing though, it didn't seem as if `graph_object.persisted?` ever `== true` while setting an association, so I decided to just leave it as is. in `create_query!`

cheerfulstoic:
Are these analogous to `association = node` and `association << node`?  If so, do those methods not raise errors?  Or are these new methods filling out compatibility with methods which `ActiveRecord` adds?

thefliik:
1. Yes, the are analogous to `association = node` and `association << node`
2. No, those methods do not raise errors.

Which I think is fine, except, I would argue, that they should have a bang equivalent which does raise an error. Hence this PR

cheerfulstoic:
I'm fine with the methods, though I think I'm not sure in which case errors are / aren't raised.  If the database gives an error we'll get an exception.  Is this if the associated model doesn't validate?  If so, I think this is a great addition

thefliik:
Currently (sans-PR), calling `association << node` makes a `CREATE` db call to persist the relation and that db call has no `RETURN`.

The bang version of `association << node` ("`add_#{association}!`") returns `true` or `false` depending on whether the newly created pattern `exists()`.
```
exists_string = "(#{graph_object.from_node_identifier})-[#{identifier}]->(#{graph_object.to_node_identifier})"
create_query.return("exists(#{exists_string}) as result")
```
If it returns false, an exception is raised.
```
def _create_relationship_with_factory(verify=false)
  Array(other_node_or_nodes).each do |other_node|
    wrapper = _rel_wrapper(properties)
    base = _match_query(other_node, wrapper)
    factory = Neo4j::Shared::RelQueryFactory.new(wrapper, wrapper.rel_identifier)
    factory.base_query = base

    if verify
      if factory.query!.response.pluck(:result).first
        true
      else
        raise PersistanceError, "It appears the relation failed to persist in the database"
      end
    else
      factory.query.exec
    end
  end
end
```

cheerfulstoic:
Cool, I'm down with this, then :+1:

thefliik:
Awesome! Just coming back to this. Do you have any opinion on my implementation strategy? I've taken an additive approach with this patch. Rather than refactoring the underlying code, I simply added new args to various methods to make things work. This approach makes backwards compatibility easy (and _especially_ backwards compatibility with the specs), but it _feels_ suboptimal. The underlying logic was already pretty complex, and this definitely increases it.

It kinda feels like a refactoring of the code would be appropriate at some point (though maybe you disagree). Should I flesh out this patch as is (and worry about any possible refactorization as some future project), or should any patch like this attempt to "get things right" the first time?

The large code comment I wrote above, mentions one bit of the implementation I don't like.

No rush on replying if you're busy

cheerfulstoic:
I'm all for a refactoring for sure if it doesn't change the API and if feel up to tackling it.  I definitely think there's some weirdness around how the cache works for associations in how it goes to the DB when it doesn't need to.

If it's not backwards compatible I'm still open to it, I would just lay it out.  From what I see so far I don't suspect that there would be a problem.

PR #1465, opened 2018-01-11
Add: notes about ActiveBase transactions

thefliik:
I added some notes about `ActiveBase.run_transaction` to the upgrade guide. I also referenced the recommended `neo4j-core` transaction API which you added to the testing guide (though it's unclear to me if that API is different in version 7.0 of `neo4j-core` vs version 6.0)

### DIFF
--- a/docs/UpgradeGuide.rst
+++ b/docs/UpgradeGuide.rst
@@ -80,7 +80,7 @@ Or via the `Neo4j::Core::Query` class
 
   neo4j_session.query(query_obj)
 
-Making multiple queries with one request is support with the HTTP Adaptor:
+Making multiple queries with one request is supported with the HTTP Adaptor:
 
 .. code-block:: ruby
 
@@ -101,6 +101,15 @@ When doing batched queries, there is also a shortcut for getting a new `Neo4j::C
   end
 
   results[0] # result
+  
+With your session object, you can wrap multiple queries inside a transaction like so:
+
+.. code-block:: ruby
+
+  neo4j_session.transaction do |tx|
+    # do stuff
+    tx.mark_failed
+  end
 
 The ``neo4j`` gem
 ~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


docs/UpgradeGuide.rst
```diff
@@ -119,6 +119,42 @@ If you are using version ``8.0`` of the ``neo4j`` gem, that will be accessible,
...
+  end
+  
+In ``8.0``, the API for interacting with transactions looks like
+
+.. code-block:: ruby
+
+  Neo4j::ActiveBase.run_transaction do |tx|
+    # do stuff
+    tx.mark_failed
+  end
```
cheerfulstoic: The examples are good, but the first example is a pure `neo4j-core` example, so we should present an the new `neo4j-core` example below first and then say something like "If you are using `ActiveRecord` / `ActiveRel` in the `neo4j` gem, you should use..." and then present this example.

cheerfulstoic: Also, since we'd be talking about the `neo4j-core` gem the version change wouldn't be from 7 to 8.  But since this is all under the "The neo4j gem from 7.x to 8.x and the neo4j-core gem from 6.x to 7.x" header we could just skip talking about versions and talk about what's changed

docs/UpgradeGuide.rst
```diff
@@ -119,6 +119,42 @@ If you are using version ``8.0`` of the ``neo4j`` gem, that will be accessible,
...
+
+.. seealso::
+  .. raw:: html
+
+    Check out the ActiveBase source code to learn about some other neat helper methods <a href='https://github.com/neo4jrb/neo4j/blob/master/lib/neo4j/active_base.rb'>ActiveBase has</a>
```
cheerfulstoic: Totally agreed that people should know about `ActiveBase` methods, though it would be nice to have them documented somewhere so that people don't need to look at the source (though this is fine for a start, thanks!).

thefliik:
I hadn't noticed that these examples were under the "neo4j gem" section. I've rearranged to mirror the existing format.

cheerfulstoic:
:+1: Thanks!

cheerfulstoic:
(Thanks for basing against 9.1.x, BTW.  Makes the deployment of the changes easier!)

PR #1475, opened 2018-01-24
replaces session call  with hash value call on railtie

### LINKED ISSUE #1471: neo4j_config.session leading to stack too deep
Found this weird issue after upgrading to rails 5.1.
Whenever the `final_session_config ` method is called on `neo4j/railtie.rb` and it calls neo4j_config.session I get a stack level too deep error. The issue seems to be related with the ActiveSupport::OrderedOptions which allows to call `neo4j_config.session`.

I was able to get around it by changing the code to call `neo4j_config[:session]` instead.
 Would gladly submit a PR with the alteration if need be.

`neo4j` gem version: 8.3.4
`neo4j-core` gem version: 8.1.0

RicardoTrindade:
Fixes #1471

This pull introduces/changes:
 * Replaces `neo4j_config.session` method calls with `neo4j_config[:session]` on `railtie.rb`, due to Object.session method from Rails5 being called and leading to stack too deep error.

### DIFF
--- a/lib/neo4j/railtie.rb
+++ b/lib/neo4j/railtie.rb
@@ -73,20 +73,20 @@ module Neo4j
     def final_session_config!(neo4j_config)
       support_deprecated_session_configs!(neo4j_config)
 
-      neo4j_config.session.empty? ? yaml_config_data : neo4j_config.session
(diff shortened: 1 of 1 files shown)


cheerfulstoic:
Sorry for the delay, I just got around to testing this.  Thanks so much for the contribution!

PR #1482, opened 2018-02-13
ActiveRel find_by_id should throw RecordNotFound when no results

lshimokawa:
Currently when doing `ActiveRel.find(999)`, the method throws `undefined method r for nil`.

This PR handles `ActiveRel.find` similar to `ActiveNode.find` throwing a `Neo4j::RecordNotFound`.

### DIFF
--- a/lib/neo4j/active_rel/query.rb
+++ b/lib/neo4j/active_rel/query.rb
@@ -2,6 +2,8 @@ module Neo4j::ActiveRel
   module Query
     extend ActiveSupport::Concern
 
+    class RecordNotFound < Neo4j::RecordNotFound; end
+
     module ClassMethods
       # Returns the object with the specified neo4j id.
       # @param [String,Integer] id of node to find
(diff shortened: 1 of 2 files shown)


cheerfulstoic:
I'll merge this and release, but I would very much recommend not using `ActiveRel.find`, and it may go away at some point in the future.  Neo4j doesn't allow for any indexing for relationships, and so a query like this (`()-[r]-()`) will scale very poorly as your database grows.  Neo4j is (as I understand it) designed around the idea that you should start with nodes in your query as use relationships for traversal to other nodes.  Thus one of the nodes should be part of your query, or perhaps the relationship should be a node itself

cheerfulstoic:
(and after all of that explination, BTW: thanks for the submission!)

cheerfulstoic:
Released as 9.1.4

klobuczek:
@cheerfulstoic mapping a neo4j db model to jsonapi.org resources is not straightforward. The choice we made was to model relationships with at least 1 property as resource. Consequently we made such resources addressable. I am not sure if it would have been better not to expose the show routes to such resources and if this is a viable option in our domain.

cheerfulstoic:
I'm also on a project where we used JSON API and when we represent relationships they are always either an extension of the URL from the source node or returned in the body of the data for a node.  It is my understanding (and I'm happy to be corrected if this is wrong) that relationships aren't first class citizens in Neo4j with regards to queries and you can be in for a bad time if you treat them that way.  If relationships were more directly queryable in Neo4j / Cypher then I would be happy to support query methods in `ActiveRel`, but as it is I feel the need to steer people away from them as a steward of the Ruby libraries

PR #1484, opened 2018-02-20
Add TypeConverter DSL

jgaskins:
This pull introduces/changes:
 * Added `Neo4j::Shared::TypeConverters.create` DSL method to create a simple converter

I noticed that most of the converters are written as classes with all their methods defined on `self` and they could just be object instances with the `primitive_type` and `convert_type` being stored as instance variables rather than full-on methods. And since most of the type converters I write are simple conversions, I extracted most of my converters to look like this:

```ruby
Neo4j::Shared::TypeConverters.create db: String, ruby: BCrypt::Password do
  db(&:to_s)
  ruby do |value|
    if BCrypt::Password.valid_hash?(value)
      BCrypt::Password.new(value)
    else
      BCrypt::Password.create(value)
    end
  end
end
```

In my own app, this is all within the `Conversion` class, which is why it's still called that in this PR —  I straight-up pasted from my own code and extracted the method to live on the `TypeConverters` module. I wasn't sure if this was something people might want, so I haven't cleaned it up much, but I wanted to offer it up in case anyone found it useful.

### DIFF
--- a/lib/neo4j/shared/type_converters.rb
+++ b/lib/neo4j/shared/type_converters.rb
@@ -324,6 +324,51 @@ module Neo4j::Shared
       end
     end
 
+    # DSL method to create a simple type converter
+    #
+    # @param db [Class] The primitive type to be stored in the database
+    # @param ruby [Class] The class to deserialize into
+    # @example
+    #   require 'bcrypt'
+    #   Neo4j::Shared::TypeConverters.create db: String, ruby: BCrypt::Password do
+    #     db(&:to_s)
+    #     ruby do |value|
+    #       if value.start_with? '$2a$'
+    #         BCrypt::Password.new(value)
+    #       else
+    #         BCrypt::Password.create(value)
+    #       end
+    #     end
+    #   end
+    def self.create(**attrs, &block)
+      Conversion.new(**attrs).tap do |converter|
+        converter.instance_exec(&block)
+        Neo4j::Shared::TypeConverters.register_converter converter
+      end
+    end
+
+    class Conversion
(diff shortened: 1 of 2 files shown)


cheerfulstoic:
This is awesome, thank you!  A couple of things:

 * Could you change the documentation to reflect this DSL?  I think that [this](https://github.com/neo4jrb/neo4j/blob/master/docs/Properties.rst#custom-converters) would be the appropriate section
 * For an example, I didn't think about the fact that `db(&:to_s)` was equivalent to `db {|value| value.to_s }`, leading me to think that the whole block was uneven (I thought that you returned the DB value and defined a `ruby` block).  It might be good in the docs to use a explicit block for whatever example it is to avoid confusion and people can optimize as they like

jgaskins:
> Could you change the documentation to reflect this DSL?

Heck yes I can.

Speaking of docs, this project has had great docs ever since I first looked at it, but they've improved a lot since 8.0 and I really appreciate your continued efforts on that front.

> I didn't think about the fact that `db(&:to_s)` was equivalent to `db {|value| value.to_s }`

I actually meant to change that here for that exact reason and forgot before posting. The shorthand is not very expressive in this situation at all, but easy to write. My coworkers always call me out on stuff like that, too. My bad. :-)

cheerfulstoic:
Thanks on all counts ;)  I try to operate under the assumption that is somebody has a question then the docs fell short, but there are still some holes I'd like to fill eventually ;)

And I totally do stuff like `db(&:to_s)` myself and am fine with it, I just think the other way would be more accessible (though I don't think I need to explain this, given your comments ;) )

PR #1485, opened 2018-03-02
Add: ability to create association with `labels: false` by default

thefliik:
As brought up on Gitter, this PR adds the ability to create an association where `model_class != false` but the cypher generated still does not specify labels for the target node. Currently, the only way to have an association's cypher not specify labels for the target node by default is to use`model_class: false`, which prevents someone from easily query chaining.

***Update: below is fixed***
~~_Note: I've commented out one spec I added for the has_one association which is failing. As far as I can tell, the problem is not with the code, and is only a problem with the spec (the exact same test works fine in the console). I think it has to do with how the class in the spec is being stubbed, but I really just guessing. I don't know much about stubbing. See the code for more info_~~

### Additional

As I discovered while writing this PR, the ability to remove the target node labels on a case-by-case basis already existed as an association getter option (e.g. `person.friends(labels: false)`). What this PR does is add the ability to specify default association getter options when you create the association.

In this PR, the only configurable default option is `:labels`, but this can easily be changed in the future. I also added a section to the docs covering this PR

### DIFF
--- a/spec/e2e/has_many_spec.rb
+++ b/spec/e2e/has_many_spec.rb
@@ -387,12 +387,28 @@ describe 'has_many' do
 
   describe 'association "getter" options' do
     before do
+      Person.has_many :out, :best_friend, model_class: 'Person', type: nil, labels: false
       node.knows << friend1
       friend1.knows << friend2
     end
 
     it 'allows passing only a hash of options when naming node/rel is not needed' do
       expect(node.knows(rel_length: :any).to_a).to match_array([friend1, friend2])
+    end
+
+    it 'honors default options' do
+      expect(node.knows.instance_variable_get('@query_proxy').instance_variable_get('@association_labels')).to eq(nil)
+      expect(node.best_friend.instance_variable_get('@query_proxy').instance_variable_get('@association_labels')).to eq(false)
+    end
+
+    it 'allows overriding of default options' do
+      expect(node.knows(labels: false).instance_variable_get('@query_proxy').instance_variable_get('@association_labels')).to eq(false)
+      expect(node.best_friend(labels: true).instance_variable_get('@query_proxy').instance_variable_get('@association_labels')).to eq(true)
+    end
+
+    it 'provided options are merged into default options' do
+      expect(node.knows({}).instance_variable_get('@query_proxy').instance_variable_get('@association_labels')).to eq(nil)
+      expect(node.best_friend({}).instance_variable_get('@query_proxy').instance_variable_get('@association_labels')).to eq(false)
     end
   end
 
--- a/spec/e2e/has_many_spec.rb
+++ b/spec/e2e/has_many_spec.rb
@@ -397,18 +397,18 @@ describe 'has_many' do
     end
 
     it 'honors default options' do
-      expect(node.knows.instance_variable_get('@query_proxy').instance_variable_get('@association_labels')).to eq(nil)
-      expect(node.best_friend.instance_variable_get('@query_proxy').instance_variable_get('@association_labels')).to eq(false)
+      expect(node.as(:person).knows(:known, :r).to_cypher).to include('MATCH (person)-[r:`KNOWS`]->(known:`Person`)')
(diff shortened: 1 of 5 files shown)


spec/e2e/has_one_spec.rb
```diff
@@ -206,6 +208,41 @@
...
+          .instance_variable_get('@association_labels')
+      ).to eq(false)
+    end
+
+    # This is failing with a "NameError: undefined local variable or method `node'".
+    # As far as I can tell, this exact test works fine in console testing, making me think this is
+    # just a problem with the spec. My guess is it has to do with
+    # how the class is stubbed ?
+
+    # it 'allows overriding of default options' do
```
thefliik: This is the problem spec I was referring to

thefliik:
I've noticed that if `model_class: []` then neo4jrb still adds in the node labels as part of a where clause. i.e.
```
class Person
  has_many :out, :entities, type: :ENTITY, model_class: [:LabelOne, :LabelTwo], labels: false
end

person.entities
# =>
# MATCH (person)-[:ENTITY]->(result)
# WHERE result:`LabelOne` OR result:`LabelTwo`
```

Note that this seems to be a bug in the existing `labels` option. Since this PR just allows the labels option to be turned on or off by default for an association, it doesn't fix this.

spec/e2e/has_many_spec.rb
```diff
@@ -387,13 +387,29 @@
...
       node.knows << friend1
       friend1.knows << friend2
     end
 
     it 'allows passing only a hash of options when naming node/rel is not needed' do
       expect(node.knows(rel_length: :any).to_a).to match_array([friend1, friend2])
     end
+
+    it 'honors default options' do
+      expect(node.knows.instance_variable_get('@query_proxy').instance_variable_get('@association_labels')).to eq(nil)
```
cheerfulstoic: I wonder if this could be testing by calling `to_cypher` on the `QueryProxy` and doing a `MATCH` on the string perhaps

spec/e2e/has_one_spec.rb
cheerfulstoic: If it's working in the console I'm fine with leaving this out for now.  There are definitely some hairy parts with stubbing in the gem (partially because of stubbing and partially because of various interdependencies within the modules, I think)

spec/e2e/has_one_spec.rb
```diff
@@ -206,6 +208,41 @@
...
       expect(employee.manager(rel_length: 1)).to eq(manager)
     end
+
+    # since chainable: true is an option, this test both checks to see that default options
+    # are honored, and checks to make sure the provided options are merged into the default options
+    it 'honors default options' do
+      expect(
+        employee.manager(chainable: true)
+          .instance_variable_get('@query_proxy')
+          .instance_variable_get('@association_labels')
```
cheerfulstoic: Again on these, would like to investigate the possibility of checking the Cypher query.  It might be a bit hacky, but perhaps less hacky then getting instance variables since that spoils encapsulation, I think

spec/e2e/has_one_spec.rb
thefliik: @cheerfulstoic Should I leave the spec there but keep it commented out (as a note for people in the future)? Or should I just delete the commented out spec?

thefliik: Nvm, while addressing

> I wonder if this could be testing by calling to_cypher on the QueryProxy and doing a MATCH on the string perhaps

I was able to get the spec working

thefliik:
@cheerfulstoic, assuming the specs pass, this is ready to go

spec/e2e/has_one_spec.rb
thefliik: For the curious: I think the issue was that I misinterpreted the ```NameError: undefined local variable or method `node` ```. The `has_one` method has an arg named `node`, and I thought the error was coming from there. In actuality, I think the error was coming from the node variable in the spec:
`expect(node.favorite_subordinte ...`

I think this was a copy-paste error from the `has_many` version of this spec.

cheerfulstoic: Awesome work, thanks!

PR #1487, opened 2018-03-17
Optimize property declaration

jgaskins:
Previously, the call to `property` was invoking ActiveModel's `respond_to?` method, which is actually quite expensive - it depends on `ActiveModel::AttributeMethods#matched_attribute_method`, which was painfully slow on giant nodes.

With this commit, loading the definition for an enormous node with 234 properties (don't ask) went from 1450ms down to 145ms simply by foregoing the `respond_to?` call with `instance_methods.include?`. On a slower machine, loading those properties took over 7 seconds before this commit and only 169ms after.

The main issue seemed to be that `ActiveModel::AttributeMethods#matched_attribute_method` was invoking `attributes` 25x _per call_, resulting in 5850 hashes generated for 234 properties. You can see this in the `ruby-prof` output here:

screen shot 2018-03-17 at 6 05 06 pm

After this commit, the call stack looks like this:

screen shot 2018-03-17 at 6 08 09 pm

### DIFF
--- a/lib/neo4j/shared/attributes.rb
+++ b/lib/neo4j/shared/attributes.rb
@@ -165,8 +165,10 @@ module Neo4j::Shared
       #
       # @return [false, String] False or the conflicting method name
       def dangerous_attribute?(name)
+        methods = instance_methods
+
         attribute_methods(name).detect do |method_name|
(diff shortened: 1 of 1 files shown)


cheerfulstoic:
Awesome, thanks for an awesome first PR!

cheerfulstoic:
Released as 9.1.5!  This is in a bit of the code which I don't think has been updated for a while, so let me know if you need it backported

PR #1488, opened 2018-03-19
fix: add helpful error message to AssociationTree

thefliik:
Add helpful error message when trying to eager load "past" a polymorphic association. Currently, trying to eager load past a polymorphic association in the console receives:

```
NoMethodError: undefined method `associations' for nil:NilClass
	from (irb):2
```

Which is difficult to diagnose.

### DIFF
--- a/lib/neo4j/active_node/query/query_proxy_eager_loading.rb
+++ b/lib/neo4j/active_node/query/query_proxy_eager_loading.rb
@@ -23,6 +23,8 @@ module Neo4j
           end
 
           def add_spec(spec)
+            fail "Cannot eager load \"past\" a polymorphic association. (since the association can return multiple models, we don't how to handle the \"#{spec}\" association." unless model
+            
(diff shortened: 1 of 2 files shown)


thefliik:
Hmm. Apparently RuboCop doesn't like the long text.

cheerfulstoic:
Great, thanks!  I think the message is missing an end parenthesis.  It would be good to have a spec to cover this exception case.

For the long line, you should be able to use backslashes in the string like this:

```ruby
fail "Cannot eager load \"past\" a polymorphic association. \
(since the association can return multiple models, we don't how to handle the \"#{spec}\" association." unless model
```

thefliik:
@cheerfulstoic thanks, I didn't know about using a single `\` inside a string to split it between lines.

Spec added.

cheerfulstoic:
🎉

cheerfulstoic:
Released as 9.1.7

PR #1490, opened 2018-03-25
Optimize ActiveNode deserialization

jgaskins:
I had some relatively simple queries returning large data sets taking several seconds on my machine. `Benchmark.measure` indicates that CPU time in Ruby was 87% of the time spent in the query (76% of which was underneath `Neo4j::Shared::Initialize#convert_and_assign_attributes`), so I wanted to see what I could do to streamline it.

Several micro-optimizations involved here:
- Optimizing short-circuit conditionals by putting the condition most likely to trigger the short circuit first
- Set inclusion over array inclusion
- Extract instance variables+attr_readers instead of hash dereferencing for things like `Neo4j::Shared::DeclaredProperty#type`, `#typecaster`, and `#default`
- Remove intermediate-array generation

With these micro-optimizations, because these methods are called so many times, I got a 12-18% performance boost on some of my simpler queries where the DB isn't working hard but there's a lot of data to deserialize (like `MATCH (foo:Foo)-[bar:BAR]-(baz:Baz) RETURN foo, bar, baz`). For queries that are not bottlenecked by Ruby (for example, waiting longer for the DB response), this PR has a reduced effect (but still measurable). The idea here was to improve throughput for things like Sidekiq jobs by releasing the GIL more quickly, and for longer-running queries it will still have a throughput benefit (though less likely a latency benefit) depending primarily on the DB load.

There are other ways I think this can be improved, but I'll file a separate issue.

### DIFF
--- a/lib/neo4j/shared/declared_property.rb
+++ b/lib/neo4j/shared/declared_property.rb
@@ -10,7 +10,7 @@ module Neo4j::Shared
     attr_reader :name, :name_string, :name_sym, :options, :magic_typecaster, :type, :typecaster, :default_value
     alias default default_value
 
-    def initialize(name, type: nil, **options)
+    def initialize(name, **options)
       fail IllegalPropertyError, "#{name} is an illegal property" if ILLEGAL_PROPS.include?(name.to_s)
       fail TypeError, "can't convert #{name.class} into Symbol" unless name.respond_to?(:to_sym)
       @name = @name_sym = name.to_sym
--- a/lib/neo4j/shared/declared_property.rb
+++ b/lib/neo4j/shared/declared_property.rb
@@ -80,7 +80,7 @@ module Neo4j::Shared
 
     # Tweaks properties
     def register_magic_properties
-      @type ||= options[:type] = Neo4j::Config.timestamp_type if timestamp_prop?
+      @type ||= Neo4j::Config.timestamp_type if timestamp_prop?
 
       register_magic_typecaster
       register_type_converter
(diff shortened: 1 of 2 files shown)


cheerfulstoic:
Nice, thanks!  Though it looks like [Travis builds](https://travis-ci.org/neo4jrb/neo4j/builds/357939495?utm_source=github_status&utm_medium=notification) are failing with a number of errors.  Maybe there's a logic error in there now?

jgaskins:
Ah, crap. I have a feeling that's from the part where I [extracted some of the options from the hash into keyword args](https://github.com/neo4jrb/neo4j/pull/1490/commits/30b36b45609e0f548de6ecaf2fc7589155f0f83b#diff-9a97d25920ddc7f617762713af1073b4R13) because the failures look like missing types and type converters. I wondered if that was going to cause a problem, tbh. I'll take another look.

cheerfulstoic:
:+1: Cheers

subvertallchris:
I applaud you for joining the performance fight and look forward to seeing this and others like it get merged in!

cheerfulstoic:
Yay, it's passing! 🎉

Are you happy with the state of things for now?  Is it still reasonably faster (like, statistically significantly) for cases with lots of data?

jgaskins:
@cheerfulstoic Yep, it's still in roughly the same spot, performance-wise, that it was in after my changes but before I got the tests passing.

I'm sure there are still things that can be improved about it, but it's in a pretty decent spot. :-)

cheerfulstoic:
:tada:

cheerfulstoic:
Released as 9.1.8!

Issue #1486, opened 2018-03-16
Neo4j::ActiveNode::Query::QueryProxy bug with .first and .last methods


kopylovvlad:
QueryProxy's methods .first and .last always return first element. It makes me sad

Additional information which could be helpful if relevant to your issue:

### Code example (inline, gist, or repo)
```ruby
users = User.all
users.size
# => 12 
users.class 
# => Neo4j::ActiveNode::Query::QueryProxy 
users.first.id
# User 
#  MATCH (n:`User`)
#  RETURN n
#  ORDER BY ID(n)
#  LIMIT {limit_1} | {:limit_1=>1}
# HTTP REQUEST: 3ms POST http://localhost:7575/db/data/transaction/commit (1 bytes)
# => 47 

users.last.id
# User 
#  MATCH (n:`User`)
#  RETURN n
#  ORDER BY ID(n)
#  LIMIT {limit_1} | {:limit_1=>1}
# HTTP REQUEST: 2ms POST http://localhost:7575/db/data/transaction/commit (1 bytes)
# => 47 

users.to_a.first.id
# => 47 
users.to_a.last.id
# => 393 
users[0].id
# => 47 
users[-1].id
# => 393 
```

### Runtime information:

Neo4j database version:
`neo4j` gem version: 9.1.4
`neo4j-core` gem version: 8.1.1
`ruby` verion: 2.4.1

thefliik:
Hmmm, looking at the cypher you included with your issue, it appears the problem is that both `user.last.id` and `user.first.id` are `ORDER BY ID(n)`.

I imagine a fix is updating `user.first.id` to be `ORDER BY ID(n) ASC` and `user.last.id` to be `ORDER BY ID(n) DESC`.

I think I've run into this issue a few times myself, though I never diagnosed what was going on.

thefliik:
Hmm, looks like the `#last()` definition already is suppose to be ordering in DESC. https://github.com/neo4jrb/neo4j/blob/d3a2412a45c551225b24f4a21760183c3f7e561e/lib/neo4j/active_node/query_methods.rb#L19-L22

thefliik:
Ah ok, so while the _active node_ `.last()` query method seems to correctly order DESC, the _query proxy_ `.last()` query method does not appear to do this.

[...]

PR #1494, opened 2018-03-31
fix .first and .last methods for QueryProxyMethods

kopylovvlad:
Fixe for # https://github.com/neo4jrb/neo4j/issues/1486


lib/neo4j/active_node/query/query_proxy_methods.rb
```diff
@@ -255,11 +255,12 @@ def where_clause_params
         end
 
         def first_and_last(func, target)
+          ord_prop = (func == LAST ? {order_property => :DESC} : order_property)
```
cheerfulstoic: Since `ord_prop` is only used inside the `else`, it should be able to be defined there.  Probably not a big deal, but it should reduce computation in some cases.  I'll make this change after merging.

cheerfulstoic:
Released as 9.2.1.  Thanks for a great first commit!

PR #1498, opened 2018-04-12
added test to demonstrate issue 1466

anamba:
Hey, why wait? Here's that first pull request I promised.

Not sure if this is the right place or right way to test this case, but I thought it'd be a good way to show what I'm seeing.

### DIFF
--- a/spec/e2e/undeclared_properties_spec.rb
+++ b/spec/e2e/undeclared_properties_spec.rb
@@ -42,6 +42,20 @@ describe Neo4j::ActiveNode do
       expect(person[:foo]).to eq(123)
       expect(get_value_from_db(person, :foo)).to eq(123)
     end
+
+    context 'with Timestamps included' do
+      before(:each) do
+        stub_active_node_class('PersonWithTimestamps') do
+          include Neo4j::UndeclaredProperties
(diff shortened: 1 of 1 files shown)


cheerfulstoic:
Hey!  Thanks for this and sorry for not getting back sooner.

I'm not sure that this test is checking the right thing or not.  I tried removing `include Neo4j::UndeclaredProperties` from the `PersonWithTimestamps` definition and it still was failing.

Looking at the `timestamped model` definition (`spec/shared_examples/timestamped_model.rb`), it looks like there are some examples from those examples which depend on the `a` and `b` properties existing.  When I set those properties on the model the tests seem to pass.

Please let me know if I'm missing something

klobuczek:
@anamba indeed the `timestamped model` shared example requires too much context. So far there is a single usage of it. The shared_example would have to first made more generic and reusable. But if you see an issue in the code rather than spec I would suggest to formulate a new spec from scratch to demonstrate it. Unless of course you want to fix the shared_example first which would be appreciated.

anamba:
Wow, blast from the past. :-)  Sorry, but I no longer use this gem (using Elixir for new projects these days).

PR #1499, opened 2018-04-17
Rails 5.2 Support

### LINKED ISSUE #1496: Rails 5.2 support
Just curious where there's any progress on Rails 5.2 support, maybe a branch I can try?

I tried out neo4jrb + 5.2 briefly when rc1 came out, and again for rc2, but it is sadly still unusable. I tried cloning this repo and running the specs to see if there might be some kind of quick and easy fix, but I get thousands of spec failures, most of which are either `NoMethodError: undefined method 'changed?' for nil:NilClass` or `Neo.ClientError.Transaction.TransactionNotFound: Unrecognized transaction id. Transaction may have timed out and been rolled back.` It seems like something is amiss with constraints, but it doesn't seem to be something obvious enough for me to be able to fix on my own...

subvertallchris:
Fixes #1496 

Supporting Rails 5.2 required two small changes.

The less significant is working around ActiveModel's (totally reasonable, IMO) [change](https://github.com/rails/rails/blob/master/activemodel/lib/active_model/dirty.rb#L222) to make the output of `changed_attributes` immutable. Very straightforward. Would like better names for these methods.

The more significant is the workaround for the error causing all of the recent CI failures. The way I did it was to define our own [`mutations_from_database`](https://github.com/rails/rails/blob/master/activemodel/lib/active_model/dirty.rb#L258). I know that we're testing aspects of `ActiveModel::Dirty` but I'm still concerned we might be missing out on expected behavior. I didn't have internet while working on this so I couldn't really dig into it. For that reason, I'd like some careful eyes on this before it's merged in.

### DIFF
--- a/lib/neo4j/shared/initialize.rb
+++ b/lib/neo4j/shared/initialize.rb
@@ -25,6 +25,10 @@ module Neo4j::Shared
       end
     end
 
+    def changed_attributes_clear!
+      @attributes_changed_by_setter = ActiveSupport::HashWithIndifferentAccess.new
+    end
+
     def changed_attributes_selective_clear!(hash_to_clear)
       attributes_changed_by_setter = ActiveSupport::HashWithIndifferentAccess.new(changed_attributes)
       hash_to_clear.keys.each { |k| attributes_changed_by_setter.delete(k) }
--- a/lib/neo4j/shared/initialize.rb
+++ b/lib/neo4j/shared/initialize.rb
@@ -25,10 +25,6 @@ module Neo4j::Shared
       end
     end
 
-    def changed_attributes_clear!
-      @attributes_changed_by_setter = ActiveSupport::HashWithIndifferentAccess.new
-    end
-
     def changed_attributes_selective_clear!(hash_to_clear)
       attributes_changed_by_setter = ActiveSupport::HashWithIndifferentAccess.new(changed_attributes)
       hash_to_clear.keys.each { |k| attributes_changed_by_setter.delete(k) }
--- a/lib/neo4j/shared/initialize.rb
+++ b/lib/neo4j/shared/initialize.rb
@@ -24,5 +24,15 @@ module Neo4j::Shared
         attr[key.freeze] = v
       end
     end
+
+    def changed_attributes_clear!
+      @attributes_changed_by_setter = ActiveSupport::HashWithIndifferentAccess.new
+    end
+
+    def changed_attributes_selective_clear!(hash_to_clear)
+      attributes_changed_by_setter = ActiveSupport::HashWithIndifferentAccess.new(changed_attributes)
+      hash_to_clear.keys.each { |k| attributes_changed_by_setter.delete(k) }
+      @attributes_changed_by_setter = attributes_changed_by_setter
+    end
   end
 end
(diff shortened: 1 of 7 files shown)


subvertallchris:
@cheerfulstoic 

Is there any new JRuby-specific code that I might have overlooked? All of the failures were there.

cheerfulstoic:
Uggg, jRuby.  Yeah, maybe, I'm not sure.

I've been testing it on `ActiveModel` 5.1 and it does seem to have an issue. I just made a commit to add a Travis build to test that.  I've also changed your `changed_attributes_clear!` and used `clear_changes_information` because that is (if I understand correctly) the official method from `ActiveModel` to reset changes.

cheerfulstoic:
Actually, it seems like the errors aren't in jRuby but with MRV 2.1.10...

cheerfulstoic:
Ok, well my change to `clear_changes_information` broke something.  Though I think it highlights the fact that as a gem which works on top of `ActiveModel` that we should be using the `Attributes` API.  The error is `undefined method `forgetting_assignment' for ["prop_with_default", "something"]:Array` and it comes from here:

https://github.com/rails/rails/blob/master/activemodel/lib/active_model/dirty.rb#L270

I think our `@attributes` is supposed to be a set of `Attribute` objects from `ActiveModel`.

Maybe your original fix is the thing to do for getting 5.2.0 compatibility ASAP, but I think there's a systemic, longer-term fix here.  Does that seem right to you?

subvertallchris:
Yeah, I think that’s why I did it the way I did. Really annoying. You are
100% correct: we’d do well to sync our concept of `attributes` with
ActiveModel’s, since that disconnect is the underlying issue. My first
impulse when working on this was to do just that but it’s very clearly a
bigger project that needs some consideration, as it will have far-reaching
consequences.

cheerfulstoic:
Cool, glad we're on the same page.  I reverted my change back to yours and I'm going to add a comment explaining.  I think I might work on a larger refactor to standardize on `ActiveModel` and at the same time make the `id_property` just a plain `ActiveModel` attribute.  It would probably remove a lot of code from the gem and it may even make things more efficient (one can hope... ;) ).  I think that would make a nice 10.0 change

cheerfulstoic:
I just changed the `travis.yml` to test Ruby 2.2.10 instead of 2.1.10.  Rails 5.0 requires at least 2.2.2, so probably `ActiveModel` 5.0 also requires it, but maybe other parts of Rails were using more modern versions of Ruby while `ActiveModel` didn't actually change anything requiring newer versions of Ruby until 5.2.  That's all just a theory, but it probably is not a big deal to require 2.2.2 as well.  Theoretically this might open the gem up to errors when people are using Ruby 2.1 with versions of Rails before 5.0, but I think that's a risk that we'd need to take given the part-time support of the gem ;)

cheerfulstoic:
I put in a conditional statement on if `changed_attributes` is `frozen?` which allows for support for versions before 5.2.  Everything's passed on CI, so I'm going to merge and release.  Thanks!

cheerfulstoic:
Released as 9.2.2, thanks again!

lib/neo4j/shared/initialize.rb
```diff
@@ -24,5 +24,31 @@ def stringify_attributes!(attr, properties)
         attr[key.freeze] = v
       end
     end
+
```
subvertallchris: @cheerfulstoic I like all the changes you made below!

PR #1500, opened 2018-04-27
Rescue LoadError

Grafikart:
This fix an issue I had when the system tried to constantize labels

### DIFF
--- a/lib/neo4j/active_node/node_wrapper.rb
+++ b/lib/neo4j/active_node/node_wrapper.rb
@@ -36,7 +36,7 @@ module Neo4j
 
       def constantized_label(label)
         "#{association_model_namespace}::#{label}".constantize
-      rescue NameError
+      rescue NameError, LoadError
         nil
       end
 


cheerfulstoic:
Thanks for this!

What specifically is the `LoadError` that you're getting?  I'd like to know more about it before we rescue from it.

It would also be good to have a spec which causes this to both explain the issue and to prevent regressions

cheerfulstoic:
When I try that in my Rails app I get a `NameError`.  Maybe it's a version of Rails thing?  What version are you using?

It looks like `LoadError` is what you get when you try to `require` something that isn't there:

```ruby
2.5.1 :003 > require 'aoeundaeound'
Traceback (most recent call last):
        1: from (irb):3
LoadError (cannot load such file -- aoeundaeound)
```

So I feel like there's something I'm not completely understanding.

Also, I believe that it's `ActiveSupport` specifically that loads files based on namespace/class conventions and that's a dependency of the `neo4j` gem, so maybe there is some sort of tool that could be used from the gem to test this.

Grafikart:
@cheerfulstoic Here are the steps to produce a load error on constantize 

1. Create a new rails app (without ORM) `rails new app -O`
2. Inside this app create a simple `Demo` model

app/model/demo.rb
```ruby
class Demo
end
```

3. `rails c`, then `"DEMO".constantize` will throw a LoadError instead of a NameError

cheerfulstoic:
OK, thanks.  I see how there's an issue where two different class names (`Demo` and `DEMO`) map to the same file and then it fails because it can't find `DEMO` in the file.

It's a bit weird to me because it *seems* like `LoadError` comes from Ruby but `ActiveSupport` is using it for a somewhat separate thing.  But it does make me feel better that this is rescuing from a base Ruby exception rather than `ActiveSupport`.  And since the line is so targeted, I don't see a possibility of the other sort of `LoadError` situation happening erroneously here.  So :+1: thanks!

cheerfulstoic:
Released as 9.2.3.  Thanks again!

PR #1505, opened 2018-05-18
made requiring adaptor classes dynamic

klobuczek:
Fixes #
BigDecimalConverter which should store this data type as String in the database.
This pull introduces/changes:
 *  additionally made adapter class loading more dynamic
 *  added a way to specify adaptor classes directly without assigning a type to them


lib/neo4j.rb
```diff
@@ -1,8 +1,6 @@
 require 'neo4j/version'
 
 require 'neo4j-core'
-require 'neo4j/core/cypher_session'
-require 'neo4j/core/query' # From the `neo4j-core` gem
```
klobuczek: I got problems with those requires. They shouldn't be necessary as they are already required in the line above.

cheerfulstoic: Ideally I'd like to bring the code from this PR in as a patch for the current release of the `neo4j` gem.  Does this change work correctly against that (as opposed to `master` of `neo4j-core`)?

lib/neo4j/session_manager.rb
```diff
@@ -69,6 +56,25 @@ def wait_for_value(wait, exception_class)
...
+          require 'neo4j/core/cypher_session/adaptors/http'
+          Neo4j::Core::CypherSession::Adaptors::HTTP
+        when :bolt
+          require 'neo4j/core/cypher_session/adaptors/bolt'
+          Neo4j::Core::CypherSession::Adaptors::Bolt
+        else
+          extra = ' (`server_db` has been replaced by `http` or `bolt`)'
+          fail ArgumentError, "Invalid session type: #{type.inspect}#{extra if type == :server_db}"
+        end
+      end
```
cheerfulstoic: :+1:

lib/neo4j.rb
klobuczek: @cheerfulstoic are you specifically interested in the dynamic loading or the converter fix? If it is the former I still have to do some work in railtie which still loads everything. Now I see I should've done 2 PRs, but bear with me. I'll let you know as well if it can be applied to the current release.

cheerfulstoic: The dynamic loading would probably technically be a minor version bump (backward-compatible change), though I could also see it as a fix because only what is needed is being loaded.  Basically I just wanted to make sure that the `BigDecimnalConverter` fix could be made into a 9.2.4 release so that people could get the fix without needing to update to the major version

klobuczek:
@cheerfulstoic what's your take on too long spec blocks?

cheerfulstoic:
I've used this in other projects:

```yaml
Metrics/ModuleLength:
  Exclude:
    - 'spec/**/*'
```

I think it would be just fine here.

klobuczek:
There is a negative coverage delta for 2 files due to the removal of covered lines. No touched lines are missing a test coverage. And of course a huge delta due to the removal of specs from the statistics which were distorting the result anyway.

lib/neo4j/railtie.rb
```diff
@@ -132,23 +132,15 @@ def register_neo4j_cypher_logging(session_type)
...
-    def subscribe_to_session_type_logging!(session_type, logger_proc)
-      if TYPE_SUBSCRIBERS.key?(session_type.to_sym)
-        TYPE_SUBSCRIBERS[session_type.to_sym].call(&logger_proc)
-      else
-        fail ArgumentError, "Invalid session type: #{session_type.inspect} (expected one of #{TYPE_SUBSCRIBERS.keys.inspect})"
-      end
+    def subscribe_to_session_type_logging!(session_type, options, logger_proc)
+      SessionManager
+        .adaptor_class(session_type, options)
+        .send(session_type.to_sym == :embedded ? :subscribe_to_transaction : :subscribe_to_request, &logger_proc)
```
cheerfulstoic: Did the argument validation get lost here?  Was that on purpose?

klobuczek:
I have added a deprecation warning for embedded_db so I am not sure if this makes it more than a bug fix now.

klobuczek:
@cheerfulstoic could we move ahead on that? It is not ideal to point our production to forks and keep synching the fork with upstream.

Issue #1504, opened 2018-05-15
Fake collections


Grafikart:
I'm encountering a problem with a deep relation and I was wondering if it could be solved within neo4jrb scope.

Sometimes I have specific relations like this

(:Tutorial)<-[:TEACH|:USE]<-(:Technology)
(:Tutorial)<-[:INCLUDE]-(:Chapter)<-[:INCLUDE]-(:serie)

I was wondering if it was possible to declare a fake relation in the model

```
has_many :virtual, :technologies
has_one :virtual, :formation
```

Then we would be able to collect with a specific alias 

```
    self
      .query_as(:t)
      .with(:t)
      .optional_match('rel = (t)-[r:TEACH|:REQUIRE]->(tech:Technology)')
      .with(:t, :tech, :r)
      .optional_match('(t)<-[:INCLUDE*]-(f:Formation)')
      .return('t as tutoriel, collect(tech) as technologies_collection, collect(r) as requirements, f as formation_collection')
```

The model would create automatically the accessors for the subcollections

cheerfulstoic:
I think that there are a lot of ways to go about this ;)

Firstly, you should know about the `rel_length` options (documented [here](http://neo4jrb.readthedocs.io/en/latest/Querying.html#variable-length-relationships))

```ruby
as(:t).branch { technologies(:tech, :r) }.formation(:f, nil, rel_length: :any)
      .pluck(:t,
      technologies_collection: 'collect(tech)',
      requirements: 'collect(r)',
      formation_collection: 'f')
```

But, of course, you would need to define the associations.  The `formations` association should be pretty easy:

```ruby
has_one :in, :formation, type: :INCLUDE
```

For `technologies`, it might be tricker because of the multiple relationships.  I don't remember if we ever implemented this:

```ruby
has_many :out, :technologies, type: [:TEACH, :REQUIRE]
```

If that doesn't work you can probably hack it like this:

```ruby
has_many :out, :technologies, type: "TEACH|:REQUIRE"
```

But maybe you're also wanting something higher level so that you don't need to do the `collect`s.  For the `technologies` association you'd probably be fine because the gem auto-loads associations (though you can do it all with one query using `with_associations`), but the variable length relationship part won't, I don't think, play well with auto-loading because variable-length relationships are defined on query, not on the association.

thefliik:
Not 100% sure I understand the question, but to add to cheerfulstoic's response:

You can create a scopes on your `Tutorial` model such as
```
[...]

PR #1506, opened 2018-05-18
Feat length rel

Grafikart:
Fixes parts of #1504

This pull introduces/changes:
 * has_{many,one} accepts a new option "rel_length"


thefliik:
Sorry, I could have been clearer. `rel_length` is an association getter option. #1485 introduced the ability to specify new default association getter options.

In order to implement the ability to make an association get with `rel_length: 2` (or any other length) by default, I think all you need to update is:

```
module Neo4j::ActiveNode
  module HasN

    class Association
      VALID_ASSOCIATION_OPTION_KEYS = [:type, :origin, :model_class, :rel_class, :dependent, :before, :after, :unique, :labels, :rel_length]
    end

    module ClassMethods

      def define_has_many_methods(name, association_options)
        default_options = association_options.slice(:labels, :rel_length)

        . . .
      end

      def define_has_one_methods(name, association_options)
        default_options = association_options.slice(:labels, :rel_length)

        . . .
      end
    end

  end
end
```

This being said, this method of setting a new default association getter option only touches the association's getter. Unlike what you currently have going on in this PR, someone would still be able to set on

```
class Person
  has_many :out, :friends, type: :FRIEND, rel_length: 2
end
```

But `person.friends << friend` would persist `(person)-[:FRIEND]->(friend)` (in the current PR, looks like you raise an error).

Instead of introducing a separate method for setting a default `rel_length` association getter, maybe add a warning that is logged during dev mode saying that `rel_length` is ignored when setting relationships? This being said, it *seems* like it would be obvious that `rel_length` would be ignored when setting relationships.

Anyway, my 2 cents.

thefliik:
Also, the default association getter options code already has some tests

Grafikart:
@thefliik the end game is to make this work with the eager loading too (I don't know if the default association getter work with this case). 

For me raising an error is important since if we set a `rel_length: 2` we don't necessary expect a direct relation. If I wanted to connect my model directly I would use another association (I would use friends instead of acquaintances).

klobuczek:
`rel_length` is a query concern not association definition concern.

`has_many :out, :friends, type: :FRIEND, nodel_class: 'Person'`
defines automatically an association chain of any length by the virtue of recursion as the target is of the same type. 
Making the addition you suggested would totally change the semantics of associations and introduce a need for difficult if not inconsistent validations.

thefliik:
That's a good point. This change feels like it's mixing associations and scopes.

cheerfulstoic:
I did think about the fact that it would change the nature of associations, though I wasn't sure if it might not be fine, especially if it's an option that you can choose to not use.  I think I'd want to see examples of how it might be a bad idea.

klobuczek:
Such an addition to the association (assuming the rel_lenght 2 means length 1 or 2) would actually mean creating a schema constraint specifying that the database cannot have any paths of friends longer than 2. Something like this is quite difficult to enforce. If I am missing something then please explain what the semantics of the `rel_length` on association would be. 

If you want to have a method for 2nd degree friends then you can defined a scope:
```scope :seconds_degree_friends, ->() { friends(rel_length: {min: 2, max: 2}) }```

On another hand there is already a way to specifiy a deeper retrieval of associtiations with `with_associations`.
e.g. `.with_assoctiations(friends: :friends)` for 2 levels or
`.with_assoctiations(friends: {firiends:: friends})` for 3 levels.
What is not possible is specifying the relationships numerically or specifying unlimited length. I would like to implement that but I couldn't come up with a nice nottation. So far I was only able to think of:
```with_associations(friends: {rel_length: {min: 1, max: 2})``` or ```with_associations(friends: {rel_length: {min: 1})``` for unlimited lenght.  The flow in this notation is that `rel_length` appears in place where other association names can be and thus would have to be a reserved name requiring that there is never an association named `rel_length`. I am open to suggestions.

klobuczek:
@Grafikart the association specification `has_one/has_many` is declarative. The semantics of `rel_length` is unclear. The most intuitive semantics as a constraint is very difficult to implement and probably not practical. Most likely benefit from this implementation can be achieved with a scope without the need to hard code a length into the association definition. Closing for now.

PR #1507, opened 2018-05-20
big decimal converter fix

klobuczek:
Fixes #
Extract the BigDecimalConverter only against 9.2.x
This pull introduces/changes:
 * 
 *


cheerfulstoic:
Oy, and of course it's trying to run against `master`.  I'll merge it in and then change the `.travis.yml` to run against the right branch of `neo4j-core` and then revert that in `master`

PR #1509, opened 2018-05-28
order with limit and with_association bug

klobuczek:
Fixes #
There is a bug when all 3 order, limit and with_associations is used. The limit is applied before order by. 
The main problem around order and `with_associations` is that the `collect` used in `with_associations` reshuffles already sorted set. So the order has to come at the end. Since however, the `with_associations` is very expensive since it potentially touches and returns a large amount of data it would be nice that the result is already limited before it is processed by with_associations, which requires order and limit happening before. 
The approach I have taken here is to repeat the order at the end of the query. Another approach would be to move the `limit` to the very end. Both approaches are functionally the same. But the former is more performant in general case. 
I am open for suggestions either way. But one thing is clear that it cannot stay the way it is as pagination would return totally wrong results.

This pull introduces/changes:
 * 
 *

### DIFF
--- a/lib/neo4j/active_node/query/query_proxy_eager_loading.rb
+++ b/lib/neo4j/active_node/query/query_proxy_eager_loading.rb
@@ -157,11 +157,10 @@ module Neo4j
         end
 
         def build_query
-          before_pluck.pluck(identity, "[#{with_associations_return_clause}]")
+          before_pluck(query_from_association_tree).pluck(identity, "[#{with_associations_return_clause}]")
         end
 
-        def before_pluck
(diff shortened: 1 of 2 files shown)


klobuczek:
@cheerfulstoic could we move ahead on that? It is not ideal to point our production to forks and keep synching the fork with upstream.

cheerfulstoic:
Hey @klobuczek 

Unfortunately I haven't been able to continue maintaining the Neo4j.rb as I have other projects (see [this issue](https://github.com/neo4jrb/neo4j/issues/1516).

I'll be happy to provide advice on whoever would take over.  I have one person who's potentially interested.

PR #1512, opened 2018-07-11
Adding companion instance method with scope

amitsuryavanshi:
Fixes #

This pull introduces/changes:
 * For Every scope defined on model, there will be a instance method with same name.
Most ofter when I define a scope like `scope :active_friends, -> { friends.where(status: 'active') }` and I have to get active friends of a particular user I have to do something like `user.as(:u).active_friends`. With the change in this PR, I will have by default a `user.active_friends` method.


thefliik:
This will also address #1461

lib/neo4j/active_node/scope.rb
```diff
@@ -45,6 +45,10 @@ def scope(name, proc)
             _call_scope_context(eval_context, query_params, proc)
           end
         end
+
+        define_method(name) do |*query_params|
+          as(:n).send(name, *query_params)
```
thefliik: not that it really matters, but perhaps this should be `public_send`

amitsuryavanshi: Okay Sure.

amitsuryavanshi:
Ahh I missed that PR @thefliik, Let me know if you want incorporate this change in your PR, I can close this.

thefliik:
No, the conclusion reached in #1461 was that this is a better solution to the problem.

cheerfulstoic:
Released as 9.3.0

PR #1533, opened 2019-01-17
Query proxy where or

amitsuryavanshi:
Fixes #

This pull introduces/changes:
 * This PR will introduce `.where_or` method to `QueryProxy` objects
This will be a logical or for conditions passed to `.where_or` call
eg.
` User.where_or(name: 'CEO', title: 'sd')`
will produce a query like
```
MATCH (result_user:`User`)
WHERE (result_user.name = {result_user_name}) OR (result_user.title = {result_user_title})
RETURN result_user | {:result_user_name=>"CEO", :result_user_title=>"sd"}
```


amitsuryavanshi:
Closing this PR as this does not sound like a viable option.

PR #1537, opened 2019-02-21
Issue 1531 on assignment of relationship objects,  deleting rels that are required to be deleted

### LINKED ISSUE #1531: user.posts = [same posts as earlier], should not delete all existing posts relations and create similar new ones.
`post1, post2, post3` are  all `Post`s related to a user `tom`, `tom.posts.to_a` returns me array of posts `[post1, post2, post3]`. 
Now if I do `tom.posts = [post1, post2, post3]`, it deletes all existing relationships between `tom` and 'post1, post2, post3' and creates same new relations for `tom` with same posts `post1, post2, post3`.

Expected behaviour would be to not delete existing relations and create same ones. We should delete or create only delta.

### Runtime information:

Neo4j database version:
`neo4j` gem version: 
`neo4j-core` gem version:

amitsuryavanshi:
… 

Fixes #1531 

This pull introduces/changes:
 * Into the way relationships are updated when we assign relationship objects
  eg. `user.posts = [new_posts]` would fist check difference in existing `user.posts` with `new_posts` and create/delete only relevant posts. Earlier implementation would delete all existing posts relationships and create new relationships.


subvertallchris:
Are there performance considerations to think through? It’s been a while since I’ve looked through this code but based on your description, it sounds like this would need to get the IDs of all relationships to make its comparison. If a node has a significant number of existing rels, this could result in a large amount of data coming down the pipe.

amitsuryavanshi:
@subvertallchris I followed activerecord implementation https://github.com/rails/rails/blob/master/activerecord/lib/active_record/associations/collection_association.rb#L244 Assuming they would have done some benchmarking on performance.

subvertallchris:
I still have some questions about how it's loading nodes to make that comparison but I'm so uninvolved that I shouldn't even be poking my nose in at this point. Sorry for the distraction. :-)

amitsuryavanshi:
@subvertallchris It is loading the nodes the same way as earlier. It is just at a few steps earlier than before. Your comments are always welcomed :)

klobuczek:
I had another comment that this might be a breaking change if the relationship had properties which would currently be cleared, but with the new implementation would be preserved.

amitsuryavanshi:
Closing this as this has been addressed with https://github.com/neo4jrb/neo4j/pull/1584.

PR #1547, opened 2019-05-03
Refactor HasnN#association_proxy

### LINKED ISSUE #1529: has_n results get mixed up if objects in previous result are == to each other

yourpalal:
Clean up and comment the HashN#association_proxy method which was pretty convoluted.
Functionally equivalent to #1530, so this is a style-only fix, once that PR has been merged.

 * add comments
 * reduce indentation with an early next when @source_proxy_result_cache
 is falsey
 * rename `hash` variable to `proxy_hash_key`
 * rename `cache` variable to `association_query_results`
 * rename result_cache to source_result_cache
 * create proc on its own line, with simpler code
 * instead of using inject to iterate and also find the return value,
    use each to iterate while setting association_proxy_cache for all
    the objects, then use association_proxy_cache to get the right
    return value.

Technically, this also fixes #1529, but there is another PR that fixes it with a smaller change (#1530)

### DIFF
--- a/lib/neo4j/active_node/has_n.rb
+++ b/lib/neo4j/active_node/has_n.rb
@@ -211,21 +211,32 @@ module Neo4j::ActiveNode
 
     def association_proxy(name, options = {})
       name = name.to_sym
-      hash = association_proxy_hash(name, options)
-      association_proxy_cache_fetch(hash) do
-        if result_cache = self.instance_variable_get('@source_proxy_result_cache')
-          cache = nil
-          result_cache.inject(nil) do |proxy_to_return, object|
-            proxy = fresh_association_proxy(name, options.merge(start_object: object),
-                                            proc { (cache ||= previous_proxy_results_by_previous_id(result_cache, name))[object.neo_id] })
+      proxy_hash_key = association_proxy_hash(name, options)
 
-            object.association_proxy_cache[hash] = proxy
+      association_proxy_cache_fetch(proxy_hash_key) do
+        source_result_cache = self.instance_variable_get('@source_proxy_result_cache')
+        next fresh_association_proxy(name, options) unless source_result_cache
 
-            (self == object ? proxy : proxy_to_return)
-          end
-        else
-          fresh_association_proxy(name, options)
+        # The query to load this association will load objects not just for this node,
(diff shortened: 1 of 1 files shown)


yourpalal:
BTW, CI failed due to rubocop issues:

```
Offenses:
lib/neo4j/active_node/has_n.rb:212:5: C: Metrics/AbcSize: Assignment Branch Condition size for association_proxy is too high. [18.06/18] (http://c2.com/cgi/wiki?AbcMetric)
    def association_proxy(name, options = {}) ...
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
lib/neo4j/active_node/has_n.rb:212:5: C: Metrics/MethodLength: Method has too many lines. [17/15] (https://github.com/bbatsov/ruby-style-guide#short-methods)
    def association_proxy(name, options = {}) ...
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

2 lines too many and 0.06 too many ABC points :/ I guess splitting into 2 methods would do better?

PR #1562, opened 2019-07-23
Add rake as non-development dependency

ttrmw:
Rake is needed in any project requiring the gem, but is not in the
gemspec or Gemfileas a dependency. Most projects have it already, so I guess this
flew under the radar

This pull introduces/changes:
 * removes rake as development dependency
 * adds rake as a fully fledged dependency


klobuczek:
@ttrmw could you explain how rake is required at runtime? An analogous gem `activerecord` does not list it as runtime dependency either. Maybe we have differing interpretations of what runtime means.

PR #1584, opened 2020-01-22
Executing relationship callbacks on relationship deletion and respecting has_one constraint on relationship

amitsuryavanshi:
Fixes :
 - Respecting has_one constraint on relationships.

Introduces :  
- Only deleting relationships that needs to be deleted.
- Executing relationship callbacks on deletion of relationship

This pull introduces/changes:
 * Respecting has_one constraint on relationships
 `User.has_many :snaps` and `Snap.has_one :user` `user_1.snaps = [snap]` and now if we do `user_2.snaps = [snap]` then the relationship between `user_1` and `snap` will be deleted, earlier this relationship would be preserved and we would end up with two relationships from snap to user, which would violate  `Snap.has_one :user`

 * Only deleting relationships that needs to be deleted. 
Given `User.has_many :snaps` and `user.snaps = [snap_1]`. Now if we do `user.snaps = [snap_1, snap_2]` then the relationship between `user` and `snap_1` would not be deleted. Earlier this would delete relationship between `user` and `snap_1` and create again.

 * Executing relationship callbacks on deletion of relationship
`User.has_many :snaps, dependent: :delete` and `user.snaps = [snap_1]`. This dependent callback would be only executed when `user` is deleted. Now this callback would be executed when the relationship between `user` and `snap_1` is also deleted. eg. `user.snaps = [snap_2]` would also delete `snap_1`.

### DIFF
--- a/spec/e2e/association_proxy_spec.rb
+++ b/spec/e2e/association_proxy_spec.rb
@@ -398,16 +398,17 @@ describe 'Association Proxy' do
       end
     end
 
-    it 'raises error in case of inverse has_one rel is enforced' do
+    it 'deletes inverse has_one rel and does not call callbacks in inverse rel' do
       person3 = Person.create(name: '3')
       person2 = Person.create(name: '2', children: [person3])
       person1 = Person.create(name: '1', children: [person2])
       person1.update(children: [person2, person3.id])
 
       expect(person3.as(:p).parent.count).to eq(1)
+      expect{Person.find(person2.id)}.not_to raise_error Neo4j::ActiveNode::Labels::RecordNotFound
     end
 
-    it 'raises error in case of inverse has_one rel is enforced and two relationships with same type' do
+    it 'deletes rel in case of inverse has_one rel and two relationships with same type' do
       person1 = Person.create(name: 'person-1')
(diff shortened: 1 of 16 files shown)


spec/e2e/association_proxy_spec.rb
```diff
@@ -392,28 +398,22 @@ def deep_traversal(person)
...
-      Neo4j::Config[:enforce_has_one] = true
+    it 'deletes inverse has_one rel and does not call callbacks in inverse rel' do
       person3 = Person.create(name: '3')
       person2 = Person.create(name: '2', children: [person3])
       person1 = Person.create(name: '1', children: [person2])
-      expect { person1.update(children: [person2, person3.id]) }.to raise_error(Neo4j::ActiveNode::HasN::HasOneConstraintError)
+      person1.update(children: [person2, person3.id])
+
+      expect(person3.as(:p).parent.count).to eq(1)
+      expect{Person.find(person2.id)}.not_to raise_error Neo4j::ActiveNode::Labels::RecordNotFound
```
klobuczek: this is a trivial codeclimate issue. Aren't you using any formatter?

PR #1603, opened 2020-05-05
Allow specify column(s) to touch

JacobGalati:
Fixes #

This pull introduces/changes:
 * Implementing similar `touch` functionality to [ActiveRecord ](https://apidock.com/rails/ActiveRecord/Timestamp/touch)
 * Allows specifying column(s) to touch in addition to `updated_at`.
 * If no attributes are passed, it will simply touch `updated_at`

### DIFF
--- a/lib/active_graph/shared/callbacks.rb
+++ b/lib/active_graph/shared/callbacks.rb
@@ -30,7 +30,7 @@ module ActiveGraph
         raise
       end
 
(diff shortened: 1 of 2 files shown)


JacobGalati:
ActiveRecord's `touch` does not trigger `update` callbacks, but ActiveGraph does. 
I was considering changing that as well, but that would introduce backwards compatibility issues.

PR #1612, opened 2020-07-16
Issue-1611 Fix in updating relations with ids

### LINKED ISSUE #1611: Relational query in wrong order
Suppose a book model `has_many` pages like this
```ruby
class Book
  has_many :out, :pages, type: :has, model_class: :Page
end
```
`book.page_ids`  can return an array of unique ids of related pages instances
`book.page_ids = [....]; book.save`  is supposed to update its relation, as it does in v9.x.

But in v10.0, this doesn't work, since it will issue delete queries that delete existing relation only **AFTER** issuing create queries.
This causes trouble if some id in the array is included in the existing relations.
In an extreme case, if you pass the exact same id array, then this will completely destroy existing relations.

This can be solved by just reversing the order of relational queries.

### Runtime information:

Neo4j database version: neo4j 3.4.0
`neo4j` gem version: v10.0
`neo4j-core` gem version:

amitsuryavanshi:
Fixes #1611 

This pull introduces/changes:
 * The way we get what relations to delete on update of relationships
  In case there was a string ID property, the `to_i` on passed id parameter would cause it be inconsistent with IDs of existing relations and we would end up with deleting all existing relations.

### DIFF
--- a/lib/active_graph/node/query/query_proxy_methods_of_mass_updating.rb
+++ b/lib/active_graph/node/query/query_proxy_methods_of_mass_updating.rb
@@ -52,7 +52,7 @@ module ActiveGraph
           node_or_nodes = Array(node_or_nodes).map { |arg| arg.is_a?(ActiveGraph::Node) ? arg : @model.find(arg) }
           original_ids = self.pluck(:id)
           new_nodes = add_rels(node_or_nodes, original_ids)
-          delete_rels_for_nodes(original_ids, node_or_nodes)
+          delete_rels_for_nodes(original_ids, node_or_nodes.collect(&:id))
           new_nodes
         end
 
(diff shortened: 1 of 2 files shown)


lib/active_graph/node/query/query_proxy_methods_of_mass_updating.rb
```diff
@@ -49,31 +49,20 @@ def delete_all_rels
         # Deletes the relationships between all nodes for the last step in the QueryProxy chain and replaces them with relationships to the given nodes.
         # Executed in the database, callbacks will not be run.
         def replace_with(node_or_nodes)
-          node_hash = idify_hash(node_or_nodes)
+          node_or_nodes = Array(node_or_nodes).map { |arg| arg.is_a?(ActiveGraph::Node) ? arg : @model.find(arg) }
```
klobuczek, line 52:
```suggestion
          nodes = Array(node_or_nodes).map { |arg| arg.is_a?(ActiveGraph::Node) ? arg : @model.find(arg) }
```
This is guaranteed to be an array

PR #1617, opened 2020-07-26
Don't use local path for the activegraph gem

ekampp:
This is referenced in the [official installation documentation](https://neo4jrb.readthedocs.io/en/stable/Setup.html#generating-a-new-app). It should not assume that there is any gems installed in local paths.

Alternatively, the `-B` option should be added to the documentation, and information added there as well about how to fix the Gemfile befire running `bundle install`.


klobuczek:
@ekampp Thanks. This was a glitch. It should've never been committed.

PR #1627, opened 2020-09-24
Don't use exceptions as flow control in for_clause

speedshopnate:
Hi, this is @nateberkopec's alternate account.

I was noticing ~half a dozen exceptions per page being silently raised and rescued on an app. Raising exceptions is pretty slow, much slower than just asking if the object responds or not.


klobuczek:
@speedshopnate thanks for that!

klobuczek (approved):

PR #1630, opened 2020-10-30
Fix errors in example code

malandrina:
:wave: Hi there! I encountered a couple minor errors in the "Driver Instance" docs and took the opportunity to correct them.


klobuczek:
Thank you very much @malandrina. It was 2 a.m. when I wrote those docs. :)

PR #1636, opened 2020-12-07
Prevent after_commit from being called multiple times

joshjordan:
Ran into this while upgrading from neo4jrb 9 to activegraph. Our rails model test detected that `after_commit` hooks were being called multiple times. Something like this failed:

```
it "schedules do not disturb after the flag is changed" do
      expect(@user).to receive(:schedule_notification)

      @user.update_attributes(
        pending_notification: true
      )
end
```

where our User model had:

```
  after_update_commit do
    self.schedule_notification
  end
```

Our test failed claiming that `schedule_notification` was being called twice, and I went on to do some logging in `activegraph` to find that there was no guard against multiple `close` calls on a transaction, and `auto_closable` is causing `close` to be called both as part of successfully running the transaction and as part of auto-closing.

This pull introduces/changes:
 * Prevent erroneous calls to `after_commit` multiple times when a single transaction is closing
 * Prevent unnecessary calls, which add overhead, to other `close` functions including the `super` method from `neo4j-ruby-driver`
* Minor fix to README, which tripped me up, because it does not reflect what the port that `set_default_driver` is acutally attempting to use

### DIFF
--- a/spec/README.md
+++ b/spec/README.md
@@ -1,10 +1,9 @@
 ## Configuration
 
-The specs are configured to run against `http://localhost:7474` by default. In order to change this in your local setup you can set the `NEO4J_URL` environment variable on your system to suit your needs.
+The specs are configured to run against `bolt://localhost:6998` by default. In order to change this in your local setup you can set the `NEO4J_URL` environment variable on your system to suit your needs.
 
 To make this easier, the neo4j spec suite allows you to add a `.env` file locally to configure this. For example, to set the `NEO4J_URL` you simply need to add a `.env` file that looks like this:
 
 ```
-NEO4J_URL=http://localhost:7475
+NEO4J_URL=bolt://localhost:7687
 ```
-
--- a/spec/active_graph/base_spec.rb
+++ b/spec/active_graph/base_spec.rb
@@ -135,47 +135,47 @@ describe ActiveGraph::Base do
 
     describe 'after_commit hook' do
       it 'gets called when the root transaction is closed' do
-        data = false
+        data = 0
         expect do
           subject.transaction do |tx1|
             subject.transaction do |tx2|
               subject.transaction do |tx3|
-                tx3.after_commit { data = true }
+                tx3.after_commit { data += 1 }
               end
             end
           end
-        end.to change { data }.to(true)
-        expect(data).to be_truthy
+        end.to change { data }.to(1)
+        expect(data).to eq(1)
       end
 
       it 'is ignored when the root transaction fails' do
-        data = false
+        data = 0
         expect do
           subject.transaction do |tx1|
             subject.transaction do |tx2|
               subject.transaction do |tx3|
-                tx3.after_commit { data = true }
+                tx3.after_commit { data += 1 }
               end
             end
             tx1.failure
           end
         end.not_to change { data }
-        expect(data).to be_falsey
+        expect(data).to eq(0)
       end
 
       it 'is ignored when a child transaction fails' do
-        data = false
+        data = 0
         expect do
           subject.transaction do |tx1|
             subject.transaction do |tx2|
               subject.transaction do |tx3|
-                tx3.after_commit { data = true }
+                tx3.after_commit { data += 1 }
                 tx3.failure
               end
             end
           end
         end.not_to change { data }
-        expect(data).to be_falsey
+        expect(data).to eq(0)
       end
     end
   end
--- a/lib/active_graph/transaction.rb
+++ b/lib/active_graph/transaction.rb
@@ -6,6 +6,9 @@ module ActiveGraph
     end
 
     def close
+      return if @closed
+      @closed = true
+
       success
       super
       after_commit_registry.each(&:call) unless @failure


spec/README.md
```diff
@@ -1,10 +1,9 @@
 ## Configuration
 
-The specs are configured to run against `http://localhost:7474` by default. In order to change this in your local setup you can set the `NEO4J_URL` environment variable on your system to suit your needs.
+The specs are configured to run against `bolt://localhost:6998` by default. In order to change this in your local setup you can set the `NEO4J_URL` environment variable on your system to suit your needs.
 
 To make this easier, the neo4j spec suite allows you to add a `.env` file locally to configure this. For example, to set the `NEO4J_URL` you simply need to add a `.env` file that looks like this:
 
 ```
-NEO4J_URL=http://localhost:7475
+NEO4J_URL=bolt://localhost:7687
```
joshjordan, line 8: The previous value did not for me, failing with
```
Neo4j::Driver::Exceptions::ClientException:
  Unsupported URI scheme: http
```

So, updated to a good default for bolt instead.

spec/active_graph/base_spec.rb
```diff
@@ -135,47 +135,47 @@ def get_object_by_id(id)
...
             subject.transaction do |tx2|
               subject.transaction do |tx3|
-                tx3.after_commit { data = true }
+                tx3.after_commit { data += 1 }
               end
             end
           end
-        end.to change { data }.to(true)
-        expect(data).to be_truthy
+        end.to change { data }.to(1)
```
joshjordan, line 147: This error was not caught before because to the test only expected `data` to equal `true`, rather than checking the number of times it was called.

klobuczek:
@joshjordan thanks for your contribution. However, I don't see how the transaction could be closed multiple times. Are you calling explicitly `close` in addition to autoclose? The changed specs pass even without the changes to  `ActiveGraph::Transaction` so we need to understand a bit more here.
Please note that activegraph does not support explicit transactions anymore as neo4j(rb) did. Only transaction functions (lambdas in ruby) are supported.

joshjordan:
No, not calling close explicitly. I’m confused — I discovered this in our
neo4j rails project, but a fresh check out of active graph  on my machine
with this spec fails because data equals 2 at the end of the test. Why
would our tests be producing different results?

joshjordan:
If I removed my change in `transaction.rb`:

```
$ NEO4J_URL=bolt://localhost:7687 rspec spec/active_graph/base_spec.rb:136
Run options: include {:locations=>{"./spec/active_graph/base_spec.rb"=>[136]}}
F..

Failures:

  1) ActiveGraph::Base transactions after_commit hook gets called when the root transaction is closed
     Failure/Error:
       expect do
         subject.transaction do |tx1|
           subject.transaction do |tx2|
             subject.transaction do |tx3|
               tx3.after_commit { data += 1 }
             end
           end
         end
       end.to change { data }.to(1)

       expected `data` to have changed to 1, but is now 2
     # ./spec/active_graph/base_spec.rb:139:in `block (4 levels) in <top (required)>'

Finished in 0.21798 seconds (files took 0.72649 seconds to load)
3 examples, 1 failure

Failed examples:

rspec ./spec/active_graph/base_spec.rb:137 # ActiveGraph::Base transactions after_commit hook gets called when the root transaction is closed

Coverage report generated for RSpec to /Users/jjordan/code/activegraph/coverage. 2412 / 5141 LOC (46.92%) covered.
Stopped processing SimpleCov as a previous error not related to SimpleCov has been detected
```

joshjordan:
Full diff to show that's my only local change:

image

joshjordan:
Would you like me to push another branch to Travis to see if CI experiences the same failure?

klobuczek:
yes, please do. It would be good to know where the extra close is coming from.

klobuczek:
I think I know. The session.close is calling tx.close assuming it is idempotent. This is a neo4j driver problem and only on MRI. The jRuby version does not have the issue and hence our discrepancy. The travis build should show it too.

joshjordan:
In any case, it would make sense to prevent double closed even if the
client code (eg my code) attempted to do so, no?

klobuczek:
for the client-side I would say no. There is no api for that. Someone would have to call close within that block, which is not well defined and may have other side effects, but the session.close has to be addressed either in the driver or activegraph.
BTW: Transaction has `open?` method.

joshjordan:
Well, I'm mostly indifferent to where this gets fixed, but for what its
worth: the activegraph read_transaction and write_transaction methods do
expose the `transaction` object, so the API does exist. We've certainly
seen third party code that calls `close` at the end of a transaction when
its actually unnecessary to do so (where open? would return true, but close
would still end up called twice). But, as a larger issue ignoring the
client-side, consider this: even the maintainers of activegraph made this
mistake. Isn't that evidence enough that the code should guard against
accidental double-calls to close?

hut8:
Thanks for raising this. We just discovered that we've been double-emailing users from one of our after commit hooks due to this. Ouch.

Any word on when this will be merged? It does look like it should be fixed in Active Graph instead of the ruby driver, since it is Active Graph that has the code that's erroneous to repeat. The ruby driver's code appears to be safe if called more than once.

klobuczek:
@joshjordan @hut8 the issue has already been solved in a different way in activegraph 11. I retrofitted the change into activegraph 10 to keep consistency. See https://github.com/neo4jrb/activegraph/commit/0dad770ff84a365f3cbc72fd2138c9e650bd4e9e
Release of 10.0.2 within the next hour.
Thank you again.

joshjordan:
Thanks @klobuczek !

PR #1637, opened 2020-12-07
CI: Prevent multiple close failure example

joshjordan:
Not for merge. Pushing to see how Travis does with this test, which I expect to fail on data = 2 in `rspec spec/active_graph/base_spec.rb:136`


joshjordan:
@klobuczek I know you expected this after we talked, but just to confirm, the error is shown here:

image

https://travis-ci.org/github/neo4jrb/activegraph/jobs/748237336

Re #1636

PR #1638, opened 2020-12-08
Remove unused exception class

joshjordan:
This appears to be unused. We had some code that handled `ConnectionFailedError` in `neoj4rb` 9, and that code was throwing `uninitialized constant ActiveGraph::Core::ConnectionFailedError`. While `ActiveGraph::Core::ConnectionFailedError` was actually written, it was never loaded and the code makes no reference to it.

I expect the test suite to prove out that this file has no effect on the codebase.


klobuczek (approved):

klobuczek:
thx @joshjordan

PR #1641, opened 2020-12-23
Dont clear all associations

joshjordan:
Ran across an odd case in our application that led me to discover this `TODO`. We have some code that sets two associations, e.g:

```
@contact.emails = [Email.find(email_id)]
@contact.owner = current_user
@contact.save!
```

Strangely, `save!` would throw an exception because the contact validation would fail for having no email. And, indeed, after the `@contact.owner=` call, `@contact.emails` would be empty. I found a nasty TODO that turned out to be masking other silent failures in our application as well, where unrelated cached associations would be cleared every time an association setter was called.

Again, I suspect this is not the right fix and I'd like your guidance on what I should do here to make a proper fix. Also, what is the appropriate spec file in which to test this?

### DIFF
--- a/lib/active_graph/node/has_n.rb
+++ b/lib/active_graph/node/has_n.rb
@@ -184,7 +184,8 @@ module ActiveGraph::Node
     end
 
     # Returns the current AssociationProxy cache for the association cache. It is in the format
-    # { :association_name => AssociationProxy}
+    # { hash => AssociationProxy}
+    # where hash is the result of calling #association_proxy_hash with the association name
     # This is so that we
     # * don't need to re-build the QueryProxy objects
     # * also because the QueryProxy object caches it's results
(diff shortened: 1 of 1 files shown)


klobuczek:
Thanks for this. It looks like indeed a problem. I wonder why the original authors decided to keep it like that with a comment and not fixing it. Maybe we both are missing something.
This hasn't been probably detected for that long time because there is a better api usage for your case:
```
@contact.update!(email_ids: [email_id], owner: current_user)
```
The assignment operators save immediately without calling callbacks.

klobuczek:
@amitsuryavanshi could you look at that one?

amitsuryavanshi:
@klobuczek @joshjordan This is a known issue, where entire association cache is cleared when we do a assignment on association. We need to check all possible scenarios with the fix provided.

PR #1643, opened 2021-01-11
ENV-based login/pass/encryption set

EugeneIstomin:
This pull introduces/changes:
 * Add possibility to set login, password, encryption on/off for using migration in non-Rails projects


klobuczek:
Thanks @EugeneIstomin . Have you considered extracting this https://github.com/neo4jrb/activegraph/blob/master/lib/active_graph/railtie.rb#L82-L85 out of railtie.rb? There are many configuration options. Creating an environment variable each time someone needs one of those options doesn't seem like a sustainable solution. It would be great to solve the problem for any possible configuration.

PR #1654, opened 2021-05-27
Fix caching bugs in rails reloading

yourpalal:
Add slightly more aggressive clearing of caches during rails reloads.

Fixes query results being wrapped in stale classes after a reload.

This pull introduces/changes:
 * adds ActiveGraph::Node::Labels::Reloading.prepare_for_unload! method
 * call to prepare_for_unload! in an ActiveSupport::Reloader.before_class_unload callback
 * document use of deprecated/removed method
 * replace MODELS_FOR_LABELS_CACHE.clear with ActiveGraph::Node::Labels.clear_wrapped_models,
    which clears that cache as well as another one!

### DIFF
--- a/lib/active_graph/node/labels/reloading.rb
+++ b/lib/active_graph/node/labels/reloading.rb
@@ -4,15 +4,32 @@ module ActiveGraph::Node::Labels
 
     MODELS_TO_RELOAD = []
 
+    def self.prepare_for_unload!
+      # clear label caches
+      ActiveGraph::Node::Labels.clear_wrapped_models
+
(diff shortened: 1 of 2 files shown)


yourpalal:
@klobuczek the build errored here due to [a failing wget command](https://travis-ci.com/github/neo4jrb/activegraph/jobs/508759012#L266)

> The command "wget https://bootstrap.pypa.io/get-pip.py && python3 get-pip.py --user" failed and exited with 1 during .

I'm not sure if this is just a transient issue, but could the build be re-run?

Issue #1660, opened 2021-08-21
creates_unique on ActiveGraph::Relationship deletes to_node's :out relationship, too?


bobmazanec:
### Summary
`ActiveGraph::Relationship` with `creates_unique` (both `:none` and `:all`) removes existing similar relationship **out** of new relationship's `to_node` 

#### Given
``` ruby
# models/is_response_to.rb
class IsResponseTo
  include ActiveGraph::Relationship

  from_class :Post
  to_class   :Post
  type       :IS_RESPONSE_TO

  creates_unique :none
end

# models/post.rb
class Post 
  include ActiveGraph::Node
  [...]
  has_one  :out, :stimulus,  rel_class: :IsResponseTo
  has_many :in,  :responses, rel_class: :IsResponseTo
end
```

![image](https://user-images.githubusercontent.com/2031462/130331266-a29ccc52-4c16-4803-94a0-0c43e261de45.png)

#### When
I try to create `(new:Post)-[:IS_RESPONSE_TO]->(<Node 27>)` (without affecting `<Node 27>`'s existing existing relationship to `<Node 26>`)

``` ruby
[...]
ActiveGraph::Base.transaction do
  response = Post.create stimulus: @post, [...]
  [...]
end

respond_to do |format|
  if @reacted_to.save
    [...]
```

#### Then
in the log, I see `post4` created as `response` in the controller code, then this surprising/suspicious part
```
 Post#stimulus
  MATCH(post27)
  WHERE(ID(post27) = $ID_post27)
  MATCH(post27)-[r:`IS_RESPONSE_TO`]->(n:`Post`)
  DELETE r | {:ID_post27=>27}
 BOLT:0ms
```

and this expected stuff
```
 Post#stimulus
  MATCH(post4)
  WHERE(ID(post4) = $ID_post4)
  MATCH(post4)-[r:`IS_RESPONSE_TO`]->(n:`Post`)
  DELETE r | {:ID_post4=>4}
BOLT: 24ms

CYPHER 
 MATCH 
   (from_node), 
   (to_node)
 WHERE 
   (ID(from_node) = $from_node_id) AND 
   (ID(to_node) = $to_node_id)
 MERGE (from_node)-[rel:`IS_RESPONSE_TO` {}]->(to_node)
 SET rel.`created_at` = $setter_rel_created_at
 RETURN rel | {:from_node_id=>4, :to_node_id=>27, :setter_rel_created_at=>1629570639}
```

### Expected
[...]

PR #1662, opened 2021-08-29
Match has_one direction before deleting rel

### LINKED ISSUE #1650: Creating relations with ActiveGraph::Relationship may delete other existing :has_one relations from the involved nodes
The problem occurs when using the same relation class for both :in and :out relations.

I am building a tree structure using ActiveGraph::Nodes, and use ActiveGraph::Relationship for the relations between the nodes (to store some information in the relations). 

```
class NodeWithRel 
  include ActiveGraph::Node
  
  has_one  :out, :parent,   rel_class: :NodeRel
  has_many :in,  :subnodes, rel_class: :NodeRel
  
  property :title
end

class NodeRel 
  include ActiveGraph::Relationship
  
  from_class :NodeWithRel
  to_class   :NodeWithRel
end
```
When I add a subnode to an existing node, the existing node's parent relation is deleted when it is saved to the database:

```
irb(main):001:1* n = NodeWithRel.new(title: "root")
=> #<NodeWithRel uuid: nil, title: "root">
irb(main):002:0> n.save
=> true
irb(main):003:0> n1 = NodeWithRel.new(title: "n1")
=> #<NodeWithRel uuid: nil, title: "n1">
irb(main):004:0> n.subnodes << n1
=> #<QueryProxy NodeWithRel#subnodes [#<NodeWithRel uuid: "d66f8b32-961c-4113-9673-3e67c1f14973", title: "n1">]>
irb(main):005:0> n1.parent
=> #<NodeWithRel uuid: "ba965b32-b0a3-444f-b872-8eb0c9912b91", title: "root">
irb(main):006:0> n1_1 = NodeWithRel.new(title: "n1_1")
=> #<NodeWithRel uuid: nil, title: "n1_1">
irb(main):007:0> n1.subnodes << n1_1
=> #<QueryProxy NodeWithRel#subnodes [#<NodeWithRel uuid: "3874ae42-6379-4d8c-9580-0faf75ce8c88", title: "n1_1">]>
irb(main):008:0> n1.parent
=> nil # <-- should be the first node created, with the title "root"
```

This doesn't happen if I use "ordinary" relations instead of ActiveGraph::Relationship:

```
class Node 
  include ActiveGraph::Node
  
  has_one  :out, :parent,   type:   :parent,  model_class: :Node
  has_many :in,  :subnodes, origin: :parent,  model_class: :Node
  
  property :title
end

irb(main):010:0> m = Node.new(title: "root")
=> #<Node uuid: nil, title: "root">
irb(main):011:0> m.save
=> true
irb(main):012:0> m1 = Node.new(title: "m1")
=> #<Node uuid: nil, title: "m1">
irb(main):013:0> m.subnodes << m1
=> #<QueryProxy Node#subnodes [#<Node uuid: "46975b28-9f3b-4edf-947f-87a6f3428be8", title: "m1">]>
irb(main):014:0> m1.parent
=> #<Node uuid: "c6bc7943-3186-46f2-ad6a-8aa804b450a1", title: "root">
irb(main):015:0> m1_1 = Node.new(title: "m1_1")
=> #<Node uuid: nil, title: "m1_1">
irb(main):016:0> m1.subnodes << m1_1
=> #<QueryProxy Node#subnodes [#<Node uuid: "531ee78a-603f-47b0-be2b-bfd1581a0bf7", title: "m1_1">]>
irb(main):017:0> m1.parent
=> #<Node uuid: "c6bc7943-3186-46f2-ad6a-8aa804b450a1", title: "root">
```

I think the reason might be that when the relation is set up, ActiveGraph::Relationship first deletes any existing relations that are :has_one relation.
In active_graph/relationship/persistance.rb, line 55:
```
    def delete_has_one_rel
      to_node.delete_reverse_has_one_relationship(self, :in, from_node) if to_node.persisted?
      from_node.delete_reverse_has_one_relationship(self, :out, to_node) if from_node.persisted?
    end
```
    
In active_graph/node/has_n.rb, line 244:
```
    def delete_reverse_has_one_relationship(relationship, direction, other_node)
      rel = relationship_corresponding_rel(relationship, direction, other_node.class)
      delete_has_one_rel!(rel.last) if rel && rel.last.type == :has_one
    end
```
and line 254:
```
    def relationship_corresponding_rel(relationship, direction, target_class)
      self.class.associations.find do |_key, assoc|
        assoc.relationship_class_name == relationship.class.name ||
          (assoc.relationship_type == relationship.type.to_sym && assoc.target_class == target_class && assoc.direction == direction)
      end
    end
```
When using ActiveGraph::Relationship, #relationship_corresponding_rel doesn't consider the direction of the relation when selecting, so when there are two relations using the same relationship class (as in my tree example), both relationships are returned, and one of them (the last) is deleted by #delete_reverse_has_one_relationship.

Changing line 256 from:
`        assoc.relationship_class_name == relationship.class.name ||
`
to:
`        (assoc.relationship_class_name == relationship.class.name && assoc.direction == direction) ||
`
does fix the problem for me, but I am not sure if this change has any other side effects.

### Runtime information:

Neo4j database version: 4.0.11
`activegraph` gem version: 10.0.2
`neo4j-ruby-driver` gem version: 1.7.4

bobmazanec:
Fixes #1650

When `relationship_corresponding_rel` matches on the rel class name,
it stops on the first match, ignoring the `direction` parameter.

When `target_class` represents the `to_node`, direction is `:in`.

When the relationship is a `has_one`/`has_many` between a pair
of nodes of the same class, the match could find and return
the `to_node`'s `:out` association first, causing
`delete_reverse_has_one_relationship` to delete it.

Checking the direction first prevents that.

Sadly, I worked through #1660 before seeing #1650 and that I had reached the same basic conclusion, only factoring the fix slightly differently than the OP proposed.

### Test runs
I added 4 examples—2 simply show that the 'setup' in the 'real' examples works as expected. The other 2 (really the same example but under both `creates_unique :all` and `creates_unique :none`) show the problem.

#### Before
`1907 examples, 0 failures, 5 pending`

##### Without the fix
```
[...]
1) between nodes of the same class with creates_unique(:all) given (gen1)-->(gen0) #create (gen2)-->(gen1) adds a new node and relationsihp without affecting (gen1)-->(gen0)
     Failure/Error: expect(@gen1.parent).to(be_present, "gen1 lost its parent")
       gen1 lost its parent
     Shared Example Group: :creating_gen2 called from ./spec/e2e/relationship_spec.rb:650
     # ./spec/e2e/relationship_spec.rb:636:in `block (5 levels) in <top (required)>'

  2) between nodes of the same class with creates_unique(:none) given (gen1)-->(gen0) #create (gen2)-->(gen1) adds a new node and relationsihp without affecting (gen1)-->(gen0)
     Failure/Error: expect(@gen1.parent).to(be_present, "gen1 lost its parent")
       gen1 lost its parent
     Shared Example Group: :creating_gen2 called from ./spec/e2e/relationship_spec.rb:650
     # ./spec/e2e/relationship_spec.rb:636:in `block (5 levels) in <top (required)>'

1911 examples, 2 failures, 5 pending
```
##### With the fix
`1911 examples, 0 failures, 5 pending`

### DIFF
--- a/spec/e2e/relationship_spec.rb
+++ b/spec/e2e/relationship_spec.rb
@@ -610,43 +610,39 @@ describe 'Relationship' do
 
     after(:each) { [@gen0, @gen1, @gen2].compact.each(&:destroy) }
 
-    shared_examples_for :creating_gen2 do
-      it "creates (gen1)-->(gen0)" do
-        @gen1 = NodeClass.create parent: @gen0
-
-        expect(@gen1.parent.id).to eq @gen0.id
-
-        expect(@gen0.children.count).to eq 1
-        expect(@gen0.children.first.id).to eq @gen1.id
-      end
-
-      context "given (gen1)-->(gen0)" do
-        before(:each) { @gen1 = NodeClass.create parent: @gen0 }
-
-        describe "#create (gen2)-->(gen1)" do
-          it "adds a new node and relationsihp without affecting (gen1)-->(gen0)" do
-            @gen2 = NodeClass.create parent: @gen1
-
-            expect(@gen2.parent.id).to eq @gen1.id
-
-            expect(@gen1.children.count).to eq 1
-            expect(@gen1.children.first.id).to eq @gen2.id
-
-            expect(@gen1.parent).to(be_present, "gen1 lost its parent")
-            expect(@gen1.parent.id).to eq @gen0.id
-
-            expect(@gen0.children.count).to eq 1
-            expect(@gen0.children.first.id).to eq @gen1.id
-          end
-        end
-      end
-    end
-
     %i(all none).each do |uniqueness|
       context "with creates_unique(:#{uniqueness})" do
         before(:each) { IsParentOf.creates_unique uniqueness }
 
-        include_examples :creating_gen2
+        it "creates (gen1)-->(gen0)" do
+          @gen1 = NodeClass.create parent: @gen0
+
+          expect(@gen1.parent.id).to eq @gen0.id
+
+          expect(@gen0.children.count).to eq 1
+          expect(@gen0.children.first.id).to eq @gen1.id
+        end
+
+        context "given (gen1)-->(gen0)" do
+          before(:each) { @gen1 = NodeClass.create parent: @gen0 }
+
+          describe "#create (gen2)-->(gen1)" do
+            it "adds a new node and relationsihp without affecting (gen1)-->(gen0)" do
+              @gen2 = NodeClass.create parent: @gen1
+
+              expect(@gen2.parent.id).to eq @gen1.id
+
+              expect(@gen1.children.count).to eq 1
+              expect(@gen1.children.first.id).to eq @gen2.id
+
+              expect(@gen1.parent).to(be_present, "gen1 lost its parent")
+              expect(@gen1.parent.id).to eq @gen0.id
+
+              expect(@gen0.children.count).to eq 1
+              expect(@gen0.children.first.id).to eq @gen1.id
+            end
+          end
+        end
       end
     end
   end
--- a/spec/e2e/relationship_spec.rb
+++ b/spec/e2e/relationship_spec.rb
@@ -589,4 +589,65 @@ describe 'Relationship' do
       end
     end
   end
+
+  context "between nodes of the same class" do
+    before(:each) do
+      clear_model_memory_caches
+
+      stub_node_class('NodeClass') do
+        has_one  :out, :parent,   rel_class: :IsParentOf
+        has_many :in,  :children, rel_class: :IsParentOf
+      end
+
+      stub_relationship_class('IsParentOf') do
+        from_class :NodeClass
+        to_class   :NodeClass
+      end
+
+      @gen0 = NodeClass.create
+      @gen1 = @gen2 = nil
+    end
+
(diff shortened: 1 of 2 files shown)


spec/e2e/relationship_spec.rb
```diff
@@ -589,4 +589,65 @@ class ActiveRelSpecTypesInheritedRelClass < ActiveRelSpecTypesAutomaticRelType
...
+        to_class   :NodeClass
+      end
+
+      @gen0 = NodeClass.create
+      @gen1 = @gen2 = nil
+    end
+
+    after(:each) { [@gen0, @gen1, @gen2].compact.each(&:destroy) }
+
+    shared_examples_for :creating_gen2 do
```
amitsuryavanshi, line 613: This is used just once, Do we need a shared example here?

amitsuryavanshi (changes requested):

spec/e2e/relationship_spec.rb
bobmazanec: Nope. I was thinking I wanted to use them in more contexts (e.g., `creates_unique true`, etc.), but then didn't.
I'll just roll them into the loop; I/you/we/someone can expand the loop if/when that's desired.

bobmazanec: 26da404

EthanReinholt2001:
Same thing here

PR #1667, opened 2022-01-04
Zero length path eagerload

### LINKED ISSUE #1666: Allow zero length paths in eagerload
Given,
```
class Person
        has_many :out, :knows, model_class: 'Person', type: nil
        has_many :in, :posts, type: :posts
end

class Post
      has_one :out, :owner, origin: :posts, model_class: 'Person'
end

```

Currently we only allow fixed length or infinite length paths in eager load like `Person.where(id: [1,5,6]).with_association('knows*')` or `Person.where(id: [1,5,6]).with_association('knows*4')`.  If we want to eager load post on persons and the persons that they know, we have to do `Person.where(id: [1,5,6]).with_association(['posts', 'knows*.posts'])`. This results in tow optional matches for person's posts and persons's know's posts. To avoid this we can have `Person.where(id: [1,5,6]).with_association('knows*0...posts')`, which will result in only one optional match for posts, hence shorter and faster query.

amitsuryavanshi:
Fixes #1666

This pull introduces/changes:
 * Allows zero length in eager load.

### DIFF
--- a/activegraph.gemspec
+++ b/activegraph.gemspec
@@ -36,6 +36,7 @@ DESCRIPTION
   s.add_dependency('activesupport', '>= 4.0')
   s.add_dependency('i18n', '!= 1.8.8') # https://github.com/jruby/jruby/issues/6547
   s.add_dependency('orm_adapter', '~> 0.5.0')
+  s.add_dependency('parslet')
   s.add_dependency('sorted_set')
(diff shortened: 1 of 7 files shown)


amitsuryavanshi:
Closing in favour of activegraph-edge gem. This will moved to  activegraph-edge gem

PR #1668, opened 2022-02-22
fix: Update ActiveGraph::Core::Schema to to handle labelless indexes

krzyczak:
When connecting to Neo4j Aura some inexes have labelsOrTypes set to `[]`.
This causes https://github.com/neo4jrb/activegraph/blob/master/lib/active_graph/core/schema.rb#L49 to throw `NoMethodError: undefined method` error.

This pull introduces/changes:
 * Updates ActiveGraph::Core::Schema#label not to throw an error when `row[:labelsOrTypes]` is an empty array. It returns `row["name"]` in such case.


klobuczek:
merged otherwise

PR #1693, opened 2023-04-20
union query support implementation 2

mrhardikjoshi:
Fixes #

This pull introduces/changes:
 * adds specs which for union query implementation
 * add support for doing union in subquery (alternate solution)


lib/active_graph/node/query/query_proxy_link.rb
```diff
@@ -27,6 +48,75 @@ def for_clause(clause, arg, model, *args)
...
+              end
+            end
+
+            def uniq_param_generator_lambda
+              lambda do |proxy_params, proxy_cypher, identity, counter|
+                prefix = "#{identity}_UNION#{counter}_"
+                proxy_params.each_with_object({}) do |(param_name, param_val), new_params|
+                  new_params_key = "#{prefix}#{param_name}".to_sym
+                  new_params[new_params_key] = param_val
+                  proxy_cypher.gsub!(/(?<![`'"])\$#{param_name}(?![^\s)])(?![`'"])/, "$#{new_params_key}")
```
klobuczek, line 114: I was just about to approve the PR when I saw the regular expression. It is better not to handle the '"` at all than to handle random cases.
 E.g. (multiple issues) 
```
3.1.2 :001 > "`a`, $var, `b`".gsub(/(?<![`'"])\$var(?![^\s)])(?![`'"])/, "new_var")
 => "`a`, $var, `b`"
 ```
This problem can be solved with a regular expression (no stack necessary) but the complex regular expression has to ensure that on either side there are even numbers of unescaped quotes.

**Review by klobuczek — CHANGES_REQUESTED** [review]

File: lib/active_graph/core/query.rb [hunk 2]
```diff
@@ -162,8 +162,8 @@ def inspect
       # DETACH DELETE clause
       # @return [Query]
 
-      METHODS = %w[start match optional_match call using where create create_unique merge set on_create_set on_match_set remove unwind delete detach_delete with with_distinct return order skip limit] # rubocop:disable Metrics/LineLength
-      BREAK_METHODS = %(with with_distinct call)
+      METHODS = %w[start match optional_match call using where create create_unique merge set on_create_set on_match_set remove unwind delete detach_delete with with_distinct return order skip limit union call_subquery_start call_subquery_end] # rubocop:disable Metrics/LineLength
```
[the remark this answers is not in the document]
mrhardikjoshi: I thought and did the same initially when I started on this feature.
but unfortunately we cannot change the order of elements of this constant or else the cypher generation gets wrong (as it depends on it). following is an example with one of the failing spec after this change.
```
"MATCH (s:`Student`) LIMIT $limit_5 WITH s  RETURN count(s) AS s"
```

lib/active_graph/node/query/query_proxy_link.rb
```diff
@@ -111,7 +111,7 @@ def uniq_param_generator_lambda
                 proxy_params.each_with_object({}) do |(param_name, param_val), new_params|
                   new_params_key = "#{prefix}#{param_name}".to_sym
                   new_params[new_params_key] = param_val
-                  proxy_cypher.gsub!(/(?<![`'"])\$#{param_name}(?![^\s)])(?![`'"])/, "$#{new_params_key}")
+                  proxy_cypher.gsub!(/\$#{param_name}(?=(?:[^`'"]|'(?:\\.|[^'])*'|"(?:\\.|[^"])*"|`(?:\\.|[^`])*`)*[^'"`]*\z)/, "$#{new_params_key}")
```
klobuczek, line 114: Excellent! Just the `?:` are unnecessary as in lookahead no backreferences are created anyway.

I would've probably used lookbehind as I'm thinking about state machine. When I get to the $var I would like to know in what state I am, quoted or unquoted. But that's just a taste. Both alternatives rely on the assumption that the entire string is syntactically correct.

klobuczek (approved):

lib/active_graph/node/query/query_proxy_link.rb
mrhardikjoshi: Already in lookahead so group won't capture it 👍

klobuczek (approved):

PR #1697, opened 2023-06-18
Remove last traces of Travis CI

leoarnold:


klobuczek (approved):
@leoarnold thank you for the contribution and sorry that it took so long

PR #1698, opened 2023-06-18
Update RuboCop config

leoarnold:
RuboCop v1.52.1 threw some errors, so we rengenerate the `.rubocop_todo.yml` and make minimal changes to get RuboCop working again.

We also add RuboCop to CI to ensure this does not happen again.


klobuczek (approved):

klobuczek:
@leoarnold I feel like we should not have the rubocop_todo.yml as this disables all those checks and new problems of the same type can be added if I understand it correctly. With codeclimate we will only get newly introduced problems on each PR. I'm merging this PR for codeclimate to succeed on master. Please make another PR if you agree..

PR #1706, opened 2023-12-05
fixes #1705 by locking node while Assigning rel

### LINKED ISSUE #1705: `has_one` association allows creation of two association when done concurrently
When a node's has_one association is updated concurrently, it does not respect the one association limit and due to race condition it may result in wrong data. This is very low probable scenario. More details with the spec proving the bug is in below PR

https://github.com/neo4jrb/activegraph/pull/1704

### Runtime information:

`neo4j` gem version: master branch
ruby: ruby 3.1.2p20

mrhardikjoshi:
Fixes #1705 

This pull introduces/changes:
 * makes #replace_with method safe for concurrent use by locking node before read/update

### DIFF
--- a/lib/active_graph/node/query/query_proxy_methods_of_mass_updating.rb
+++ b/lib/active_graph/node/query/query_proxy_methods_of_mass_updating.rb
@@ -50,11 +50,8 @@ module ActiveGraph
         # Executed in the database, callbacks will not be run.
         def replace_with(node_or_nodes)
           node_or_nodes = Array(node_or_nodes).map { |arg| arg.is_a?(ActiveGraph::Node) ? arg : @model.find(arg) }
+          ActiveGraph::Base.lock_node(start_object) unless start_object.new_record?
           original_ids = self.pluck(:id)
-          $concurrency_queue << 'ready' if $concurrency_test # completed read above so send signal in the queue
-          while $concurrency_test && $concurrency_test_wait # wait till other thread has also completed its read
-            sleep 1
-          end
           delete_rels_for_nodes(original_ids, node_or_nodes.collect(&:id))
           add_rels(node_or_nodes, original_ids)
         end
@@ -66,12 +63,13 @@ module ActiveGraph
         end
 
         def delete_rels_for_nodes(original_ids, new_ids)
-          ids = original_ids.select { |id| !new_ids.include?(id) }
-          return unless ids.present?
+          ids_to_be_removed = original_ids - new_ids
+          return unless ids_to_be_removed.present?
+
           if association.dependent
-            start_object.public_send("dependent_#{association.dependent}_callback", association, ids)
+            start_object.public_send("dependent_#{association.dependent}_callback", association, ids_to_be_removed)
           else
-            self.where(id: ids).delete_all_rels
+            self.where(id: ids_to_be_removed).delete_all_rels
           end
         end
 
--- a/lib/active_graph/node/query/query_proxy_methods_of_mass_updating.rb
+++ b/lib/active_graph/node/query/query_proxy_methods_of_mass_updating.rb
@@ -51,6 +51,10 @@ module ActiveGraph
         def replace_with(node_or_nodes)
           node_or_nodes = Array(node_or_nodes).map { |arg| arg.is_a?(ActiveGraph::Node) ? arg : @model.find(arg) }
           original_ids = self.pluck(:id)
+          $concurrency_queue << 'ready' if $concurrency_test # completed read above so send signal in the queue
+          while $concurrency_test && $concurrency_test_wait # wait till other thread has also completed its read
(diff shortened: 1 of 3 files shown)


lib/active_graph/node/query/query_proxy_methods_of_mass_updating.rb
```diff
@@ -62,12 +63,13 @@ def add_rels(node_or_nodes, original_ids)
         end
 
         def delete_rels_for_nodes(original_ids, new_ids)
-          ids = original_ids.select { |id| !new_ids.include?(id) }
```
mrhardikjoshi, line 65: using `-` is faster compared to using `select` loop and then checking with `include?` again

lib/active_graph/transactions.rb
```diff
@@ -28,6 +28,10 @@ def read_transaction(**config, &block)
 
       alias transaction write_transaction
 
+      def lock_node(node)
+        node.as(:n).query.remove('n._LOCK_').exec if tx&.open? || explicit_session&.open?
```
klobuczek: we talked about not reusing the documentation `_LOCK_` name. Although I cannot think of a way those would ever conflict if used properly.

klobuczek: it would be good to ensure that your specs fail with this code:
```suggestion
        node.as(:n).query.exec if tx&.open? || explicit_session&.open?
```

klobuczek: Where is the tx boundary set in specs? What tx do new threads use?

klobuczek: The specs are too complex to be confident without failing version.

klobuczek (changes requested):

lib/active_graph/transactions.rb
mrhardikjoshi: with disabling my code in `lock_node` I get following error so we now have a simpler specs with proper failure without fix.

```
  1) ActiveGraph::Relationship::Persistence::QueryFactory#build! non-subclassed concurrent update does not create duplicate has_one relationship
     Failure/Error: expect(ActiveGraph::Base.query("MATCH (node2:`ToClass`)<-[rel1:`HAS_REL`]-(from_class:`FromClass`) return from_class").to_a.size).to eq(1)
     
       expected: 1
            got: 19
     
       (compared using ==)
     # ./spec/e2e/relationship/persistence/query_factory_spec.rb:126:in `block (5 levels) in <top (required)>'
     ```

Comment (mrhardikjoshi) [reply lib/active_graph/transactions.rb]: @klobuczek every thread has unique tx, I checked their object_id.

File: lib/active_graph/transactions.rb [hunk 3]
```diff
@@ -28,6 +28,10 @@ def read_transaction(**config, &block)
 
       alias transaction write_transaction
 
+      def lock_node(node)
+        node.as(:n).query.remove('n._AGLOCK_').exec if tx&.open? || explicit_session&.open?
```
klobuczek: why the condition? No harm would happen if the condition is not met.

spec/e2e/relationship/persistence/query_factory_spec.rb
```diff
@@ -103,6 +103,30 @@ def self.count
         expect(from_node.reload.to_classes).to be_empty
       end
 
+      context 'concurrent update' do
+        before do
+          allow_any_instance_of(ActiveGraph::Node::Query::QueryProxy).to receive(:replace_with).and_wrap_original do |original, *args|
+            $concurrency_queue << 'ready'
+            Thread.stop
```
klobuczek: I think the right thing would be to stop at `delete_rels_for_nodes` after the rels are read so you can force failure with just 2 threads not 100. And then after adding lock see it passing.

spec/e2e/relationship/persistence/query_factory_spec.rb
```diff
@@ -103,6 +103,30 @@ def self.count
...
+          count = 100
+          $concurrency_queue = Thread::Queue.new
+          threads = count.times.map { Thread.new { to_node.update(from_class: from_node) } }
+          sleep(0.1) until $concurrency_queue.size == count
+          $concurrency_queue.clear
+          threads.each(&:run)
+          threads.each(&:join)
+          expect(ActiveGraph::Base.query("MATCH (node2:`ToClass`)<-[rel1:`HAS_REL`]-(from_class:`FromClass`) return from_class").to_a.size).to eq(1)
+        end
+      end
```
klobuczek: I am not thrilled about this spec. `sleep`, 100 threads, no guarantee of failure without lock, low level api. I think this would be possible async.

klobuczek (changes requested):

klobuczek:
closing as already merged to zeitwerk branch

PR #1709, opened 2023-12-24
Zeitwerk autoload fixes

### LINKED ISSUE #1702: Rails 7.1 Incompatibility
I am running a Rails 7.0 app with ActiveGraph 11.4.0 and see a deprecation warning when the app starts:

> DEPRECATION WARNING: ActiveSupport::PerThreadRegistry is deprecated and will be removed in Rails 7.1.
> Use `Module#thread_mattr_accessor` instead.

It references the line in `application.rb` which calls `require 'active_graph/railtie'`.

It looks like `ActiveSupport::PerThreadRegistry` is included here: https://github.com/neo4jrb/activegraph/blob/master/lib/active_graph/node/scope.rb#L147. Since that block of code references following ActiveRecord, perhaps whatever the latest ActiveRecord is doing could be used there as well?

Thank you.

mrhardikjoshi:
Fixes #1702 

Problem:
1) From Rails application (which uses `active_graph` but do not explicitly load(require)it ) the `neo4j-ruby-driver` and other dependacies of `active_graph` are not loaded. So starting rails environment (by `rails s` , `rails c` or any of the rake task) results in uninitialized constant error of that dependancy.

2) generating new rails application using `rails new myapp -m activegraph.rb` was creating incorrect gemfile.

Solution:
1) Updated `railtie.rb` to require `active_graph` gem. The zeitwerk within `active_graph` will take care of loading all depending files and gems.

2) fixed the template which generated gemfile to use correct operator.

### DIFF
--- a/docs/activegraph.rb
+++ b/docs/activegraph.rb
@@ -1,6 +1,6 @@
 # Usage: rails new myapp -m activegraph.rb
 
-gem 'activegraph', '=> 11.1'
+gem 'activegraph', '>= 11.1'
 
 gem_group :development do
   gem 'neo4j-rake_tasks'
--- a/Gemfile
+++ b/Gemfile
@@ -2,10 +2,6 @@ source 'http://rubygems.org'
 
 gemspec
 
-# gem 'neo4j-ruby-driver', path: '../neo4j-ruby-driver'
-
-# gem 'listen', '< 3.1'
-
 active_model_version = ENV['ACTIVE_MODEL_VERSION']
 gem 'activemodel', "~> #{active_model_version}" if active_model_version&.length&.positive?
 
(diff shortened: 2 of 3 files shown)


docs/activegraph.rb
```diff
@@ -1,6 +1,6 @@
 # Usage: rails new myapp -m activegraph.rb
 
-gem 'activegraph', '=> 11.1'
+gem 'activegraph', '>= 11.1'
```
mrhardikjoshi, line 3: I generated fresh rails application before and after this change to compare exact file difference after this change, [here is its diff](https://github.com/mrhardikjoshi/myapp/pull/1/files)

PR #1712, opened 2023-12-25
Fix single rel caching issue.

huseynov206:
Fixes #

This pull introduces/changes:
 *  Resolved an issue where single relationships were not being cached accurately, leading to performance inefficiencies.

### DIFF
--- a/lib/active_graph/node/has_n.rb
+++ b/lib/active_graph/node/has_n.rb
@@ -111,11 +111,15 @@ module ActiveGraph::Node
 
       def add_to_cache(object, rel = nil)
         (@cached_rels ||= []) << rel if rel
-        (@cached_result ||= []).tap { |results| results << object if object && !results.include?(object) }
+        (@cached_result ||= []).tap { |results| results << object if !results.include?(object) } if object
       end
 
       def rels
         @cached_rels || super.tap { |rels| rels.each { |rel| add_to_cache(nil, rel)  } }
+      end
+
+      def rel
+        rels.first
       end
 
       def cache_query_proxy_result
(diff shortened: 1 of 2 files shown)


lib/active_graph/node/has_n.rb
```diff
@@ -111,13 +111,17 @@ def init_cache
 
       def add_to_cache(object, rel = nil)
         (@cached_rels ||= []) << rel if rel
-        (@cached_result ||= []).tap { |results| results << object if object && !results.include?(object) }
+        (@cached_result ||= []).tap { |results| results << object if !results.include?(object) } if object
```
klobuczek, line 114: do you realize that this change is not equivalent? If `object` is nil then `@cached_result` will not be initialized if nil, while it was before. However this will make a difference only in case of a call with only nil parametesr `add_to_cache(nil, nil)`.

huseynov206: There is a reason why I changed this code. After adding the `rel` function to AssociationProxy, I noticed that the following case is working incorrectly.

```
role = Role.find(xxx)
role.manager.source_rel
role.manager.source
```

Even though the manager has a source, if we call `source` after calling `source_rel`, it returns nil. Because we used the `add_to_cache` method in line `has_n.rb:118`, `@cached_result` was initialized as an empty array then in line `has_n.rb:81` we check `@cached_result`, since it is initialized we use it, i.e. the empty array. To avoid this, I made `@cached_result` initialize only when the object exists. All specs and manual tests passed successfully.

* Object parameter is nil only when we use the `add_to_cache` method for rels.

PR #1723, opened 2024-09-23
Update rollback syntax in Testing doc

danielmbarlow:
When using the suggested code from [the documentation](https://neo4jrb.readthedocs.io/en/stable/Testing.html):

```ruby
# For the `neo4j` gem
config.around do |example|
  ActiveGraph::Base.transaction do |tx|
    example.run
    tx.failure
  end
end
```

We get this error:

```ruby
NoMethodError:
       undefined method `failure' for an instance of Neo4j::Driver::Internal::InternalTransaction
```

The syntax is apparently now `tx.rollback`, however when the transaction has already been rolled back we get this error

```ruby
     Neo4j::Driver::Exceptions::ClientException:
       Can't rollback, transaction has been rolled back
```

This pull request changes the documention to recommend:
 * Usage of the new syntax for rollback
 * A check to see if the transaction is open before rolling it back

### DIFF
--- a/docs/Testing.rst
+++ b/docs/Testing.rst
@@ -86,7 +86,7 @@ If you are using RSpec you can perform tests in a transaction as you would using
   config.around do |example|
     ActiveGraph::Base.transaction do |tx|
       example.run
-      tx.failure
+      tx.rollback if tx.open?
     end
   end
 


danielmbarlow:
Let me know if there's anything you would like me to change in this little PR.

klobuczek:
Thanks @danielmbarlow. I just looked at your PR this week and was just surprised about the condition  `if tx.open?`. Did you have to do it because of some error? In the context of activegraph the transaction should be always open if in scope.

PR #1735, opened 2026-03-03
Use driver version 6

nsauk:
This PR makes activegraph use neo4j driver version 6.


nsauk:
SortedSet was removed because of https://github.com/knu/sorted_set/issues/18

lib/active_graph/transactions.rb
```diff
@@ -14,31 +14,33 @@ def session(**session_config)
...
 
       def send_transaction(method, **config, &block)
-        return yield tx if tx&.open?
+        return yield tx if tx
         return run_transaction_work(explicit_session, method, **config, &block) if explicit_session&.open?
-        driver.session do |session|
+        driver.session(bookmarks: last_bookmarks) do |session|
           run_transaction_work(session, method, **config, &block)
+        ensure
+          self.last_bookmarks = session.last_bookmarks
```
klobuczek: I would prefer not to deviate too much from the driver API. The user should create an explicit session if they want to get bookmarks.

lib/active_graph/transactions.rb
```diff
@@ -14,31 +14,33 @@ def session(**session_config)
...
       end
 
       private
 
       def send_transaction(method, **config, &block)
-        return yield tx if tx&.open?
+        return yield tx if tx
         return run_transaction_work(explicit_session, method, **config, &block) if explicit_session&.open?
-        driver.session do |session|
+        driver.session(bookmarks: last_bookmarks) do |session|
```
klobuczek: This unnecessarily chains all implicit sessions, potentially degrading performance due to routing to up-to-date servers only.

klobuczek (changes requested):

lib/active_graph/migrations/runner.rb
```diff
@@ -11,12 +11,14 @@ class Runner
...
 
+      attr_reader :up_versions
+
       def initialize(options = {})
         @silenced = options[:silenced] || !!ENV['MIGRATIONS_SILENCED']
         label = SchemaMigration.mapped_label
         label.create_constraint(:migration_id, type: :unique) unless label.constraint?(:migration_id)
-        @schema_migrations = SchemaMigration.all.to_a
-        @up_versions = SortedSet.new(@schema_migrations.map(&:migration_id))
+        @schema_migrations = SchemaMigration.all.to_a.sort_by(&:migration_id)
```
klobuczek:
```suggestion
        @schema_migrations = SchemaMigration.order(:migration_id).to_a
```

activegraph.gemspec
```diff
@@ -31,15 +31,13 @@ DESCRIPTION
   }
 
   s.add_dependency('activemodel')
+  s.add_dependency('benchmark')
```
klobuczek, line 34: Is this really needed as runtime dependency?

lib/active_graph/migrations/runner.rb
klobuczek: Additionally, remove the comparator <=> from SchemaMigration since it's not used anymore.

klobuczek (changes requested):

activegraph.gemspec
klobuczek: It is in activegraph.rb so not to investigate in this PR.

klobuczek (approved):

PR #1737, opened 2026-03-18
Fix rollback exception propagating

nsauk:
Changes in `Transaction#rollback` #1736 lead to the issue with a dependent app's ability to handle failed validation, resulting in HTTP 200 in places where HTTP 422 was before. This PR restores the behavior.

### DIFF
--- a/spec/active_graph/transactions_spec.rb
+++ b/spec/active_graph/transactions_spec.rb
@@ -122,4 +122,14 @@ describe ActiveGraph::Transactions do
       end
     end
   end
+
+  describe 'rollback in nested transaction' do
+    it 'returns false from the inner transaction when rollback is called' do
+      inner_result = nil
+      ActiveGraph::Base.transaction do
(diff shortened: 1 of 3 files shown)


spec/active_graph/transactions_spec.rb
```diff
@@ -122,4 +122,22 @@
...
+      stub_node_class('ValidatedStudent') do
+        property :name
+        validates :name, presence: true
+      end
+
+      student = ValidatedStudent.create!(name: 'Alice')
+
+      errors = ActiveGraph::Base.transaction do
+        student.update(name: nil)
+        student.errors.full_messages
```
nsauk, line 137: This may look unnatural, but graphiti performs validations inside the transaction.

klobuczek:
@nsauk I created an alternative way to solve it https://github.com/neo4jrb/activegraph/pull/1738. It fixes my faulty refactoring in the previous PR.

PR #1741, opened 2026-07-18
Add db_runtime instrumentation for ActiveGraph

### LINKED ISSUE #1291: Add `db_runtime` support to Neo4jrb
In `ActiveRecord`, when a call is made to a database, the execution time for that query is recorded and returned as payload information. This is typically used with `ActiveSupport::Notifications` to deliver more information regarding the total round trip time your application takes to respond to requests. Unfortunately, this is missing in `ActiveNode`/`ActiveRel`.

It would be awesome if we could add this into the gem to be more feature complete to `ActiveRecord`.

The code used to apply the `db_runtime` field is [here](https://github.com/rails/rails/blob/b326e82dc012d81e9698cb1f402502af1788c1e9/activerecord/lib/active_record/railties/controller_runtime.rb)

### Runtime information:
N/A

Neo4j database version:
`neo4j` gem version: 8.0.0.alpha.10
`neo4j-core` gem version: 7.0.0.alpha.6

NA-V10:
Fixes #1291 

This pull introduces/changes:
 * Adds ActiveGraph database runtime instrumentation through `ActiveSupport::Notifications`
 * Adds `ActiveGraph::RuntimeRegistry` to accumulate Neo4j Bolt request runtime
 * Exposes accumulated database runtime as `payload[:db_runtime]` for Rails controller actions
 * Integrates controller runtime tracking through `ActiveSupport.on_load(:action_controller)`
 * Adds unit tests for runtime accumulation, reset behavior, and controller payload integration

## Summary

Adds ActiveGraph database runtime instrumentation to expose `db_runtime` in Rails controller payloads.

This brings ActiveGraph closer to ActiveRecord's database runtime instrumentation behavior.

## Changes

- Added `ActiveGraph::RuntimeRegistry` to accumulate Neo4j Bolt request runtime.
- Subscribed to `neo4j.core.bolt.request` using `ActiveSupport::Notifications.monotonic_subscribe`.
- Added controller runtime integration through `ActiveSupport.on_load(:action_controller)`.
- Reset runtime before controller actions.
- Added accumulated database runtime to `payload[:db_runtime]`.
- Added unit tests for runtime accumulation and controller payload integration.

## Testing

- Added unit tests for runtime accumulation and reset behavior.
- Added controller runtime tests.

Local test execution is currently pending Ruby/Bundler setup on the development environment.

## Summary by CodeRabbit

* **New Features**
  * Added database runtime tracking for Neo4j Bolt requests.
  * Controller request payloads now include accumulated database runtime (`db_runtime`), including support for incrementing when a value is already present.
  * Ensures runtime counters are cleared/reset around each controller action for accurate per-request metrics.
* **Tests**
  * Added unit specs covering runtime accumulation/reset behavior and controller lifecycle integration.

### DIFF
--- a/lib/active_graph/railtie.rb
+++ b/lib/active_graph/railtie.rb
@@ -64,7 +64,13 @@ module ActiveGraph
         config.app_middleware.insert_after ::ActionDispatch::Callbacks, ActiveGraph::Migrations::CheckPending
       end
     end
+    initializer 'activegraph.controller_runtime' do
+      require 'active_graph/railties/controller_runtime'
 
+      ActiveSupport.on_load(:action_controller) do
+        include ActiveGraph::Railties::ControllerRuntime
+      end
+    end
     initializer "activegraph.deprecator" do |app|
       app.deprecators[:activegraph] = ActiveGraph.deprecator
     end
--- a/lib/active_graph/railties/controller_runtime.rb
+++ b/lib/active_graph/railties/controller_runtime.rb
@@ -1,0 +1,22 @@
+require 'active_graph/runtime_registry'
+
+module ActiveGraph
+  module Railties
+    module ControllerRuntime
+      extend ActiveSupport::Concern
+
+      private
+
+      def process_action(*)
+        ActiveGraph::RuntimeRegistry.reset
+        super
+      end
+
+      def append_info_to_payload(payload)
+        super
+
+        payload[:db_runtime] = (payload[:db_runtime] || 0) + ActiveGraph::RuntimeRegistry.reset
+      end
+    end
+  end
+end
--- a/lib/active_graph/runtime_registry.rb
+++ b/lib/active_graph/runtime_registry.rb
@@ -1,0 +1,38 @@
+require 'active_support/isolated_execution_state'
+
+module ActiveGraph
+  module RuntimeRegistry
+    class Stats
+      attr_accessor :db_runtime
+
+      def initialize
+        @db_runtime = 0.0
+      end
+
+      def reset_runtime
+        db_runtime_was = @db_runtime
+        @db_runtime = 0.0
+        db_runtime_was
+      end
+    end
+
+    extend self
+
+    def call(_name, start, finish, _id, _payload)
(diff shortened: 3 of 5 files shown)


klobuczek (commented):
Thanks for this — the core is well modeled on ActiveRecord and looks correct: a per-request `RuntimeRegistry` backed by `ActiveSupport::IsolatedExecutionState`, `process_action` resetting before the action, `append_info_to_payload` using the same `(… || 0) + RuntimeRegistry.reset` shape, milliseconds throughout, and `monotonic_subscribe` (matching how modern AR accumulates). The total `db_runtime` value should be captured accurately.

Comparing against `ActiveRecord::Railties::ControllerRuntime` (7.1.4), two behaviors are missing that we'd want for parity:

1. **`cleanup_view_runtime` is not overridden.** AR overrides it to subtract DB time incurred *during view rendering* from the reported "Views:" runtime (`runtime - queries_rt`), so the log breakdown doesn't double-count. Without it, a Neo4j query triggered during rendering is counted in **both** the view runtime and `db_runtime`. The total `db_runtime` stays correct; only the Views/DB split in the log is skewed. Worth adding the same two-phase reset (before render into an instance var, then after render).

2. **No `log_process_action`.** AR's `ClassMethods#log_process_action` appends `"ActiveRecord: %.1fms"` to the completed-request log line. As written, the value only lands in `payload[:db_runtime]` — available to subscribers/APM, but there's no `"ActiveGraph: X.Xms"` in the default Rails request log. Adding a `log_process_action` that prints it would complete the parity.

One deliberate divergence to call out: `append_info_to_payload` reads `payload[:db_runtime]` (AR reads its own instance var). That's a reasonable *coexistence* choice — in an app running both AR and ActiveGraph, ActiveGraph's time adds onto AR's — but note it creates an ordering dependency between the two concerns, and the single existing log label ("ActiveRecord: …") would then show the combined total. If coexistence is the intent, a short comment saying so would help; otherwise mirroring AR's instance-var approach keeps the two independent.

Also minor: once CI runs (I've approved it), let's make sure the new specs pass on both MRI and JRuby.

